PR #4, opened 2017-06-12
Basic Travis CI enabled with pylint and unit tests

KellenSunderland:
I've enabled some basic tests and checks to be run with Travis.  After pulling in this PR we'll still have to enable Travis builds, and verify builds are working properly.  Then of course comes the most important step of updating our README with a travis badge.

Small note that the pylint run picked up a missing dep (tensorboard). It's been added in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+sudo: required
+language: python
+
+services:
+  - docker
+
+before_install:
+  - docker pull ubuntu:16.04
+
+python:
+  - "3.5"
+  - "3.6"
+
+install:
+  - pip install -r requirements.txt
+  - pip install pylint
+
+script:
+  - python3 setup.py test
+  - pyline --version
+  - pylint --rcfile=sockeye.pylintrc sockeye -E
+  - pylint --rcfile=sockeye.pylintrc test -E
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 mxnet==0.10.0
 numpy>=1.12
+tensorboard>=1.0.0a3
```
fhieber: pypi is at 1.0.0a6. why not use this version?

tdomhan: I'm anyway not sure we want to add tensorboard as it is a optional dependency and is marked as that in the setup.py already. Do we need it for the Travis CI tests?

KellenSunderland: Good question.  The relevant [python API](https://github.com/dmlc/tensorboard/blob/master/python/tensorboard/writer.py) and  [proto](https://github.com/dmlc/tensorboard/blob/master/tensorboard/src/summary.proto) haven't changed since initial release.  I think we should be ok there from a compatibility point of view.  Version a3 is the first version that supports python3, any version before that will be incompatible with sockeye as it doesn't have a python3 wheel.

KellenSunderland: @tdomhan It's causing lint to fail, but we can easily add a comment

``` pylint 
import tensorflow  # pylint: disable=import-error
```

fhieber: that sounds like the better option to me. The callback unit test also does not use tensorboard.

KellenSunderland: Updated

PR #7, opened 2017-06-13
Support for continuation of aborted training

davvil:
Transparently continue a training if it was aborted. **Important:** Start sockeye with exactly the same arguments.

Because of the asynchronous decoding for BLEU evaluation, some BLEU scores may get lost in the continuation, but this should not have a big effect on the final result.

Results are not yet 100% equal in the continuation, but are certainly in the same range. There must be still some part of the "traning state" missing, put it is surely usable, especially if dealing with long trainings.

### DIFF
--- a/sockeye/callback.py
+++ b/sockeye/callback.py
@@ -18,6 +18,7 @@ early-stopping.
 import logging
 import multiprocessing as mp
 import os
+import pickle
 import shutil
 import time
 from typing import Optional, Tuple, Dict
@@ -243,6 +244,21 @@ class TrainingMonitor(object):
         self._empty_decoder_metric_queue()
         self._write_scores()
 
+    def save_state(self, fname):
+        """
+        Save the state: current metrics and best checkpoint
+        """
+        with open(fname, "wb") as fp:
+            pickle.dump(self.metrics, fp)
+            pickle.dump(self.best_checkpoint, fp)
+
+    def load_state(self, fname):
+        """
+        Load the state: current metrics and best checkpoint
+        """
+        with open(fname, "rb") as fp:
+            self.metrics = pickle.load(fp)
+            self.best_checkpoint = pickle.load(fp)
 
 def _decode_and_evaluate(checkpoint_decoder: sockeye.checkpoint_decoder.CheckpointDecoder,
                          checkpoint: int,
--- a/sockeye/callback.py
+++ b/sockeye/callback.py
@@ -244,17 +244,21 @@ class TrainingMonitor(object):
         self._empty_decoder_metric_queue()
         self._write_scores()
 
-    def save_state(self, fname):
+    def save_state(self, fname: str):
         """
-        Save the state: current metrics and best checkpoint
+        Saves the state: current metrics and best checkpoint.
+
+        :param fname: Name of the file to save the state to.
         """
         with open(fname, "wb") as fp:
             pickle.dump(self.metrics, fp)
             pickle.dump(self.best_checkpoint, fp)
 
-    def load_state(self, fname):
+    def load_state(self, fname: str):
         """
-        Load the state: current metrics and best checkpoint
+        Loads the state: current metrics and best checkpoint.
+
+        :param fname: Name of the file to load the state from.
         """
         with open(fname, "rb") as fp:
             self.metrics = pickle.load(fp)
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -93,6 +93,9 @@ SCHEDULER_STATE_NAME = "scheduler.pkl"
 TRAINING_STATE_PARAMS_NAME = "params"
 ARGS_STATE_NAME = "args.json"
 
+# Arguments that may differ and still resume training
+ARGS_MAY_DIFFER = ["overwrite_output", "use-tensorboard", "quiet", "align_plot_prefix", "sure_align_threshold"]
+
 # data layout strings
 BATCH_MAJOR = "NTC"
 TIME_MAJOR = "TNC"
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -80,6 +80,7 @@ SYMBOL_NAME = "symbol" + JSON_SUFFIX
 METRICS_NAME = "metrics"
 TENSORBOARD_NAME = "tensorboard"
 
+# training resumption constants
 TRAINING_STATE_DIRNAME= "training_state"
 TRAINING_STATE_TEMP_DIRNAME = "tmp.training_state"
 TRAINING_STATE_TEMP_DELETENAME = "delete.training_state"
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -80,6 +80,17 @@ SYMBOL_NAME = "symbol" + JSON_SUFFIX
 METRICS_NAME = "metrics"
 TENSORBOARD_NAME = "tensorboard"
 
+TRAINING_STATE_DIRNAME= "training_state"
+TRAINING_STATE_TEMP_DIRNAME = "tmp.training_state"
+TRAINING_STATE_TEMP_DELETENAME = "delete.training_state"
+MODULE_OPT_STATE_NAME = "mx_optimizer.pkl"
+BUCKET_ITER_STATE_NAME = "bucket.pkl"
+RNG_STATE_NAME = "rng.pkl"
+MONITOR_STATE_NAME = "monitor.pkl"
+TRAINING_STATE_NAME = "training.pkl"
+SCHEDULER_STATE_NAME = "scheduler.pkl"
+TRAINING_STATE_PARAMS_NAME = "params"
+
 # data layout strings
 BATCH_MAJOR = "NTC"
 TIME_MAJOR = "TNC"
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -91,6 +91,7 @@ MONITOR_STATE_NAME = "monitor.pkl"
 TRAINING_STATE_NAME = "training.pkl"
 SCHEDULER_STATE_NAME = "scheduler.pkl"
 TRAINING_STATE_PARAMS_NAME = "params"
+ARGS_STATE_NAME = "args.json"
 
 # data layout strings
 BATCH_MAJOR = "NTC"
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -439,8 +439,8 @@ class ParallelBucketSentenceIter(mx.io.DataIter):
         Appends the actual data, selected by the given indices, to the NDArrays
         of the appropriate bucket. Use when reshuffling the data.
 
-        :param bucket: Current bucket
-        :param shuffled_indices: Indices indicating which data to select
+        :param bucket: Current bucket.
+        :param shuffled_indices: Indices indicating which data to select.
         """
         self.nd_source.append(mx.nd.array(self.data_source[bucket].take(shuffled_indices, axis=0), dtype=self.dtype))
         self.nd_length.append(mx.nd.array(self.data_length[bucket].take(shuffled_indices, axis=0), dtype=self.dtype))
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -16,6 +16,7 @@ Implements data iterators and I/O related functions for sequence-to-sequence mod
 """
 import bisect
 import gzip
+import pickle
 import logging
 import random
 from typing import Dict, Iterator, Iterable, List, NamedTuple, Optional, Tuple
@@ -323,6 +324,7 @@ class ParallelBucketSentenceIter(mx.io.DataIter):
         self.curr_idx = 0
 
         # holds NDArrays
+        self.indices = []  # This will define how the data arrays will be organized
         self.nd_source = []
         self.nd_length = []
         self.nd_target = []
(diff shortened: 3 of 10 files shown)


sockeye/callback.py
```diff
@@ -243,6 +244,21 @@ def stop_fit_callback(self):
         self._empty_decoder_metric_queue()
         self._write_scores()
 
+    def save_state(self, fname):
+        """
+        Save the state: current metrics and best checkpoint
```
fhieber: add
:param fname: Name of file to save the state to.

Also for the method below

fhieber: Change to SaveS the state [...]
Missing period at the end of the docstring.

sockeye/constants.py
```diff
@@ -80,6 +80,17 @@
 METRICS_NAME = "metrics"
 TENSORBOARD_NAME = "tensorboard"
```
fhieber: Maybe add a comment:

```# training resumption constants```

File: sockeye/data_io.py [hunk 3]
```diff
@@ -426,13 +428,24 @@ def reset(self):
...
-            self.nd_label.append(mx.nd.array(self.data_label[i].take(indices, axis=0), dtype=self.dtype))
+            self.indices.append(np.random.permutation(len(self.data_source[i])))
+            self._append_ndarrays(i, self.indices[-1])
+
+    def _append_ndarrays(self, bucket: int, shuffled_indices: np.array):
+        """
+        Appends the actual data, selected by the given indices, to the NDArrays
+        of the appropriate bucket. Use when reshuffling the data.
+
+        :param bucket: Current bucket
```
fhieber: missing period

sockeye/train.py
```diff
@@ -179,9 +187,14 @@ def main():
                                                lr_scheduler=lr_scheduler,
                                                rnn_forget_bias=args.rnn_forget_bias)
 
-        if args.params:
+        # We may consider loading the params in TrainingModule, for consistency
```
fhieber: +1
train.py is getting quite big over time and this is inherently logic that belongs to the training module.
TrainingModel.module is built in the constructor of TrainingModel but only initialized within fit()/_fit(), so we could check for resume_training at that point.

sockeye/training.py
```diff
@@ -33,6 +38,18 @@
 
 logger = logging.getLogger(__name__)
 
+class _TrainingState:
```
fhieber: Could you add a minimal docstring to this class to make sphinx happy?

sockeye/training.py
```diff
@@ -184,35 +201,33 @@ def fit(self,
             if monitor_bleu else None
 
         logger.info("Training started.")
-        training_monitor = sockeye.callback.TrainingMonitor(train_iter.batch_size, output_folder,
-                                                            optimized_metric=optimized_metric,
-                                                            use_tensorboard=use_tensorboard,
-                                                            checkpoint_decoder=checkpoint_decoder)
+        self.training_monitor = sockeye.callback.TrainingMonitor(train_iter.batch_size, output_folder,
```
fhieber: should we set self.training_monitor = None in the constructor then?

sockeye/training.py
```diff
@@ -14,12 +14,17 @@
 """
 Code for training
 """
+from collections import namedtuple
```
fhieber: is this used somewhere?

sockeye/training.py
```diff
@@ -223,16 +238,25 @@ def _fit(self,
...
-        epoch = 0
-        checkpoint = 0
-        updates = 0
-        samples = 0
+
+        training_state_dir = os.path.join(output_folder, C.TRAINING_STATE_DIRNAME)
+        if os.path.exists(training_state_dir):
+            training_state = self.load_checkpoint(training_state_dir, train_iter)
+        else:
+            training_state = _TrainingState(
```
fhieber: why set the defaults explicitly if you have them in your minimal TrainingState class?
If we make them explicit here we can just as well use typing.NamedTuple and have type annotations as in AttentionState/DecoderState

sockeye/training.py
```diff
@@ -275,16 +299,31 @@ def _fit(self,
...
+
+                if training_state.num_not_improved == max_num_not_improved:
+                    logger.info("Model has not improved for %d checkpoints. Stopping fit.", training_state.num_not_improved)
+                    self.training_monitor.stop_fit_callback()
+                    final_training_state_dirname = os.path.join(output_folder, C.TRAINING_STATE_DIRNAME)
+                    if os.path.exists(final_training_state_dirname):
+                        shutil.rmtree(final_training_state_dirname)
                     break
+                else:
+                    self._checkpoint(training_state, output_folder, train_iter)
```
fhieber: moving the checkpoint call to the end of the logic here skips the checkpointing when num_not_improved == max_num_not_improved. Is this what we want ?

sockeye/training.py
```diff
@@ -275,16 +299,31 @@ def _fit(self,
...
 
-    def _evaluate(self, checkpoint, val_iter, val_metric, training_monitor):
+    def _save_params(self, output_folder: str, checkpoint: int):
+        """
+        sync aux params across devices
```
fhieber: This method doesn't really sync aux params across devices, the module.get_params() call does.

Since this is an internal method we dont need to have a docstring necessarily, but if we have one, it should describe what the method overall does (saving the current module parameters to a file)

sockeye/training.py
```diff
@@ -296,16 +335,101 @@ def _evaluate(self, checkpoint, val_iter, val_metric, training_monitor):
...
+
+        # Optimizer state (from mxnet)
+        opt_state_fname = os.path.join(training_state_dirname, C.MODULE_OPT_STATE_NAME)
+        if self.bucketing:
+            # This is a bit hacky, as BucketingModule does not provide a
+            # save_optimizer_states call. We take the current active module and
+            # save its state. This should work, as the individual modules in
+            # BucketingModule share the same optimizer through
+            # borrow_optimizer.
+            self.module._curr_module.save_optimizer_states(opt_state_fname)
```
fhieber: exposing this should be an easy addition to MXNet's BucketingModule

sockeye/training.py
```diff
@@ -296,16 +335,101 @@ def _evaluate(self, checkpoint, val_iter, val_metric, training_monitor):
...
+        # loss if the process is aborted during deletion (which will be slower
+        # than directory renaming)
+        delete_training_state_dirname = os.path.join(output_folder, C.TRAINING_STATE_TEMP_DELETENAME)
+        if os.path.exists(final_training_state_dirname):
+            os.rename(final_training_state_dirname, delete_training_state_dirname)
+        os.rename(training_state_dirname, final_training_state_dirname)
+        if os.path.exists(delete_training_state_dirname):
+            shutil.rmtree(delete_training_state_dirname)
+
+    def save_state(self, training_state: _TrainingState, fname: str):
```
fhieber: missing docstring

sockeye/training.py
```diff
@@ -296,16 +335,101 @@ def _evaluate(self, checkpoint, val_iter, val_metric, training_monitor):
...
+        with open(fname, "wb") as fp:
+            pickle.dump(training_state, fp)
+
+    def load_state(self, fname):
```
fhieber: missing docstring (should also include return value & type)

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+# permissions and limitations under the License.
+
+from math import isclose
+import mxnet as mx
+import random
+import tempfile
+
+import sockeye.data_io
+
+def generate_random_sentence(vocab_size, max_len):
```
fhieber: this could go into test_utils.py, might be useful in the future

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+    length = random.randint(1, max_len)
+    sentence = []
+    for _ in range(length):
```
fhieber: Add a comment why +2
Also:
return [random.randint(3, vocab_size + 2) for _ in range(length)]

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        sentence.append(random.randint(3, vocab_size + 2))
+    return sentence
+
+def create_ParallelBucketSentenceIter(source_sentences, target_sentences, max_len):
```
fhieber: does pylint accept these? otherwise maybe create_parallel_bucket_sentence_iter

PEP8: two newlines between functions

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fhieber: these tests could go into test_data_io.py

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+                                                         eos, pad, unk)
+    return bucket_iterator
+
+def data_batches_equal(db1, db2):
+    # We just compare the data, should probably be enough
+    equal = True
+    for data1, data2 in zip(db1.data, db2.data):
+        equal = equal \
+                and (data1.shape == data2.shape) \
+                and isclose(mx.nd.prod(data1.reshape((-1,)) == data2.reshape((-1,))).asnumpy()[0], 1.0)
```
fhieber: there is an np.isclose as well

test/test_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+    # Simulate some iterations
+    ori_iterator.next()
+    ori_iterator.next()
+    ori_iterator.next()
+    ori_iterator.next()
+    expected_output = ori_iterator.next()
+    # expected_output because the user is expected to call next() after loading
+
+    # Save the state to disk
+    tmp_file = tempfile.NamedTemporaryFile()
```
fhieber: did you make sure this is deleted afterwards?
You can also use a with block (see test_vocab.py)

sockeye/train.py
davvil: Completely agree, but will do this in another PR.

sockeye/training.py
davvil: Not anymore. _TrainingState was initially a namedtuple, but I updated to a class in order to be able to update the values.

sockeye/training.py
davvil: Yep, will be cleaner.

sockeye/training.py
davvil: NamedTuples are problematic here because they are inmutable. Agreed that we do not need the defaults in two places.

sockeye/training.py
davvil: Yes, if the train is finished we delete the state stored by the _checkpoint function.

sockeye/training.py
davvil: Right, leftover from a previous iteration.

test/test_checkpoint.py
davvil: Was thinking about it, put them in a new file in the end to leave the door open to additional tests for checkpointing.

test/test_checkpoint.py
davvil: Will move.

test/test_checkpoint.py
davvil: Pylint complains, but I found it clearer because of the class name. I am open to changing it, though.

test/test_checkpoint.py
davvil: Yes, but it converts again into "float booleans" (i.e. 0.0 for false, 1.0 for true). (Actually it should be safe to compare against 1.0)

test/test_checkpoint.py
davvil: Yep. From the docs: "It will be destroyed as soon as it is closed (including an implicit close when the object is garbage collected)."

sockeye/train.py
fhieber: makes sense

sockeye/training.py
fhieber: oh yes, I missed this. Simple class is fine then

test/test_checkpoint.py
fhieber: ok fine

sockeye/train.py
```diff
@@ -56,6 +58,12 @@ def _build_or_load_vocab(existing_vocab_path: Optional[str], data_path: str, num
         vocabulary = sockeye.vocab.vocab_from_json(existing_vocab_path)
     return vocabulary
 
+def _dict_difference(dict1: Dict, dict2: Dict):
```
fhieber: how about:
diff = set(dict1.keys()) - set(dict2.keys())

maybe you don't need a function then anymore

sockeye/train.py
```diff
@@ -79,14 +87,41 @@ def main():
...
         "Must optimize either BLEU or one of tracked metrics (--metrics)"
 
+    # Checking status of output folder, resumption, etc.
     output_folder = os.path.abspath(args.output)
-    if not os.path.exists(output_folder):
+    resume_training = False
+    training_state_dir = os.path.join(output_folder, C.TRAINING_STATE_DIRNAME)
+    if os.path.exists(output_folder):
+        if args.overwrite_output:
+            shutil.rmtree(output_folder)
```
fhieber: add a logging message

sockeye/train.py
```diff
@@ -79,14 +87,41 @@ def main():
...
+            os.makedirs(output_folder)
+        elif os.path.exists(training_state_dir):
+            with open(os.path.join(output_folder, C.ARGS_STATE_NAME), "r") as fp:
+                old_args = json.load(fp)
+            arg_diffs = _dict_difference(vars(args), old_args) | _dict_difference(old_args, vars(args))
+            # Remove args that may differ without affecting the training.
+            # TODO: should we include device-ids and use-cpu?
```
fhieber: lets not do device-ids since using fewer/more devices affects convergence speed.

sockeye/train.py
```diff
@@ -79,14 +87,41 @@ def main():
...
+            # Are there more?
+            arg_diffs -= set(["overwrite_output", "use-tensorboard", "quiet", "align_plot_prefix", "sure_align_threshold"])
```
fhieber: lets define this set in constants.py

sockeye/train.py
```diff
@@ -79,14 +87,41 @@ def main():
...
+            if not arg_diffs:
+                resume_training = True
+            else:
+                # We do not have the logger yet
```
fhieber: I see, 

can we create a logger two times?
once before this block without a path argument, and then with path as before?

sockeye/train.py
davvil: This does not work because we want to compare the values. In fact the keys normally won't differ as they will be set with default values. I put the two ways mainly for "new versions", when arguments have been added or removed.

fhieber:
Thanks a lot for iterating!

PR #8, opened 2017-06-14
bugfix when counting vocab coverage

beckdaniel:
Small bug in the I/O that would add unks from the target sentence to the source counts. This seems to only affect logging though.


davvil:
Thanks for catching this!

beckdaniel:
No worries!

PR #9, opened 2017-06-15
Allow automatic acquisition of n GPUs through --device-ids -n

fhieber:
More flexible way of automatically acquiring multiple GPUs.

```--device-ids -4``` is equivalent to ```--device-ids -1 -2 -1```

### DIFF
--- a/docs/user_documentation.md
+++ b/docs/user_documentation.md
@@ -78,20 +78,22 @@ Then start tensorboard and point it to the model directory (or any parent direct
 ### CPU/GPU training
 
 By default, training is carried out on the first GPU device of your machine.
-You can specify alternative devices with the `--device-ids` option, with
+You can specify alternative GPU devices with the `--device-ids` option, with
 which you can also activate multi-GPU training (see below). If
 `--device-ids -1`, sockeye will try to find a free GPU on your machine and block
-until one is available. The locking mechanism is based on files and therefore assumes all processes are running
+until all of them are available. The locking mechanism is based on files and therefore assumes all processes are running
 on the same machine with the same file system.
 If this is not the case there is a chance that two processes will be using the same GPU and you run out of GPU memory.
 If you do not have or do not want to use a GPU, specify `--use-cpu`.
 In this case a drop in performance is expected.
 
-Training can be carried out on multiple GPUs using the `--device-ids` flag and specifying multiple GPU device ids:
-`--device-ids 0 1 2 3`.
+##### Multi-GPU training
+Training can be carried out on multiple GPUs by either specifying multiple GPU device ids:
+`--device-ids 0 1 2 3`, or specifying the number GPUs required: `--device-ids -n` attempts to acquire `n` GPUs through
+the locking mechanism described above.
 This will train using [Data Parallelism](https://github.com/dmlc/mxnet/blob/master/docs/how_to/multi_devices.md).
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/user_documentation.md [hunk 1]
```diff
@@ -78,20 +78,22 @@ Then start tensorboard and point it to the model directory (or any parent direct
 ### CPU/GPU training
 
 By default, training is carried out on the first GPU device of your machine.
-You can specify alternative devices with the `--device-ids` option, with
+You can specify alternative GPU devices with the `--device-ids` option, with
 which you can also activate multi-GPU training (see below). If
 `--device-ids -1`, sockeye will try to find a free GPU on your machine and block
-until one is available. The locking mechanism is based on files and therefore assumes all processes are running
+until all of them are available. The locking mechanism is based on files and therefore assumes all processes are running
```
davvil: Unclear, specially the "all of them" as in the example you set it to -1. I would write something along the lines of "If you specify a negative number, sockeye will try to allocate this number of given GPUs, blocking until they become available."

sockeye/arguments.py
```diff
@@ -85,9 +85,11 @@ def add_device_args(params):
     device_params = params.add_argument_group("Device parameters")
 
     device_params.add_argument('--device-ids', default=[-1],
-                               help='List of GPU ids to use. Default: %(default)s. '
-                                    'Use -1 to automatically acquire a GPU through a file locking mechanism. '
-                                    '(Note that this assumes GPU processes are using automatic sockeye GPU ids).',
+                               help='List or number of GPUs ids to use. Default: %(default)s. '
+                                    'Use -x to automatically acquire x GPUs through a file locking mechanism. '
```
davvil: I would remove the "file locking mechanism".

davvil:
Additional comment: In your commit message you probably wanted "--device-ids -1 -1 -1" (not -2)

fhieber:
Actually, -1 -2 -1 was intended as this would automatically acquire 4 gpus, equivalent to the -4

docs/user_documentation.md
fhieber: Thanks, I forgot to go back to the original version in this part of the sentence. I initially had something similar to what you suggest, but the documentation first covers single-device training and then turns to multi-device. Will fix the sentence.

sockeye/arguments.py
fhieber: yeah right, its not really making it clearer there.

sockeye/arguments.py
```diff
@@ -85,9 +85,11 @@ def add_device_args(params):
     device_params = params.add_argument_group("Device parameters")
 
     device_params.add_argument('--device-ids', default=[-1],
-                               help='List of GPU ids to use. Default: %(default)s. '
-                                    'Use -1 to automatically acquire a GPU through a file locking mechanism. '
-                                    '(Note that this assumes GPU processes are using automatic sockeye GPU ids).',
+                               help='List or number of GPUs ids to use. Default: %(default)s. '
+                                    'Use -x to automatically acquire x GPUs. '
+                                    'Use x [x] to use a specific GPU id on this host. '
```
tdomhan: what's 'x [x]' supposed to mean?

fhieber: at least one id, optionally followed by another :)

PR #13, opened 2017-06-16
A simple test for io arguments

logogin:
More tests to follow.

### DIFF
--- a/sockeye/average.py
+++ b/sockeye/average.py
@@ -12,9 +12,9 @@
 # permissions and limitations under the License.
 
 """
-Average parameters from multiple model checkpoints.  Checkpoints can be either
-specificed manually or automatically chosen according to one of several
-strategies.  The default strategy of simply selecting the top-scoring N points
+Average parameters from multiple model checkpoints. Checkpoints can be either
+specified manually or automatically chosen according to one of several
+strategies. The default strategy of simply selecting the top-scoring N points
 works well in practice.
 """
 
--- a/test/test_arguments.py
+++ b/test/test_arguments.py
@@ -1,0 +1,54 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not
+# use this file except in compliance with the License. A copy of the License
+# is located at
+#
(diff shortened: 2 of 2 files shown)


requirements.dev.txt
```diff
@@ -4,3 +4,7 @@ sphinx-autodoc-typehints
 recommonmark
 pypandoc
 twine
+
```
tdomhan: extra newline?

requirements.dev.txt
```diff
@@ -4,3 +4,7 @@ sphinx-autodoc-typehints
...
+pytest
```
tdomhan: these test requirements should go to tests_require in setup.py (and are actually already there). The requirements here are additional requirements on top of that.

requirements.dev.txt
logogin: should I remove it?

requirements.dev.txt
logogin: if I understand correctly `requirements.dev.txt` is for development, so pytest etc. should be declared there. Normally you don't run `setup.py` when import the project into IDE. Also, I am not certain if `setup.py` installs into virtualenvs.

requirements.dev.txt
tdomhan: yes :)

requirements.dev.txt
tdomhan: it's actually a bit of a misnomer. Currently it's used by the document build server to install dependencies needed for generating the documents. Generally the idea was for it to hold anything that is not a direct project dependency or a test dependency, which currently means doc generation dependencies. Maybe we should rename it to requirements.doc.txt.

Now when running setup.py tests all test dependencies will be installed automatically. This should also include virtualenvs. However, IDEs may or may not parse the setup.py for tests_require.

Anyway, not sure what the best setup is. Maybe a requirements.doc.txt for the documentation build server and a requirements.dev.txt with test dependencies (and maybe twine). Could we leave requirements.dev.txt as is for now and then separately decide whether to split it up into doc/dev?

logogin: ok, I will revert the change to `requirements.dev.txt` for now

logogin:
rebased on top of master

PR #17, opened 2017-06-20
Transactionally allocating multiple GPUs, not logging expected Except…

tdomhan:
…ions and properly releasing locks.

### DIFF
--- a/sockeye/train.py
+++ b/sockeye/train.py
@@ -41,7 +41,7 @@ import sockeye.training
 import sockeye.utils
 import sockeye.vocab
 from sockeye.log import setup_main_logger
-from sockeye.utils import acquire_gpu, get_num_gpus
+from sockeye.utils import acquire_gpus, get_num_gpus
 
 
 def none_if_negative(val):
@@ -136,15 +136,15 @@ def main():
             assert num_gpus > 0, "No GPUs found, consider running on the CPU with --use-cpu " \
                                  "(note: check depends on nvidia-smi and this could also mean that the nvidia-smi " \
                                  "binary isn't on the path)."
+            auto_n_required = 0
             context = []
             for gpu_id in args.device_ids:
                 if gpu_id < 0:
                     n_required = -gpu_id
-                    logger.info("Attempting to acquire %d GPUs.", n_required)
-                    # get n_required automatic gpu ids and add to context
-                    context += [exit_stack.enter_context(acquire_gpu()) for _ in range(n_required)]
+                    auto_n_required += n_required
                 else:
                     context.append(gpu_id)
+            context.extend(exit_stack.enter_context(acquire_gpus(auto_n_required)))
             logger.info("Device(s): GPU %s", context)
             context = [mx.gpu(gpu_id) for gpu_id in context]
 
--- a/sockeye/train.py
+++ b/sockeye/train.py
@@ -41,7 +41,7 @@ import sockeye.training
 import sockeye.utils
 import sockeye.vocab
 from sockeye.log import setup_main_logger
-from sockeye.utils import acquire_gpus, get_num_gpus
+from sockeye.utils import acquire_gpus, get_num_gpus, expand_requested_device_ids
 
 
 def none_if_negative(val):
@@ -136,15 +136,10 @@ def main():
             assert num_gpus > 0, "No GPUs found, consider running on the CPU with --use-cpu " \
                                  "(note: check depends on nvidia-smi and this could also mean that the nvidia-smi " \
                                  "binary isn't on the path)."
-            auto_n_required = 0
-            context = []
-            for gpu_id in args.device_ids:
-                if gpu_id < 0:
-                    n_required = -gpu_id
-                    auto_n_required += n_required
-                else:
-                    context.append(gpu_id)
-            context.extend(exit_stack.enter_context(acquire_gpus(auto_n_required)))
+            if args.disable_device_locking:
+                context = expand_requested_device_ids(args.device_ids)
+            else:
+                context = exit_stack.enter_context(acquire_gpus(args.device_ids, lock_dir=args.lock_dir))
             logger.info("Device(s): GPU %s", context)
             context = [mx.gpu(gpu_id) for gpu_id in context]
 
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -274,8 +274,8 @@ def get_num_gpus() -> int:
 def expand_requested_device_ids(requested_device_ids: List[int]) -> List[int]:
     """
     Transform a list of device id requests to concrete device ids. For example on a host with 8 GPUs when requesting
-    [-3, 3, 5] you will get [1, 2, 3, 4, 5]. Namely you will get device 3 and 5, as well as 3 other available
-    device ids.
+    [-4, 3, 5] you will get [0, 1, 2, 3, 4, 5]. Namely you will get device 3 and 5, as well as 3 other available
+    device ids (starting to fill up from low to high device ids).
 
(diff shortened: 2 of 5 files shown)


sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
+            acquired_gpus = []
+            for _ in range(num_gpus):
+                gpu_id = exit_stack.enter_context(acquire_gpu(lock_dir))
+                if gpu_id is not None:
+                    acquired_gpus.append(gpu_id)
+                else:
+                    break
+            if len(acquired_gpus) == num_gpus:
+                try:
+                    yield acquired_gpus
```
davvil: Do we really need a yield here or can we simplify with a return?

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
+        if retry_wait_rand > 0:
+            retry_wait_actual = retry_wait_min + random.randint(0, retry_wait_rand)
+        else:
+            retry_wait_actual = retry_wait_min
+        logger.info("Not enough GPUs available will try again in %ss." % retry_wait_actual)
+        time.sleep(retry_wait_actual)
+
+
+@contextmanager
+def acquire_gpu(lock_dir: str = "/var/lock"):
```
davvil: This is now private to the module, right? (Single GPU should be acquired through acquire_gpus(1)). Rename to _acquire_gpu. Or even more _acquire_single_gpu to avoid confusion with the _gpus function.

tdomhan: yeah technically it's not used and it probably makes sense to make it private. The only use case would be acquiring a single GPU without retries. Anyway, I'm not sure that's a useful usecase, so I'll rename it.

sockeye/utils.py
tdomhan: if we want to use it in a with clause in train.py to automatically release them we still need to the yield. Fun fact: because the previous version of the code was not wrapped in a try: finally block the locks were not actually ever release. This wasn't a huge issue as the operating systems file system anyway releases them when the process dies, but I guess it's not very clean.

sockeye/train.py
```diff
@@ -136,15 +136,15 @@ def main():
...
+            auto_n_required = 0
             context = []
             for gpu_id in args.device_ids:
                 if gpu_id < 0:
                     n_required = -gpu_id
-                    logger.info("Attempting to acquire %d GPUs.", n_required)
-                    # get n_required automatic gpu ids and add to context
-                    context += [exit_stack.enter_context(acquire_gpu()) for _ in range(n_required)]
+                    auto_n_required += n_required
                 else:
```
KellenSunderland: As mentioned offline I think it's reasonable that even hard coded GPU ids will create a file lock indicating that GPU is reserved here.

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
+
+@contextmanager
+def _acquire_single_gpu(lock_dir: str = "/var/lock"):
+    """
+    Acquires a single GPU by locking a file (therefore this assumes that everyone using GPUs calls this method and
+    shares the lock directory).
 
     :param lock_dir: The directory for storing the lock file.
     :param retry_wait: The number of seconds to wait between retries.
+    :return: yields a single GPU id or None if none is available.
```
KellenSunderland: Should be a newline before return no?

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
+                fcntl.flock(lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                # got the lock, let's write our PID into it:
+                lock_file.write("%d\n" % os.getpid())
+                lock_file.flush()
+                logger.info("Acquired GPU %d." % gpu_id)
 
+                try:
                     yield gpu_id
-
+                finally:
```
KellenSunderland: Good catch.

sockeye/train.py
fhieber: +1, training processes take up quite a bit of GPU memory so it makes sense to reserve the full GPU for it.

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
 
 
 @contextmanager
-def acquire_gpu(lock_dir: str = "/var/lock", retry_wait: int = 10):
+def acquire_gpus(num_gpus:int, lock_dir: str = "/tmp", retry_wait_min: int = 30, retry_wait_rand: int = 30):
```
fhieber: missing space: num_gpus: int

missing return value: List[int]

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
+        if retry_wait_rand > 0:
+            retry_wait_actual = retry_wait_min + random.randint(0, retry_wait_rand)
+        else:
+            retry_wait_actual = retry_wait_min
+        logger.info("Not enough GPUs available will try again in %ss." % retry_wait_actual)
+        time.sleep(retry_wait_actual)
+
+
+@contextmanager
+def _acquire_single_gpu(lock_dir: str = "/var/lock"):
```
fhieber: missing return value: -> Optional[int]

sockeye/utils.py
```diff
@@ -271,49 +272,99 @@ def get_num_gpus() -> int:
...
-            with open(lockfile_path, 'w') as lock_file:
-                try:
-                    # exclusive non-blocking lock
-                    fcntl.flock(lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
-                    # got the lock, let's write our PID into it:
-                    lock_file.write("%d\n" % os.getpid())
-                    lock_file.flush()
-                    logger.info("Acquired GPU %d." % gpu_id)
+    for gpu_id in range(num_gpus):
+        # try to acquire a lock
```
fhieber: redundant comment compared to line 339

sockeye/train.py
tdomhan: I'll update the pull request.

sockeye/utils.py
tdomhan: between the :param and :return:? Not that I know of.

sockeye/utils.py
tdomhan: well technically it would be a generator ;)

tdomhan: but then also the @contextmanager makes the whole function a context manager. So technically calling the function gives you a context manager object. So overall I'm not sure how to annotate the function and I'd rather just leave out the annotation of the return type.

sockeye/utils.py
tdomhan: true, removed.

tdomhan:
I pushed to my local branch (https://github.com/tdomhan/sockeye/tree/gpuutilcheck) but somehow this is not showing up yet. Either this is eventually consistent or I'm doing something wrong...

sockeye/utils.py
```diff
@@ -317,28 +392,18 @@ def acquire_gpus(num_gpus:int, lock_dir: str = "/tmp", retry_wait_min: int = 30,
 
 
 @contextmanager
-def _acquire_single_gpu(lock_dir: str = "/var/lock"):
+def _acquire_single_gpu(candidates: List[int], lock_dir: str = "/var/lock"):
```
fhieber: you could make the lock_dir argument not optional or also change it to /tmp to be consistent

tdomhan: oh good catch! will make it non-optional, to not have the default in three places...

tdomhan:
FYI: I'll add 1-2 tests before we should merge this.

fhieber:
awesome, thanks!

PR #23, opened 2017-06-22
Split the _beam_search() method into two parts to enable access to the full beam

artemsok:
Might be useful if some external code needs full beams, not just the best hypothesis.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -510,7 +510,7 @@ class Translator:
         # TODO: max_output_length adaptive to source_length
         max_output_length = bucket_key * 2
 
-        return self._beam_search(source, source_length, bucket_key, max_output_length)
+        return self._get_best_from_beam(*self._beam_search(source, source_length, bucket_key, max_output_length))
 
     def _combine_predictions(self,
                              predictions: List[mx.nd.NDArray],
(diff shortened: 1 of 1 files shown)


sockeye/inference.py
```diff
@@ -638,7 +644,20 @@ def _beam_search(self,
                                     next_dynamic_source in model_next_dynamic_source]
             model_decoder_states = [[mx.nd.take(state, prev_hyp_indices_nd) for state in decoder_states] for
                                     decoder_states in model_next_decoder_states]
+        
+        return sequences, attention_lists, accumulated_scores
 
+    def _get_best_from_beam(self,
```
fhieber: this could be a static method right?

sockeye/inference.py
```diff
@@ -532,9 +532,15 @@ def _beam_search(self,
                      source: mx.nd.NDArray,
                      source_length: mx.nd.NDArray,
                      bucket_key: int,
-                     max_output_length: int) -> Tuple[List[int], np.ndarray, float]:
+                     max_output_length: int) -> Tuple[List[List[int]], List[np.ndarray], mx.nd.NDArray]:
         """
-        Translates a single sentence using beam search.
+            Translates a single sentence using beam search.
```
fhieber: unnecessary indentation. Also, please add a newline between the docstring and the ":param" lines so that sphinx can generate well-formed documentation.

sockeye/inference.py
```diff
@@ -638,7 +644,20 @@ def _beam_search(self,
...
                                     decoder_states in model_next_decoder_states]
+        
+        return sequences, attention_lists, accumulated_scores
 
+    def _get_best_from_beam(self,
+                            sequences: List[List[int]], 
+                            attention_lists: List[np.ndarray],
+                            accumulated_scores: mx.nd.NDArray) -> Tuple[List[int], np.ndarray, float]:
+        """
+            Return the best (aka top) entry from the n-best list.
```
fhieber: Indentation and newline between docstring and param description.

PR #24, opened 2017-06-22
additional tests for arguments parser

logogin:


test/test_arguments.py
```diff
@@ -48,7 +49,93 @@
...
+    _test_args(test_params, expected_params, arguments.add_inference_args)
+
+
+def _test_args(test_params, expected_params, args_func):
     test_parser = argparse.ArgumentParser()
-    arguments.add_io_args(test_parser)
+    args_func(test_parser)
     parsed_params = test_parser.parse_args(test_params.split())
-    assert dict(vars(parsed_params)) == expected_params
+    assert dict(vars(parsed_params)) == expected_params
```
fhieber: File is missing a newline at the end

test/test_attention.py
```diff
@@ -172,7 +172,7 @@ def test_get_context_and_attention_probs():
 
     # data
     source_nd = mx.nd.random_normal(shape=(batch_size, seq_len, num_hidden))
-    source_length_np = np.random.random_integers(1, seq_len, (batch_size,))
```
fhieber: why that change?

test/test_arguments.py
```diff
@@ -13,6 +13,7 @@
 
 import pytest
 
+import sockeye.constants as const
```
fhieber: we have a convention to import constants module as C (which is not ideal for code checking, I know, but we could keep it consistent here)

fhieber:
@logogin ping

test/test_attention.py
logogin: [numpy.random.random_integers](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.random_integers.html) is deprecated

logogin:
rebased

test/test_attention.py
fhieber: I see, makes sense!

PR #26, opened 2017-06-23
Only consider one metric when averaging checkpoints

KellenSunderland:
Metrics files have added additional columns since this code was written.  It is now required to select the column that you wish to use by name.  

Below is an example of a previously present line:

```1       bleu-val=0.001626       perplexity-train=454.856855     perplexity-val=266.975737       time-elapsed=1014.405378```

With the previous code this would have added two values (blue-val + perplexity-val) which would have made selecting a true top-n result based on one of them difficult.

### DIFF
--- a/sockeye/average.py
+++ b/sockeye/average.py
@@ -21,7 +21,7 @@ works well in practice.
 import argparse
 import itertools
 import os
-from typing import Dict, Iterable, Tuple
+from typing import Dict, Iterable, Tuple, List
 
 import mxnet as mx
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: sockeye/average.py [hunk 1]
```diff
@@ -146,7 +149,7 @@ def read_metrics_points(path: str) -> Iterable[Tuple[float, int]]:
                 key_value = field.split("=")
```
tdomhan: would it make sense to fail if we don't find the metric anywhere?

KellenSunderland: Yes, failing with a descriptive error message would probably be a good idea.

sockeye/average.py
```diff
@@ -166,7 +169,12 @@ def main():
         help="either a single model directory (automatic checkpoint selection) "
              "or multiple .params files (manual checkpoint selection)")
     params.add_argument(
-        "--max", action="store_true", help="maximize metric (default: min)")
+        "--max", action="store_true", help="Maximize metric.")
+    params.add_argument(
+        "--metric",
+        help="Name of the metric to choose n-best checkpoints from. (default: {})".format(C.PERPLEXITY),
```
fhieber: in the help message you can write : "Default: %(default)s" to automatically take over what is set as default in kwargs (see arguments.py)

sockeye/average.py
```diff
@@ -166,7 +169,12 @@ def main():
...
+        default=C.PERPLEXITY,
+        choices=[C.PERPLEXITY, C.BLEU])
```
fhieber: I think we can also add accuarcy here (its not tracked by default anymore but is generally supported)

PR #27, opened 2017-06-26
basic tests for selection strategies

logogin:


sockeye/average.py
```diff
@@ -155,6 +135,40 @@ def _read_metrics_points(path: str, metric: str) -> List[Tuple[float, int]]:
     return points
 
 
+def _strategy_best(points, size, maximize):
+    top_n = sorted(points, reverse=maximize)[:size]
```
fhieber: points are tuples of (metric_value, checkpoint) and here we are using the checkpoint ids as a tie-breaker, but the tie-breaker also depends on the maximize flag.
I think it should be implemented such that in case of metric_value ties, we tie-break by maximizing checkpoint id (i.e. choose the later one during training).

test/test_average.py
```diff
@@ -0,0 +1,59 @@
...
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import pytest
+
+import sockeye.average as average
+
+@pytest.mark.parametrize(
+    "test_points, expected_top_n, size, maximize", [
```
fhieber: thanks for adding these tests!
Could you modify the test cases so that they reflect better what a 'point' is, namely a tuple of [metric_value, checkpoint], i.e. having metric values that differ from checkpoint ids? Or in other words, explicitly test for conflicting ranking between metric value and checkpoint id to ensure that the strategy really looks at metric values (first element of the tuple)

fhieber:
@logogin @mjdenkowski Could you guys take a look at this?

logogin:
Rebased and updated test cases to distinguish between metric and point. Refactoring of strategy_lifespan data structure will require additional tests which are not there yet.

test/test_average.py
logogin: done

sockeye/average.py
logogin: agree, the script requires polishing

PR #31, opened 2017-06-27
Allow translate.py to operate on input files in addition to stdin.

KellenSunderland:
Refactored output_handler.py slightly as well to provide a hint about return type.

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -342,6 +342,9 @@ def add_training_args(params):
 
 def add_inference_args(params):
     decode_params = params.add_argument_group("Inference parameters")
+    decode_params.add_argument('--input', '-i',
+                               default=None,
+                               help='A file containing a list of source sentences to translate.')
     decode_params.add_argument('--models', '-m',
                                required=True,
                                nargs='+',
--- a/sockeye/translate.py
+++ b/sockeye/translate.py
@@ -18,6 +18,7 @@ import argparse
 import sys
 import time
 from contextlib import ExitStack
+from typing import Optional, Iterable, Tuple
 
 import mxnet as mx
 
@@ -28,14 +29,14 @@ import sockeye.output_handler
 from sockeye.log import setup_main_logger
 from sockeye.utils import acquire_gpus, get_num_gpus
 
+logger = setup_main_logger(__name__, file_logging=False)
+
 
 def main():
     params = argparse.ArgumentParser(description='Translate from STDIN to STDOUT')
     params = arguments.add_inference_args(params)
     params = arguments.add_device_args(params)
     args = params.parse_args()
-
-    logger = setup_main_logger(__name__, file_logging=False)
 
     assert args.beam_size > 0, "Beam size must be 1 or greater."
     if args.checkpoints is not None:
@@ -51,24 +52,7 @@ def main():
                                                                args.sure_align_threshold)
 
     with ExitStack() as exit_stack:
-        if args.use_cpu:
-            context = mx.cpu()
-        else:
-            num_gpus = get_num_gpus()
-            assert num_gpus > 0, "No GPUs found, consider running on the CPU with --use-cpu " \
-                                 "(note: check depends on nvidia-smi and this could also mean that the nvidia-smi " \
-                                 "binary isn't on the path)."
-            assert len(args.device_ids) == 1, "cannot run on multiple devices for now"
-            gpu_id = args.device_ids[0]
-            if args.disable_device_locking:
-                # without locking and a negative device id we just take the first device
-                gpu_id = 0
-            else:
-                if gpu_id < 0:
-                    # get a single (!) gpu id automatically:
-                    gpu_ids = exit_stack.enter_context(acquire_gpus([-1], lock_dir=args.lock_dir))
-                    gpu_id = gpu_ids[0]
-            context = mx.gpu(gpu_id)
+        context = _setup_context(args, exit_stack)
 
         translator = sockeye.inference.Translator(context,
                                                   args.ensemble_mode,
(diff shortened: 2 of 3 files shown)


sockeye/arguments.py
```diff
@@ -331,6 +331,10 @@ def add_training_args(params):
 
 def add_inference_args(params):
     decode_params = params.add_argument_group("Inference parameters")
+    decode_params.add_argument('--source', '-s',
```
fhieber: can we call it --input / -i ?

sockeye/output_handler.py
```diff
@@ -140,3 +115,29 @@ def handle(self, t_input: sockeye.inference.TranslatorInput, t_output: sockeye.i
                              t_input.tokens,
                              t_output.tokens,
                              self.threshold)
+
+
+def get_output_handler(output_type: str,
```
fhieber: Why do you need to move this down? This forward declaration of return type works for many other places in the code (see attention.py, decoder.py, encoder.py). I also like to have the get function first to see what is available.

sockeye/translate.py
```diff
@@ -18,25 +18,28 @@
...
 from contextlib import ExitStack
+from typing import Optional
 
 import mxnet as mx
 
 import sockeye.arguments as arguments
 import sockeye.data_io
 import sockeye.inference
 import sockeye.output_handler
+from sockeye.inference import Translator
```
fhieber: duplicate import to "import sockeye.inference".
We tend to only import the module not individual names across the project, so I'd go with sockeye.inference.Translator in the code (and in the type annotation you added)

sockeye/translate.py
```diff
@@ -18,25 +18,28 @@
...
 from sockeye.log import setup_main_logger
+from sockeye.output_handler import OutputHandler
```
fhieber: same here

sockeye/translate.py
```diff
@@ -73,27 +63,62 @@ def main():
...
+    :param output_handler: Handler that will write output to a stream.
+    :param translator: Translator that will translate each line of input.
+    :param source: Path to file which will be translated line-by-line if included.
+    """
+    total_time = 0
+    i = 0
+    if source:
+        # Translate from a file.
+        with open(source) as f:
+            i, total_time = _translate(i, output_handler, f.readlines(), total_time, translator, 0)
```
fhieber: reading all input into memory might be dangeruous if translating a HUGE input file. Why not just open a fd or assign to sys.stdin based on args.source/input, like this (also outside of this function):

if args.input is not None:
  input_fd = sockeye.utils.smart_open(args.input)
else:
  input_fd = sys.stdin

with this you can directly call _translate and iterate over it.

sockeye/translate.py
```diff
@@ -73,27 +63,62 @@ def main():
...
+        # Translate from standard in, skip the first line of input.
+        i, total_time = _translate(i, output_handler, sys.stdin, total_time, translator, 1)
+    if i != 0:
+        logger.info("Processed %d lines. Total time: %.4f sec/sent: %.4f sent/sec: %.4f", i, total_time,
+                    total_time / i, i / total_time)
+    else:
+        logger.info("Processed 0 lines.")
+
+
+def _translate(i, output_handler, source_data, total_time, translator, offset):
```
fhieber: why a private function? If we want this to be part of a library it could be a public one (with proper doc string)

test/test_translate.py
```diff
@@ -0,0 +1,57 @@
...
+# or in the "license" file accompanying this file. This file is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+from io import StringIO
+from unittest.mock import patch, Mock, mock_open
+
+import pytest
+
+from sockeye.inference import Translator
```
fhieber: I'd prefer module imports

sockeye/output_handler.py
KellenSunderland: No major motivation on my part, I'll revert it back.  If we used the type as a return hint we'll get a not declared error, but I'm assuming you're already aware of this given that you're using a string here.  

Any motivation other than consistency for the forward declaration?

sockeye/arguments.py
KellenSunderland: Fixed

sockeye/translate.py
KellenSunderland: Fixed

sockeye/translate.py
KellenSunderland: Fixed

sockeye/translate.py
KellenSunderland: Wow, good catch.  I was assuming readlines was lazy, will change.

test/test_translate.py
KellenSunderland: Fixed

sockeye/output_handler.py
fhieber: no, but if you don't mind move it back to the top. The string declaration is exactly to avoid the type error.

sockeye/translate.py
```diff
@@ -78,27 +62,76 @@ def main():
...
+    :param translator: Translator that will translate each line of input.
+    :param source: Path to file which will be translated line-by-line if included.
+    """
+
+    if source:
+        # Translate from a file.
+        i, total_time = translate_lines(output_handler, sockeye.data_io.smart_open(source), translator)
+    else:
+        # Translate from standard in, skip the first line of input.
+        i, total_time = translate_lines(output_handler, sys.stdin, translator, 1)
```
tdomhan: ok, the enumerate(..., 1) actually means that we start counting at 1:
```
In [2]: list(enumerate([1,2]))
Out[2]: [(0, 1), (1, 2)]

In [3]: list(enumerate([1,2], 1))
Out[3]: [(1, 1), (2, 2)]```
I guess we don't need to do this differently for stdin vs from a file and just always start counting from 1.

File: sockeye/arguments.py [hunk 9]
```diff
@@ -342,6 +342,10 @@ def add_training_args(params):
 
 def add_inference_args(params):
     decode_params = params.add_argument_group("Inference parameters")
+    decode_params.add_argument('--input', '-i',
```
tdomhan: we should also enable writing to a file, '-o'

KellenSunderland: will do in a follow up PR.

sockeye/arguments.py
```diff
@@ -342,6 +342,10 @@ def add_training_args(params):
...
+                               required=False,
```
KellenSunderland: fixed.

sockeye/translate.py
KellenSunderland: Ok, thanks for clarifying, updating.

sockeye/translate.py
```diff
@@ -78,27 +62,76 @@ def main():
...
-
-        if i != 0:
-            logger.info("Processed %d lines. Total time: %.4f sec/sent: %.4f sent/sec: %.4f", i, total_time,
-                        total_time / i, i / total_time)
+        read_and_translate(translator, output_handler, args.source)
+
+
+def read_and_translate(translator: sockeye.inference.Translator, output_handler: sockeye.output_handler.OutputHandler,
+                       source: Optional[str] = None) -> None:
+    """
```
KellenSunderland: fixed.

sockeye/translate.py
```diff
@@ -78,27 +62,76 @@ def main():
...
+    if i != 0:
+        logger.info("Processed %d lines. Total time: %.4f sec/sent: %.4f sent/sec: %.4f", i, total_time,
+                    total_time / i, i / total_time)
+    else:
+        logger.info("Processed 0 lines.")
+
+
+def translate_lines(output_handler: sockeye.output_handler.OutputHandler, source_data: Iterable[str],
+                    translator: sockeye.inference.Translator, offset: int = 0) -> Tuple[int, float]:
+    """
```
KellenSunderland: fixed.

sockeye/translate.py
```diff
@@ -78,27 +62,76 @@ def main():
...
+def read_and_translate(translator: sockeye.inference.Translator, output_handler: sockeye.output_handler.OutputHandler,
+                       source: Optional[str] = None) -> None:
+    """
+
+    :param output_handler: Handler that will write output to a stream.
+    :param translator: Translator that will translate each line of input.
+    :param source: Path to file which will be translated line-by-line if included.
+    """
+
+    if source:
```
KellenSunderland: Fixed.

Issue #22, opened 2017-06-21
Issues with GPUs


duyvuleo:
Hi,

I got some issues when running sockeye on GPUs. 

**Issue 1**: leaked semaphores
>> sockeye-train --source ../data/multi30k/train.en.atok \
                       --target ../data/multi30k/train.de.atok \
                       --validation-source ../data/multi30k/val.en.atok \
                       --validation-target ../data/multi30k/val.de.atok \
                       --word-min-count 2 \
                       --rnn-num-layers 1 \
                       --rnn-cell-type gru \
		       --rnn-num-hidden 128 \
		       --num-embed-source 128 \
		       --num-embed-target 128 \
		       --attention-type mlp \
		       --attention-num-hidden 128 \
		       --batch-size 64 \
		       --normalize-loss \
		       --dropout 0.1 \
		       --initial-learning-rate 0.001 \
		       --device-ids 0 \
		       --output ../models/multi30k 
>> 
... (truncated)
[INFO:sockeye.training] Training started.
[INFO:sockeye.callback] Early stopping by optimizing 'perplexity' (minimize=True)
script-multi30k.sh: line 49: 19451 Segmentation fault      sockeye-train --source ../data/multi30k/train.en.atok --target ../data/multi30k/train.de.atok --validation-source ../data/multi30k/val.en.atok --validation-target ../data/multi30k/val.de.atok --word-min-count 2 --rnn-num-layers 1 --rnn-cell-type gru --rnn-num-hidden 128 --num-embed-source 128 --num-embed-target 128 --attention-type mlp --attention-num-hidden 128 --batch-size 64 --normalize-loss --dropout 0.1 --initial-learning-rate 0.001 --device-ids 0 --output ../models/multi30k
[vhoang@gpu047 scripts]$ /pylon2/ci560op/fosterg/tools/anaconda3/lib/python3.6/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 3 leaked semaphores to clean up at shutdown
  len(cache))

**Issue 2**: Permission denied
>> sockeye-train --source ../data/multi30k/train.en.atok \
                       --target ../data/multi30k/train.de.atok \
                       --validation-source ../data/multi30k/val.en.atok \
                       --validation-target ../data/multi30k/val.de.atok \
                       --word-min-count 2 \
                       --rnn-num-layers 1 \
                       --rnn-cell-type gru \
		       --rnn-num-hidden 128 \
[...]

PR #34, opened 2017-06-28
Use the OpenBLAS instead of the MKL version of mxnet.

tdomhan:


tdomhan:
I'm removing the MKL dependency as we've seen people have issues with this version (seg faults). https://github.com/awslabs/sockeye/issues/22

PR #38, opened 2017-06-29
Refactored arguments

KellenSunderland:
I don't think we need to keep returning and assigning params for arguments.  Adding argument groups to the reference should be sufficient.  I also moved average args to our argument file to support reuse in an upcoming commit.

### DIFF
--- a/sockeye/average.py
+++ b/sockeye/average.py
@@ -27,6 +27,7 @@ import mxnet as mx
 
 import sockeye.constants as C
 import sockeye.utils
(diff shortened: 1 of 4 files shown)


sockeye/average.py
```diff
@@ -27,6 +27,7 @@
 
 import sockeye.constants as C
 import sockeye.utils
+import sockeye.arguments
```
tdomhan: I thought we do the from . import arguments things now ;)

PR #40, opened 2017-06-29
Allow for dropout in all decoder layers and fix resetting of rnn layers in decoder.decode().

philschulz:
Before the layers in layer._counter for each rnn layer would grow indefinitely.

### DIFF
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -363,6 +363,8 @@ class StackedRNNDecoder(Decoder):
         lexical_biases = []
 
         self.rnn.reset()
+        for cell in self.rnn._cells:
+            cell.reset()
 
         for seq_idx in range(target_seq_len):
             # hidden: (batch_size, rnn_num_hidden)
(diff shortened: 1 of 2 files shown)


sockeye/encoder.py
```diff
@@ -70,7 +70,7 @@ def get_encoder(num_embed: int,
     if num_layers > 1:
         encoders.append(EncoderClass(num_hidden=rnn_num_hidden,
                                      num_layers=num_layers - 1,
-                                     dropout=0.,
```
fhieber: good catch!

sockeye/decoder.py
```diff
@@ -363,6 +363,8 @@ def decode(self,
         lexical_biases = []
 
         self.rnn.reset()
+        for cell in self.rnn._cells:
```
fhieber: As noted before, this should ideally be fixed by MXNet so that we don't have to access private members.

fhieber: Actually, could you add a small TODO comment that we can remove this once MXNet changes?

philschulz: Done

PR #43, opened 2017-06-30
Clarified logging messages w.r.t early stopping. Some tweaks to Learn…

fhieber:
…ingRateSchedulerPleateuReduce

* added check that reduce factor is between 0 and 1
* not returning a PlateauReduce instance if reduce_factor is >= 1.0 (since it wouldnt reduce the learning rate anyway)
* moved logging message in training to avoid confusing logging statements like "model has not improved for 0 checkpoints"

### DIFF
--- a/sockeye/lr_scheduler.py
+++ b/sockeye/lr_scheduler.py
@@ -97,6 +97,7 @@ class LearningRateSchedulerPlateauReduce(LearningRateScheduler):
     """
 
     def __init__(self, reduce_factor: float, reduce_num_not_improved: int) -> None:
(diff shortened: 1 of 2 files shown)


sockeye/lr_scheduler.py
```diff
@@ -97,6 +97,7 @@ class LearningRateSchedulerPlateauReduce(LearningRateScheduler):
     """
 
     def __init__(self, reduce_factor: float, reduce_num_not_improved: int) -> None:
+        assert 0.0 < reduce_factor <= 1, "reduce_factor should be in ]0,1]."
```
davvil: I continue with my crusade against asserts as user validation messages :-)

PR #46, opened 2017-07-03
Added minimum number of epochs parameter before early-stopping

fhieber:


sockeye/training.py
```diff
@@ -304,20 +310,29 @@ def _fit(self,
...
+                    logger.info("Model has not improved for %d checkpoints", train_state.num_not_improved)
+
+                stop_fit = False
+
+                if train_state.num_not_improved == max_num_not_improved:
+                    logger.info("Stopping fit (no improvements for %d checkpoints)", train_state.num_not_improved)
+                    stop_fit = True
+
+                if (min_num_epochs is not None and train_state.epoch < min_num_epochs):
+                    logger.info("Minimum number of epochs not reached. Will not stop yet.")
```
tdomhan: maybe add the actual epoch in the message:
"Minimum number of epochs (%d) not reached yet."

sockeye/training.py
```diff
@@ -304,20 +310,31 @@ def _fit(self,
...
+
+                if train_state.num_not_improved == max_num_not_improved:
+                    logger.info("Stopping fit (no improvements for %d checkpoints)", train_state.num_not_improved)
+                    stop_fit = True
+
+                if min_num_epochs is not None and train_state.epoch < min_num_epochs:
+                    logger.info("Minimum number of epochs (%d) not reached: %d. Will not stop yet.",
+                                min_num_epochs,
+                                train_state.epoch)
+                    stop_fit = False
```
davvil: Couldn't we end up with contradicting messages here? "Stopping fit (no improvement)" followed with "Sorry, won't stop yet!". I would prefer a single message. And also only show the "Minimum number of epochs" message if we would stop had the option be deactivated.

fhieber: I actually prefer the verbose sequence of messages, as it makes it clear whats happening and also keeps the code simple.
Regarding your other comment, I will open another PR to fix this.

PR #52, opened 2017-07-05
Do not use assertions for checking user input

davvil:
E.g. from https://wiki.python.org/moin/UsingAssertionsEffectively:

Assertions should *not* be used to test for failure cases that can occur
because of bad user input or operating system/environment failures, such
as a file not being found. Instead, you should raise an exception, or
print an error message, or whatever is appropriate. One important reason
why assertions should only be used for self-tests of the program is that
assertions can be disabled at compile time.


fhieber:
Would it make sense to include the condition for the error in the function so that we can make all of these checks single-lined?
Something like:
check_with_msg(a==b, "error message")

Otherwise we have even more if statements  :)

fhieber:
Also, I think we should definitely not sys.exit, but throw an exception to have a stack trace! Maybe we can define a SockeyeException...

davvil:
I prefer the if blocks, I like the visual clues. But I am not completely opposed to including the condition as a parameter.

fhieber:
I guess if we call this function something like "check_condition" its clear that this might throw and avoiding if statements shortens the code. Overall the main reason for the assertions initially was to keep error checking lines as short as possible so that they don't clutter the code too much.

davvil:
I am not sure we need stack traces. Normally the error messages relate to parameters that are not set correctly. A simple error message is normally easier to parse in these cases, in my opinion.

Note also that there are still asserts in the code for internal tests.

davvil:
Changed calls check_condition, as suggested. Did not change to throwing exceptions, though, I would like a second opinion on this. It would only be a 1 line change, though.

fhieber:
Thanks, and agreed, lets get a 2nd opinion on exception/stack trace vs. sys.exit

sockeye/utils.py
```diff
@@ -32,11 +32,22 @@
...
+    """
+    Check the condition and if it is not met, exit with the given error message
+    and error_code, similar to assertions.
+
+    :param condition: Condition to check.
+    :param error_message: Error message to show to the user.
+    :param error_code: Error code to return to the system.
+    """
+    if not condition:
+        raise exception(error_message)
```
davvil: Do all exception types support a message? Or should we just fix RuntimeError and not allow flexibility in the exception?

fhieber: Would you be opposed to have a SockeyeException type?

davvil: No. :-)

tdomhan: what does the SockeyeException give you? I'd just use a RuntimeError or ValueError

tdomhan:
+1 for not doing sys.exit as then we can be called as a library and have the caller handle the exception instead of the program being killed.

sockeye/utils.py
fhieber: it'd just define the scope of the exception more clearly, but no strong feelings about it

test/test_utils.py
```diff
@@ -195,3 +195,19 @@ def test_gpu_file_lock_permission_exception(tmpdir):
...
+def test_check_condition_true():
+    check_condition(1 == 1, "Nice")
+
+
+def test_check_condition_false():
+    with pytest.raises(SockeyeError):
+        check_condition(1 == 2, "Wrong")
+
+
+def test_check_condition_message():
```
tdomhan: can be combined with the above test:
```
    with pytest.raises(SockeyeError) as excinfo:
        check_condition(1 == 2, "Wrong")
    assert "Wrong" == str(excinfo.value)
```

Issue #54, opened 2017-07-06
sockeye/git_version.py not deleted if tests fail


davvil:
If a test fails while running `python setup.py test`, the file `sockeye/git_version.py` is not deleted. Running the tests again (and probably other commands in setup) fails with
```
Traceback (most recent call last):
  File "setup.py", line 128, in <module>
    with temporarily_write_git_hash(get_git_hash()):
  File "/Users/dvilar/anaconda3/lib/python3.6/contextlib.py", line 82, in __enter__
    return next(self.gen)
  File "setup.py", line 48, in temporarily_write_git_hash
    raise RuntimeError("%s already exists, will not overwrite" % filename)
RuntimeError: sockeye/git_version.py already exists, will not overwrite
```

tdomhan:
ohh, good catch!

tdomhan:
https://github.com/awslabs/sockeye/pull/55

PR #55, opened 2017-07-06
Correctly removing git version in case of exceptions.

tdomhan:


tdomhan:
https://github.com/awslabs/sockeye/issues/54

PR #56, opened 2017-07-06
Bucketing update

mjdenkowski:
The goal is to combine two desirable properties:
(1) Use target/source length ratio to efficiently size buckets
(2) Guarantee that all training instances up to (max_seq_len, max_seq_len) are used

This update fixes the ratio calculation and adds extra buckets to "fill in" missing space due to bucket sizing.


sockeye/data_io.py
```diff
@@ -135,7 +138,7 @@ def get_training_data_iters(source: str, target: str,
                                                                           target,
                                                                           vocab_source,
                                                                           vocab_target)
-    length_ratio = sum(len(t) / float(len(s)) for t, s in zip(train_source_sentences, train_target_sentences)) / len(
+    length_ratio = sum(len(t) / float(len(s)) for t, s in zip(train_target_sentences, train_source_sentences)) / len(
```
tdomhan: Oh, good catch!

fhieber:
+1 on Tobis comment. The proposed behavior change introduces a lot of new additional buckets, each of which will incur binding another module/executor group to it. While in theory they should be able to share all of their memory, it often still leads to increased memory usage. So I would suggest keeping the old logic and increase max-seq-len if needed.

sockeye/data_io.py
fhieber: indeed, thanks for catching this!

mjdenkowski:
Another solution is to have both source and target max length options.  That would let us cover very unbalanced cases without all the extra buckets.

One issue with relying too much on the length ratio is that it's a hard cut-off for anything over the average.  For instance, if the average ratio is 0.7, we max out at (100, 70).  Examples with lower ratios such as (100, 60) are allowed while those with higher ratios such as (100, 75) aren't.  This biases our training data in a way that we probably don't want, especially since neural models already have problems producing too-short translations.

Getting around this would require pre-calculating the average ratio plus some allowance (how much?) for deviation, and scaling max_seq_len appropriately.  Even if we did this, fitting an example (100, 90) from above would require buckets up to (130, 91), wasting space on the source side.  The idea behind filling in buckets would be to add (100, 77) ... (100, 91) to better fit examples.

I think ratio plus fill-in is good for most tasks since translation tends to have ratios between 0.5 and 2.  For more unbalanced cases, we can add separate source/target max lengths.

tdomhan:
I totally agree that we should distinguish bucketing from data cleaning. However I think it's totally fine if Sockeye decides to not use 100% of the training data in order to make the model fit into memory. Also I'd rather not have any buckets on either the source or the target side that exceed max_seq_len.

I agree that we should reduce the bias we introduce by these mechanisms. Already by sorting the data and then fitting them into batches we introduce quite some bias. For the particular example you gave you could still come up with sentences of length ratio 0.75 that would fit in the last bucket, such as (91, 68). So the last bucket would not just contain data with length ratio 0.7 or smaller. But you're right that for the sentences going towards the max source length we limit ourselves to sentences with a length_ratio smaller than the average. 

Anyway, having a deviation from the length_ratio like you suggested seems like a good compromise between on memory usage of the (max_seq_len, max_seq_len) solution and what we currently have. For example we could create an additional bucket with (max_seq_len, length_ratio * max_seq_len + window_size) to keep things simple.

mjdenkowski:
Here's a second update.  Now we can choose source and target max length independently to handle cases we know will be unbalanced (subword-to-char translation).  We can also use the regular max length arg to set both at once.  This update also restricts the largest bucket size to exactly max length and deduplicates buckets (can have duplicates after enforcing min size 2).

In these updates, I would like us to be able to offer 2 things to users:
* Intuitive default behavior: max length means max length on both source and target side similar to running Moses/Joshua/cdec corpus cleaning scripts
* Total control over memory usage: set source and target length independently

To bring everything back together, I would suggest a deviation arg similar to what you suggested.  A simple implementation would be something like "--ratio-deviations N" that lets users:
* Limit by source/target max length
* Kick out data with length ratio beyond N standard deviations of the average
* Automatically reset source and target max bucket size based on the retained data
This would let users choose the intuitive default behavior or the smart ratio-based optimization.  That could be a future commit.

fhieber:
I'd vote for not adding more parameters to it but define a reasonable setting that works across all use cases.
So I am ok with merging the current state of the PR (if conflicts are resolved) and see how it behaves memory-wise. If that becomes a problem we can still revisit Tobis suggestion: creating one additional bucket with (max_seq_len, length_ratio * max_seq_len + window_size) to minimize the overall number of buckets.

mjdenkowski:
Conflicts are resolved and it looks like we've reached a version that everyone can live with for now.  We can still explore improving memory usage.  Perhaps we could supply a ratio-based corpus cleaning module and have Sockeye's default behavior be to set max bucket size to min(command line arg, actual longest sentence).

PR #57, opened 2017-07-07
Enable Layer Normalization before non-linear activations across sockeye

fhieber:
LSTM cell normalization is turned on separately via --rnn-cell-type lnlstm as this currently significantly incurs memory usage.

Specifically, with --layer-normalization flag, LN is added to decoder state initialization, MLPAttention, ActivationCoverage, and the hidden transformation in the decoder.


sockeye/coverage.py
```diff
@@ -189,6 +199,11 @@ def __init__(self, coverage_num_hidden: int, activation: str, prefix='') -> None
         self.cov_prev2h_weight = mx.sym.Variable("%scov_prev2h_weight" % self.prefix)
         # attention scores to hidden
         self.cov_a2h_weight = mx.sym.Variable("%scov_a2h_weight" % self.prefix)
+        # optional layer normalization
+        self.lnorm = None
+        if layer_normalization and not self.num_hidden != 1:
+            self.lnorm = LayerNormalization(self.num_hidden,
+                                            prefix="%scov_norm" % prefix) if layer_normalization else None
```
davvil: cov_ too many in the prefix?

sockeye/decoder.py
```diff
@@ -163,14 +171,18 @@ def __init__(self,
         # Decoder parameters
         # RNN init state parameters
         self._create_layer_parameters()
+
         # Hidden state parameters
         self.hidden_w = mx.sym.Variable("%shidden_weight" % prefix)
         self.hidden_b = mx.sym.Variable("%shidden_bias" % prefix)
+        self.hidden_norm = LayerNormalization(self.num_hidden,
+                                              prefix="%shidden_weight_norm" % prefix) if self.lnorm else None
```
davvil: Again, isn't the _weight duplicated in the prefix?

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
...
+        in_transform = mx.sym.Activation(in_transform, act_type="tanh",
+                                         name='%sc' % name)
+        out_gate = mx.sym.Activation(out_gate, act_type="sigmoid",
+                                     name='%so' % name)
+        next_c = mx.sym._internal._plus(forget_gate * states[1], in_gate * in_transform,
+                                        name='%sstate' % name)
+        next_c = self._norm_layers[4].normalize(next_c)
+        next_h = mx.sym._internal._mul(out_gate, mx.sym.Activation(next_c, act_type="tanh"),
+                                       name='%sout' % name)
+        return next_h, [next_h, next_c]
```
tdomhan: I guess the general theme is to always layer normalize before activations, but here we return next_c, which (unlike next_h) does not include the activation. Would it make sense to return the unnormalized version?

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
...
+    def __call__(self, inputs, states):
+        self._counter += 1
+        name = '%st%d_' % (self._prefix, self._counter)
+        i2h = mx.sym.FullyConnected(data=inputs, weight=self._iW, bias=self._iB,
+                                    num_hidden=self._num_hidden * 4,
+                                    name='%si2h' % name)
+        h2h = mx.sym.FullyConnected(data=states[0], weight=self._hW, bias=self._hB,
+                                    num_hidden=self._num_hidden * 4,
+                                    name='%sh2h' % name)
+        gates = i2h + h2h
```
tdomhan: looking at the LN paper (page 13 of the supplementary material) it seems like they apply LN as gates ~= ...LN(i2h) + ... LN(h2h), as well as h_t = ... LN(c_t). However c_t is not normalized.

sockeye/coverage.py
fhieber: hm no, the default prefix of ActivationCoverage is an empty string and thats not overriden anywhere when constructing an instance. I can make the default prefix "cov_" and change the name of cov variables and the prefix for LN

sockeye/decoder.py
fhieber: no, why would it be? prefix is "decoder_".
I can rename to "decoder_hidden_norm" if you dont want the weight in there.

sockeye/arguments.py
```diff
@@ -231,6 +231,12 @@ def add_model_parameters(params):
                               help="Enables a context gate which adaptively weighs the decoder input against the"
                                    "source context vector before each update of the decoder hidden state.")
 
+    model_params.add_argument('--layer-normalization', action="store_true",
```
fhieber: if we find a better name, happy to take it. Otherwise I think its not too confusing given the docstring.

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
             # TODO(fhieber): add pervasive dropout?
             rnn.add(mx.rnn.DropoutCell(dropout, prefix=cell_prefix + "_dropout"))
     return rnn
+
+
+class LnLSTMCell(mx.rnn.LSTMCell):
```
fhieber: Sure, can do it

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
...
+    :param prefix: prefix for name of layers (and name of weight if params is None).
+    :param params: RNNParams or None. Container for weight sharing between cells. Created if None.
+    :param forget_bias: bias added to forget gate, default 1.0. Jozefowicz et al. 2015 recommends setting this to 1.0.
+    :param norm_scale: scale/gain for layer normalization.
+    :param norm_shift: shift/bias after layer normalization.
+    """
+
+    def __init__(self,
+                 num_hidden: int,
+                 prefix: str = 'lstm_',
```
fhieber: probably so that one can't accidentally run with the wrong config.

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
...
+                 params: Optional[mx.rnn.RNNParams] = None,
+                 forget_bias: float = 1.0,
+                 norm_scale: float = 1.0,
+                 norm_shift: float = 0.0) -> None:
+        super(LnLSTMCell, self).__init__(num_hidden, prefix, params, forget_bias)
+
+        self._norm_layers = list()  # type: List[LayerNormalization]
+        for name in ['i', 'f', 'c', 'o', 's']:
+            scale = self.params.get('%s_shift' % name, shape=(num_hidden,),
```
fhieber: that explains a lot, great catch!!

sockeye/rnn.py
```diff
@@ -55,3 +60,68 @@ def get_stacked_rnn(cell_type: str,
...
+                                    init=mx.init.Constant(value=norm_shift))
```
fhieber: yes, fixed it

sockeye/decoder.py
```diff
@@ -155,6 +162,7 @@ def __init__(self,
             self.mapped_rnn_output_b = mx.sym.Variable("%smapped_rnn_output_bias" % prefix)
             self.mapped_context_w = mx.sym.Variable("%smapped_context_weight" % prefix)
             self.mapped_context_b = mx.sym.Variable("%smapped_context_bias" % prefix)
+        self.lnorm = layer_normalization
```
fhieber: sure thing

sockeye/rnn.py
fhieber: Interesting! I hadn't seen this supplementary material but had looked at some other implementations. We could probably do it even more efficient by having only 3 LayerNormalization instances: 2 before the slice in line 106, and 1 for next_c in line 125 (not touching next_c itself which is returned.

sockeye/rnn.py
fhieber: yes, I think we should follow the material in the paper

sockeye/coverage.py
```diff
@@ -19,23 +19,27 @@
...
     Returns a Coverage instance.
 
     :param coverage_type: Name of coverage type.
     :param coverage_num_hidden: Number of hidden units for coverage vectors.
+    :param layer_normalization: If true, applies layer normalization to ActivationCoverage.
     :return: Instance of Coverage.
     """
 
     if coverage_type == "gru":
         return GRUCoverage(coverage_num_hidden)
```
fhieber: Lets defer this to another PR maybe.

sockeye/layers.py
```diff
@@ -11,37 +11,44 @@
...
-                 beta_init: float = 0.0) -> None:
+                 prefix: Optional[str] = None,
+                 scale: Optional[mx.sym.Symbol] = None,
+                 shift: Optional[mx.sym.Symbol] = None,
+                 scale_init: float = 1.0,
+                 shift_init: float = 0.0) -> None:
         self.prefix = prefix
-        self.scale = mx.sym.Variable('%s_gamma' % prefix, shape=(num_hidden,), init=mx.init.Constant(value=gamma_init))
-        self.shift = mx.sym.Variable('%s_beta' % prefix, shape=(num_hidden,), init=mx.init.Constant(value=beta_init))
+        self.scale = scale if scale is not None else mx.sym.Variable('%s_gamma' % prefix, shape=(num_hidden,),
```
fhieber: done

fhieber: done

fhieber:
Thanks for the pointer to the paper. I added two types of layer-normalized LSTM cells, one based on the paper, and the other one with normalization per gate (which seems a bit more intuitive to me and is what other frameworks tend to do). We need to determine experimentally what works better.

sockeye/coverage.py
tdomhan: Sure. The only problem is to do that in a backwards compatible manner in the future, because now if layer_normalization is turned on, the behavior is to not use it in GRUCoverage, so we shouldn't turn LN on for GRUCoverage in the future. I guess one way would be to add another coverage type like "gruln", but it might be confusing that layer_normalization in general only covers some aspects of the network and not others.

sockeye/coverage.py
```diff
@@ -114,10 +118,10 @@ class GRUCoverage(Coverage):
     :param coverage_num_hidden: Number of hidden units for coverage vectors.
     """
 
-    def __init__(self, coverage_num_hidden: int, prefix='') -> None:
+    def __init__(self, coverage_num_hidden: int, prefix='cov_') -> None:
```
tdomhan: I guess this is only backwards compatible because we don't set prefix, right? (just checking that this is backwards compatible.)

sockeye/rnn.py
tdomhan: you should update the prefix here in the new version of the code.

sockeye/rnn.py
```diff
@@ -55,3 +62,153 @@ def get_stacked_rnn(cell_type: str,
...
+            shift = self.params.get('%s_scale' % name, shape=(num_hidden,),
+                                    init=mx.init.Constant(value=norm_scale if name != "f" else forget_bias))
+            self._norm_layers.append(
+                LayerNormalization(num_hidden, prefix="%s%s" % (self._prefix, name), scale=scale, shift=shift))
+
+    @property
+    def state_info(self):
```
tdomhan: now that we want back to inherting from LSTMCell you can remove state_info and _gate_names that were added when inherting from BaseCell.

sockeye/rnn.py
fhieber: will do

sockeye/coverage.py
fhieber: yes, its backwards compatible. The other PR #63 does something similar btw.

sockeye/coverage.py
```diff
@@ -114,10 +118,10 @@ class GRUCoverage(Coverage):
...
         self.prefix = prefix
         self.num_hidden = coverage_num_hidden
-        self.gru = mx.rnn.GRUCell(self.num_hidden, prefix="%scoverage_gru" % self.prefix)
+        self.gru = mx.rnn.GRUCell(self.num_hidden, prefix="coverage_gru")
```
fhieber: will be fixed in #63

sockeye/rnn.py
fhieber: done

PR #60, opened 2017-07-09
Refactored beam search to use only MX NDArray operations (except topk). Achieves ~13% speedup on GPU.

fhieber:
This gets rid of most of the numpy operations during beam search. Only thing missing is an efficient topk & unravel_index implementation in MXNet

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -634,7 +634,8 @@ class Translator:
         # scores_accumulated: chosen smallest scores in scores (ascending).
         scores_accumulated = mx.nd.zeros((self.beam_size, 1), ctx=self.context)
 
-        self.pad_dist += np.inf
+        # reset all padding distribution cells to np.inf
+        self.pad_dist[:] = np.inf
 
         # (0) encode source sentence
         model_states = self._encode(source, source_length, bucket_key)
@@ -650,12 +651,15 @@ class Translator:
             if t == 0:  # only one hypothesis at t==0
                 scores = scores[:1]
             else:
-                # renormalize scores by length+1, but not for finished hyps;
-                # their predicted distribution is set to one-hot at C.PAD_ID.
+                # renormalize scores by length+1 ...
                 scores = (scores + scores_accumulated * lengths) / (lengths + 1)
+                # ... but not for finished hyps.
+                # their predicted distribution is set to their accumulated scores at C.PAD_ID.
                 self.pad_dist[:, C.PAD_ID] = scores_accumulated
+                # this is equivalent to doing this in numpy:
+                #   self.pad_dist[finished, :] = np.inf
+                #   self.pad_dist[finished, C.PAD_ID] = scores_accumulated[finished]
                 scores = mx.nd.where(finished, self.pad_dist, scores)
-                self.pad_dist += np.inf
 
             # (3) get beam_size winning hypotheses
             # TODO(fhieber): once mx.nd.topk is sped-up no numpy conversion necessary anymore.
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -21,20 +21,18 @@ from typing import Dict, List, NamedTuple, Optional, Tuple
 import mxnet as mx
 import numpy as np
 
-import sockeye.bleu
-import sockeye.constants as C
-import sockeye.data_io
-import sockeye.model
-import sockeye.utils
-import sockeye.vocab
-from sockeye.attention import AttentionState
-from sockeye.decoder import DecoderState
-from sockeye.utils import check_condition
+from . import attention
+from . import constants as C
+from . import data_io
+from . import decoder
+from . import model
+from . import utils
+from . import vocab
 
 logger = logging.getLogger(__name__)
 
 
-class InferenceModel(sockeye.model.SockeyeModel):
+class InferenceModel(model.SockeyeModel):
     """
     InferenceModel is a SockeyeModel that supports three operations used for inference/decoding:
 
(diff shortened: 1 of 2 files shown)


sockeye/inference.py
```diff
@@ -393,13 +421,14 @@ def __init__(self,
...
-        self.vocab_target_inv = sockeye.vocab.reverse_vocab(self.vocab_target)
+        self.vocab_target_inv = vocab.reverse_vocab(self.vocab_target)
         self.start_id = self.vocab_target[C.BOS_SYMBOL]
         self.stop_ids = {self.vocab_target[C.EOS_SYMBOL], C.PAD_ID}
         self.models = models
         self.interpolation_func = self._get_interpolation_func(ensemble_mode)
         self.beam_size = self.models[0].beam_size
-        self.buckets = sockeye.data_io.define_buckets(self.models[0].max_input_len)
+        self.buckets = data_io.define_buckets(self.models[0].max_input_len)
+        self.pad_dist = mx.nd.zeros((self.beam_size, len(self.vocab_target)), ctx=self.context) + np.inf
```
davvil: Use mx.nd.full()?

sockeye/inference.py
```diff
@@ -513,156 +547,165 @@ def translate_nd(self, source: mx.nd.NDArray, source_length: mx.nd.NDArray, buck
...
+        """
+        Returns a ModelState for each model representing the state of the model after encoding the source.
+
+        :param source: Source ids. Shape: (1, bucket_key).
+        :param source_length: Source length. Shape: (1,).
+        :param bucket_key: Bucket key.
+        :return: List of ModelStates.
+        """
+        prev_target_word_id = mx.nd.zeros((self.beam_size,), ctx=self.context)
+        prev_target_word_id[:] = self.start_id
```
davvil: mx.nd.full()?

sockeye/inference.py
fhieber: Good idea, thanks!

sockeye/inference.py
fhieber: yep :)

sockeye/inference.py
```diff
@@ -514,156 +547,164 @@ def translate_nd(self, source: mx.nd.NDArray, source_length: mx.nd.NDArray, buck
...
+        attentions = mx.nd.zeros((self.beam_size, max_output_length, bucket_key), ctx=self.context)
+
+        # best_hyp_indices: row indices of smallest scores (ascending).
+        best_hyp_indices = mx.nd.zeros((self.beam_size,), ctx=self.context)
+        # best_word_indices: column indices of smallest scores (ascending).
+        best_word_indices = mx.nd.zeros((self.beam_size,), ctx=self.context, dtype='int32')
+        # scores_accumulated: chosen smallest scores in scores (ascending).
+        scores_accumulated = mx.nd.zeros((self.beam_size, 1), ctx=self.context)
+
+        self.pad_dist += np.inf
```
fhieber: well, its a potentially large matrix so recreating it for every translation is expensive. We can simply create it once when the translator is created.

sockeye/inference.py
```diff
@@ -514,156 +547,164 @@ def translate_nd(self, source: mx.nd.NDArray, source_length: mx.nd.NDArray, buck
...
+            # (2) compute length-normalized accumulated scores in place
+            if t == 0:  # only one hypothesis at t==0
+                scores = scores[:1]
+            else:
+                # renormalize scores by length+1, but not for finished hyps;
+                # their predicted distribution is set to one-hot at C.PAD_ID.
+                scores = (scores + scores_accumulated * lengths) / (lengths + 1)
+                self.pad_dist[:, C.PAD_ID] = scores_accumulated
+                scores = mx.nd.where(finished, self.pad_dist, scores)
+                self.pad_dist += np.inf
```
fhieber: scores are the beam_size * target_vocab_size, namely the softmax distributions for the next word for every hypothesis. We normalize with length in-place (line 655). However, if the beam contained some already finished hypothesis, they should not be re-normalized with length+1, and also should not be able to expand anything else than padding symbol.

The idea is that when we call smallest_k below, finished hypotheses can still compete for a place in smallest_k with their accumulated score from a previous iteration.

The code above just puts this into ndarray operations. The old code was equivalent (old lines 602 - 609. MXNet's only conditional operation is mx.nd.where, which requires the last two arguments to be of the same shape. That means we have to create pad_dist matrix that is of same shape as scores. To no re-create this array every single time, we use a class member that is temporarily updated with accumulated scores which are immediately overwritten after the where call.

sockeye/inference.py
davvil: I see, makes sense.

sockeye/inference.py
davvil: Ok, that's what I understood. I would rename pad_dist to padding_scores or something similar.

Still, the pad_dist is initialized to np.inf and then only the C.PAD_ID column is reset. We do not really need the += np.inf every time.

For skipping it completely I was thinking about something along the lines of
```
scores[finished, :] = np.inf
scores[finished, C.PAD_ID] = scores_accumulated[finished]
```
But this does not work for symbols, right?

fhieber: Ok, will rename and I think you are right we can get rid of the self.pad_dist += np.inf since we always overwrite the same complete column in every loop before.

Regarding your suggestion: yeah sure, if we would be in numpy dreamland that would work :) unfortunately this doesn't work for NDArrays (not symbols) afaik.

davvil: NDArrays, right. We might include these two lines as comments, might make it easier to understand.

fhieber: agreed. updated the PR

sockeye/utils.py
```diff
@@ -256,7 +256,7 @@ def get_alignments(attention_matrix: np.ndarray, threshold: float = .9) -> Itera
                 yield (src_idx, trg_idx)
 
 
-def average_arrays(arrays: List[mx.sym.NDArray]) -> mx.sym.NDArray:
+def average_arrays(arrays: List[mx.nd.NDArray]) -> mx.nd.NDArray:
```
tdomhan: huh, interesting catch.

PR #62, opened 2017-07-11
Transformer models in Sockeye

fhieber:
**This will and should not be merged.**
It's just for informational purposes and to have a first review on the code.


sockeye/encoder.py
```diff
@@ -169,11 +216,44 @@ def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int)
...
                                      output_dim=self.num_embed,
-                                     name=self.prefix + 'embed')
+                                     name=self.prefix + "embed")
+        if self.add_pos_encoding:
+            embedding = self._add_pos_encoding(embedding, seq_len)
         if self.dropout > 0:
             embedding = mx.sym.Dropout(data=embedding, p=self.dropout, name="source_embed_dropout")
         return embedding
 
+    def _add_pos_encoding(self, embedding: mx.sym.Symbol, seq_len: int) -> mx.sym.Symbol:
```
davvil: I would write out positional, not be confused with POS tags

davvil: I would write out 'positional', as to not confuse them with POS tags.

sockeye/encoder.py
```diff
@@ -169,11 +216,44 @@ def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int)
...
+        for i in range(self.num_embed):
+            # (seq_len, 1)
+            pos_encoding_dim = mx.sym.sin(position / mx.sym.pow(10000, ((2 * i) / self.num_embed)))
+            pos_encoding_dims.append(pos_encoding_dim)
+        # (seq_len, num_embed)
+        pos_encoding = mx.sym.concat(*pos_encoding_dims, dim=1)
+        # (1, seq_len, num_embed)
+        pos_encoding = mx.sym.expand_dims(data=pos_encoding, axis=0)
+        # Position encodings are constant
+        pos_encoding = mx.sym.BlockGrad(pos_encoding)
```
davvil: We should probably pre-compute the pos-encoding up to the maximum input length as to avoid recomputing the sin for each input. Then we can just slice and add.

sockeye/encoder.py
```diff
@@ -169,11 +216,44 @@ def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int)
...
+        """
+
+        # (seq_len)
+        position = mx.sym.arange(start=0, stop=seq_len)
+        # (seq_len, 1)
+        position = mx.sym.expand_dims(data=position, axis=1)
+        pos_encoding_dims = []
+        for i in range(self.num_embed):
+            # (seq_len, 1)
+            pos_encoding_dim = mx.sym.sin(position / mx.sym.pow(10000, ((2 * i) / self.num_embed)))
```
davvil: Distinction between odd and even positions not yet included, right?

sockeye/encoder.py
```diff
@@ -413,3 +493,90 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+            feed_forward = FFNRelu(prefix="%sffn_%d_" % (self.prefix, i),
+                                   num_hidden=self.feed_forward_num_hidden,
+                                   num_model=self.model_size,
+                                   dropout=self.dropout)
+            # Layer normalization after feed-forward
+            feed_forward_ln = LayerNormalization(num_hidden=self.model_size,
+                                                 prefix="%sffn_ln_%d_" % (self.prefix, i))
+
+            # Apply one layer of the encoder
+            def apply(data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int):
```
davvil: Missing return value

sockeye/layers.py
```diff
@@ -68,8 +69,205 @@ def normalize(self, inputs: mx.sym.Symbol, eps: float = 0.00001) -> mx.sym.Symbo
...
+        # logits: (batch * heads, length, length)
+        logits = mx.sym.swapaxes(data=logits, dim1=0, dim2=1)
+
+        # weights: (batch * heads, length, length)
+        weights = mx.sym.softmax(logits)
+
+        if self.dropout > 0.0:
+            weights = mx.sym.Dropout(weights, p=self.dropout)
+
+        if self.use_loop:
```
davvil: Is there any difference between use_loop or not?

sockeye/encoder.py
fhieber: I was thinking the same. @mjdenkowski are we missing something?

sockeye/encoder.py
fhieber: I changed the positional encoding to a variable initialization scheme in the last commit.

sockeye/layers.py
fhieber: no, the for loop was just my first attempt. I will leave it in a little bit more but eventually we can remove it

fhieber:
\

PR #63, opened 2017-07-12
Use basic prefix for attention and coverage layers consistently. Bump…

fhieber:
… minor version


sockeye/coverage.py
```diff
@@ -77,8 +81,8 @@ class CountCoverage(Coverage):
     Coverage class that accumulates the attention weights for each source word.
     """
 
-    def __init__(self, prefix='') -> None:
-        self.prefix = prefix
+    def __init__(self) -> None:
+        super().__init__()
```
tdomhan: why is it a good idea to not expose the prefix in all derived classes anymore? I would still expose it.

sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.0.3'
+__version__ = '1.0.4'
```
tdomhan: From what I can see this change is backward compatible, right?

sockeye/coverage.py
fhieber: They all default to the default prefix for now which is not really a problem because loading the correct type of attention is handled via attention-type not the names of symbols in the graph.

If you want I can expose, but I don't see a good reason to name them differently right now.

sockeye/__init__.py
fhieber: Yes, so should i revert this?

fhieber:
@tdomhan can I merge?

PR #64, opened 2017-07-12
Specify maximum number of params files to keep

davvil:
Argument --keep-last-params, defaults to 20

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -371,7 +371,7 @@ def add_training_args(params):
 
     train_params.add_argument('--keep-last-params',
                               type=int,
-                              default=20,
+                              default=-1,
                               help='Keep only the last n params files, use -1 to keep all files. Default: %(default)s')
 
 
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -5,7 +5,7 @@
 # is located at
 #
 #     http://aws.amazon.com/apache2.0/
-# 
+#
 # or in the "license" file accompanying this file. This file is distributed on
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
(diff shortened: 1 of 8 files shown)


sockeye/arguments.py
```diff
@@ -369,6 +369,11 @@ def add_training_args(params):
                               default=13,
                               help='Random seed. Default: %(default)s.')
 
+    train_params.add_argument('--keep-last-params',
+                              type=int,
+                              default=20,
```
tdomhan: would it make sense to set the default to -1 to stay backwards compatible? If not I'd say we should set it to something higher than 20, as with values of num_not_improved as high as 16 (which we probably should also set to as the default in arguments.py), one would have only 4 checkpoints before the last checkpoint that improved. This would then hurt the 'best' strategy of average.

tdomhan: looking at an example model that trained for 188 checkpoints and with an average strategy of 'best' with 8 checkpoints, the earliest checkpoint we use for averaging is 168. So 20 would still be fine. That said, I'd still increase the default to 30-40 to be on the save side.

sockeye/training.py
```diff
@@ -174,6 +177,7 @@ def fit(self,
         :param train_iter: The training data iterator.
         :param val_iter: The validation data iterator.
         :param output_folder: The folder in which all model artifacts will be stored in (parameters, checkpoints, etc.).
+        :param max_params_files_to_keep: Maximum number of params files to keep in the output folder
```
fhieber: period missing :)

davvil:
Do we want to keep the *best* n checkpoints instead of the *last* n?

fhieber:
I'd go for last n because we dont want to already decide during training how to choose or average checkpoints.

tdomhan:
I'd also say last n, with no being large enough to later to either best or last averaging.

sockeye/training.py
```diff
@@ -174,6 +177,7 @@ def fit(self,
         :param train_iter: The training data iterator.
         :param val_iter: The validation data iterator.
         :param output_folder: The folder in which all model artifacts will be stored in (parameters, checkpoints, etc.).
+        :param max_params_files_to_keep: Maximum number of params files to keep in the output folder.
```
tdomhan: Maybe add that we keep the last n params.

PR #65, opened 2017-07-12
More strict pep8 conformance

davvil:
Tested with flake8, maximum line length increased to 120

### DIFF
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -5,7 +5,7 @@
 # is located at
 #
 #     http://aws.amazon.com/apache2.0/
-# 
+#
 # or in the "license" file accompanying this file. This file is distributed on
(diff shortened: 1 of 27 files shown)


davvil:
The CI tests fail because in rnn.py mypy needs the 'from typing import List', which is only used in mypy-oriented comments, thus flake8 complains about an unused import. Will try to make both happy.

davvil:
It seems there is not a clean solution yet. https://gitlab.com/pycqa/flake8/issues/118

fhieber:
Should we then leave it with mypy compatibility and not care too much about flake?

davvil:
Says the one who always finds the missing two spaces before inline comments :-D

I tried with #noqa for flake8 and pylint: disable, but we cannot have both in one line AFAIK. Putting it in several lines is really ugly, so I removed them and put a comment about the import. That is the only complain by flake8 currently and mypy is happy again (CI tests).

Issue #66, opened 2017-07-12
ModelConfig getting cluttered with parameters


fhieber:
With more and more flags being added to training, the ModelConfig object gets cluttered with parameters (and their default settings) that are often not even used as their are conditional parameters.
It would be much better to have hierarchical configs that are put together in a composite configuration. We could split configurations into
- encoder
- decoder
- attention
configurations, each of one being an individual configuration object.

Configuration objects should support JSON serialization, ideally also for nested configurations.

We might not be able to continue to use a NamedTuple for a config object as it may contain another config object as a member and this can be easily serialized.

Lets discuss if we should come up with our own ConfigObject implementation or use some existing library.

fhieber:
How about we create a simple class like this, which would already support most of our use cases:

```python3
import jsonpickle
jsonpickle.set_encoder_options('json', sort_keys=True, indent=4)


class Config:
    _frozen = False

    def __init__(self, **kwargs):

        self.__dict__.update(kwargs)

    def __setattr__(self, key, value):
        if self._frozen:
            raise TypeError("Cannot set '%s' in frozen Config" % key)
        object.__setattr__(self, key, value)

    def freeze(self):
        self._frozen = True

    def __repr__(self):
        return "Config: %s" % str(self.__dict__)

    def save(self, fname):
        with open(fname, 'w') as fout:
            fout.write(jsonpickle.encode(self))

    @staticmethod
    def load(fname: str):
        with open(fname) as fin:
            return jsonpickle.decode(fin.read())


# main config with a couple of params
config = Config(num_hidden=4, rnn=True)
# set values afterwards if not frozen
config.filters = [1, 2, 3]
# nested config
config.encoder = Config(max_seq_len=100)

print(config)

config.save("config")
config2 = Config.load("config")

print(config2.encoder.max_seq_len)

config2.freeze()
# throws
config2.decoder = None
```

which returns this:
```
Config: {'num_hidden': 4, 'rnn': True, 'filters': [1, 2, 3], 'encoder': Config: {'max_seq_len': 100}}
100
Traceback (most recent call last):
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/config.py", line 65, in <module>
    config2.decoder = None
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/config.py", line 29, in __setattr__
    raise TypeError("Cannot set '%s' in frozen Config" % key)
TypeError: Cannot set 'decoder' in frozen Config
```

[...]

PR #67, opened 2017-07-12
Clean commit of character sequence encoder

mjdenkowski:
I was originally holding off merging this until we sorted out the problem with character-level decoding (offer users full char-to-char MT).  However, that's an issue with existing code that we can work out separately.  I've gotten a couple of requests to try this with char-to-bpe which runs fine.  (In short, the issues aren't with this code, so nothing should block a merge)

The rest should look familiar to people who have seen this code before:

This is an initial implementation of the character-level encoder from Lee at al. (https://arxiv.org/pdf/1610.03017.pdf)  It uses convolutional, pooling, and highway layers to encode a sequence of character embeddings into a shorter sequence of segment embeddings.  It can be seen as an extra transformation on the input embedding sequence before the bidirectional RNN runs.  The output segment embeddings become the "real" source sequence that subsequent encoders, decoders, and attention mechanisms operate on, making it compatible with other architectures including the recently published convolutional encoder and transformer.

Other implementation notes to follow.


sockeye/arguments.py
```diff
@@ -148,11 +148,41 @@ def add_model_parameters(params):
...
                               default=50000,
                               help='Maximum vocabulary size. Default: %(default)s.')
+    model_params.add_argument('--num-words-source',
+                              type=int_greater_or_equal(0),
+                              default=None,
+                              help='Maximum source vocabulary size. Overrides --num-words. Default: %(default)s')
+    model_params.add_argument('--num-words-target',
+                              type=int_greater_or_equal(0),
+                              default=None,
+                              help='Maximum target vocabulary size. Overrides --num-words. Default: %(default)s')
```
mjdenkowski: Support different source/target vocabulary sizes to enable bpe-to-char and char-to-bpe translation.

sockeye/encoder.py
```diff
@@ -49,14 +56,30 @@ def get_encoder(num_embed: int,
...
+
+    if char_seq_encoder:
+        encoders.append(CharacterSequenceEncoder(num_embed=num_embed,
+                                                 max_filter_width=cse_max_filter_width,
+                                                 num_filters=cse_num_filters,
+                                                 pool_stride=cse_pool_stride,
+                                                 num_highway_layers=cse_num_highway_layers,
+                                                 prefix=C.CHAR_SEQ_ENCODER_PREFIX,
+                                                 dropout=dropout))
+
```
mjdenkowski: After input embeddings, before conversion to time major for bidirectional RNN (or any other encoder).

sockeye/encoder.py
```diff
@@ -108,6 +131,18 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+        Returns the data lengths of the encoded sequence.
+        """
+        return data_length
+
+    def get_encoded_seq_len(self, seq_len: int) -> int:
+        """
+        Returns the size of the encoded sequence.
+        """
+        return seq_len
+
```
mjdenkowski: Encoders now report modifications to input sequence length.  Encoders that don't modify length don't need to implement these methods as the default is just returning the input length.


 
--



 
--

sockeye/encoder.py
```diff
@@ -206,8 +241,12 @@ def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int)
         :param seq_len: Maximum sequence length.
         :return: Encoded input data.
         """
+        current_data_length = data_length
+        current_seq_len = seq_len
         for encoder in self.encoders:
-            data = encoder.encode(data, data_length, seq_len)
+            data = encoder.encode(data, current_data_length, current_seq_len)
+            current_data_length = encoder.get_encoded_data_length(current_data_length)
+            current_seq_len = encoder.get_encoded_seq_len(current_seq_len)
```
mjdenkowski: Encoder sequences propagate length information (both max sequence length and the actual lengths of each sequence).

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+          Jason Lee; Kyunghyun Cho; Thomas Hofmann (https://arxiv.org/pdf/1610.03017.pdf)
+
+    :param num_embed: Input embedding size.
+    :param max_filter_width: Maximum filter width for convolutions.
+    :param num_filters: Number of filters of each width.
+    :param pool_stride: Stride for pooling layer after convolutions.
+    :param num_highway_layers: Number of highway layers for segment embeddings.
+    :param prefix: Name prefix for symbols of this encoder.
+    :param dropout: Dropout probability.
+    """
```
mjdenkowski: Right now this is a straightforward implementation of Figure 1 in the paper.  There's some room for optimization if people start using this regularly.

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+        :param data_length: Vector with sequence lengths.
+        :param seq_len: Maximum sequence length.
+        :return: Encoded input data.
+        """
+
+        total_num_filters = sum(self.num_filters)
+        encoded_seq_len = self.get_encoded_seq_len(seq_len)
+
+        # (batch_size, channel=1, seq_len, num_embed)
+        data = mx.sym.Reshape(data=data, shape=(-1, 1, seq_len, self.num_embed))
```
mjdenkowski: MXNet's convolution and pooling layers are built for image processing (with much hard-coded behavior), so we make our sentences look like images.

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+            pad_before = ceil((filter_width - 1) / 2)
+            pad_after = floor((filter_width - 1) / 2)
+            # (batch_size, channel=1, seq_len + (filter_width - 1), num_embed)
+            padded = mx.sym.pad(data=data,
+                                mode="constant",
+                                constant_value=0,
+                                pad_width=(0, 0, 0, 0, pad_before, pad_after, 0, 0))
+            # (batch_size, num_filter, seq_len, num_scores=1)
+            conv = mx.sym.Convolution(data=padded,
+                                      #cudnn_tune="off",
```
mjdenkowski: There are some memory/speed tuning options here.  This is one of many performance optimization TODOs (commented out -> pick what cuDNN thinks is best).

sockeye/inference.py
```diff
@@ -100,13 +100,15 @@ def _build_modules(self):
 
         def encoder_sym_gen(source_seq_len: int):
             source_encoded = self.encoder.encode(source, source_length, seq_len=source_seq_len)
+            source_encoded_length = self.encoder.get_encoded_data_length(source_length)
+            source_encoded_seq_len = self.encoder.get_encoded_seq_len(source_seq_len)
```
mjdenkowski: Propagate length information for inference.

sockeye/training.py
```diff
@@ -113,9 +113,11 @@ def sym_gen(seq_lens):
             source_seq_len, target_seq_len = seq_lens
 
             source_encoded = self.encoder.encode(source, source_length, seq_len=source_seq_len)
+            source_encoded_length = self.encoder.get_encoded_data_length(source_length)
+            source_encoded_seq_len = self.encoder.get_encoded_seq_len(source_seq_len)
             source_lexicon = self.lexicon.lookup(source) if self.lexicon else None
 
-            logits = self.decoder.decode(source_encoded, source_seq_len, source_length,
+            logits = self.decoder.decode(source_encoded, source_encoded_seq_len, source_encoded_length,
```
mjdenkowski: Propagate length information for training.

sockeye/arguments.py
```diff
@@ -148,11 +148,41 @@ def add_model_parameters(params):
...
     model_params.add_argument('--word-min-count',
                               type=int_greater_or_equal(1),
                               default=1,
                               help='Minimum frequency of words to be included in vocabularies. Default: %(default)s.')
 
+    model_params.add_argument('--char-seq-encoder',
```
fhieber: 1) would it make sense to rename this to a ConvolutionalEncoder? It is only a char-seq encoder if the data fed to it are sequences of characters not words. It could just as well be used for words.
2) We could change the cmdline similarly to the transformer branch where we add an --encoder flag that enables selection of different encoder types (stacked-rnn, transformer, convolutionial)

sockeye/encoder.py
fhieber: Lets have multiple functions that define reasonable defaults for different types of encoders and switch on calling these functions based on the --encoder flag

sockeye/encoder.py
```diff
@@ -226,6 +265,24 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
                 cells.append(cell)
         return cells
 
+    def get_encoded_data_length(self, data_length: mx.sym.Symbol) -> mx.sym.Symbol:
+        """
+        Returns the data lengths of the encoded sequence.
+        """
+        current_data_length = data_length
```
fhieber: why not self.encoders[-1].get_...()?

sockeye/encoder.py
```diff
@@ -226,6 +265,24 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+        for encoder in self.encoders:
+            current_data_length = encoder.get_encoded_data_length(current_data_length)
+        return current_data_length
+
+    def get_encoded_seq_len(self, seq_len: int) -> int:
```
fhieber: same here

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
         Returns a list of RNNCells used by this encoder.
         """
         return self.forward_rnn.get_rnn_cells() + self.reverse_rnn.get_rnn_cells()
+
+
+class CharacterSequenceEncoder(Encoder):
```
fhieber: As mentioned above I would rename to ConvolutionalEncoder in the spirit of Lee et al.

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+                 num_embed: int,
+                 max_filter_width: int = 8,
+                 num_filters: List[int] = None,
+                 pool_stride: int = 5,
+                 num_highway_layers: int = 4,
+                 prefix: str = C.CHAR_SEQ_ENCODER_PREFIX,
+                 dropout: float = 0.):
+        if not num_filters:
+            num_filters = [200, 200, 250, 250, 300, 300, 300, 300]
+        assert len(num_filters) == max_filter_width, "num_filters must have max_filter_width elements."
```
fhieber: lets use utils.check_condition here as this validates user input

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+        self.num_embed = num_embed
+        self.max_filter_width = max_filter_width
+        self.num_filters = num_filters[:]
```
fhieber: not sure what this is doing. Does this create a numpy array?

sockeye/encoder.py
fhieber: Instead of adding more functions to the interface (that we need to remember to be called), could we return those two pieces of information by default from the encode() method?
That is, change the signature of `encode` to 
```python3
encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int) -> Tuple[mx.sym.Symbol, mx.sym.Symbol,int]:
```

Then most of the encoders just return their encode-input (data_length and seq_len), but the convolutional encoder for example can return modified values. The encoder sequence just returns the last results of the self.encoders.encode() call.

sockeye/encoder.py
fhieber: Nevermind, got it. But if we let encode() return this information directly it might be easier for callers

sockeye/arguments.py
davvil: Wrt. 2. If I understand it correctly, this not really a new encoder but a new "embedding" if you want. I.e. we could apply it to RNNs, transformers or other architectures.

sockeye/encoder.py
davvil: #66?

sockeye/encoder.py
davvil: If I understand it correctly, it is creating a copy (instead of just creating a reference).

sockeye/encoder.py
```diff
@@ -413,3 +470,173 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+                              shape=(-1, total_num_filters, encoded_seq_len))
+        # (batch_size, seq_len/stride, total_num_filters)
+        pool = mx.sym.swapaxes(data=pool, dim1=1, dim2=2)
+        if self.dropout > 0:
+            pool = mx.sym.Dropout(data=pool, p=self.dropout)
+
+        # Highway network
+        # (batch_size * seq_len/stride, total_num_filters)
+        pool_flat = mx.sym.Reshape(data=pool, shape=(-3, total_num_filters))
+        seg_embedding = pool_flat
```
davvil: Minor: I got confused with the two step renaming. Better use `seg_embedding = mx.sym.Reshape(...)` directly.

sockeye/arguments.py
fhieber: An embedding is an encoder as well for us. But I agree, we can still mix and match between different Encoder implementations and should probably eventually expose more combinations than the simple forward + reverse [+ n * forward] RNN encoder.

fhieber:
Regarding @davvil question to bucketing: as far as I can tell this should not be a problem since the INPUT to the graphs is still given by the input to the encoder and the modified sequence lengths / seq_len values are a function of the original input size.
So bucketing should not be affected here.

davvil:
The bucketing is not "technically" affected, but if the inputs are characters and the output is bpe, the ratios will greatly change.

fhieber:
Yes, but this is not related to this PR and already handled by the recent bucketing update.

sockeye/arguments.py
mjdenkowski: We could call it ConvolutionalEmbeddingEncoder to avoid confusion with Facebook's convolutional encoder.  We could pack everything into the --encoder arg if we have a good idea for what we want that to look like.  Two obvious options are:

1. Let users list any encoders they like in sequence, basically moving get_encoder's functionality to the command line:
`--encoder embed birnn`
`--encoder embed conv-embed birnn`
`--encoder embed conv-embed transformer`
Here we have to keep track of when to add batch2time and other "helper" encoders.

2. Have some presets for users:
`--encoder rnn`
`--encoder rnn-with-conv-embed`
`--encoder transformer`
`--encoder transformer-with-conv-embed`
Here we manage everything.

My suggestion would be to go with 2 for now and move toward 1.  Thoughts?

sockeye/encoder.py
mjdenkowski: That was my other thought for propagating this information.  I think it's actually more natural but requires thinking about this for all encoders when only the minority will change sequence length.  It also constitutes an API change.  If we want to support more different kinds of encoders, it could be worth that change.

sockeye/encoder.py
mjdenkowski: Yes

sockeye/encoder.py
mjdenkowski: That's to be clear about each step.  The note that we're flattening the symbol could move to a comment.

sockeye/arguments.py
fhieber: Agreed, lets have option 2 (that is create some presets) first and gradually allow more configuration. I am currently working on the hierarchical config which should make things a bit easier.

sockeye/encoder.py
fhieber: I think in the interest of not having too many dependencies between member functions that one needs to call, we should do this. We will have a backwards incompatible change with the configuration update anyway and need to bump the version. So this might be a good time for API changes as well.

mjdenkowski:
I updated the arguments and have get_encoder to take a ModelConfig instead of a bunch of args.  This should be easy to update further when we finish splitting up config objects as in Felix's branch.

It looks like the consensus is that having encoders return triples of (data, lengths, seq_len), effectively transformations of the three inputs, is a good idea.  More commits to follow.

mjdenkowski:
The encode() method now always returns a triple corresponding to the encoded versions of the inputs: data, lengths, seq_len.  I ran some tests on toy data to make sure perplexity and output were identical to the master branch when not using the character sequence encoder (length propagation effectively a no-op with standard encoders).

This branch should now be a candidate for merging.

sockeye/arguments.py
```diff
@@ -148,11 +148,43 @@ def add_model_parameters(params):
...
+                              type=int_greater_or_equal(0),
+                              default=None,
+                              help='Maximum target vocabulary size. Overrides --num-words. Default: %(default)s')
     model_params.add_argument('--word-min-count',
                               type=int_greater_or_equal(1),
                               default=1,
                               help='Minimum frequency of words to be included in vocabularies. Default: %(default)s.')
 
+    model_params.add_argument('--encoder',
+                              choices=[C.RNN_NAME, C.RNN_WITH_CONV_EMBED_NAME],
```
fhieber: Similar to rnn types we could define that list of available encoders globally in constants.py

sockeye/encoder.py
```diff
@@ -108,21 +113,33 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
-        Encodes data given sequence lengths of individual examples (data_length) and maximum sequence length (seq_len).
+        Encodes data given sequence lengths of individual examples and maximum sequence length.
 
         :param data: Input data.
         :param data_length: Vector with sequence lengths.
         :param seq_len: Maximum sequence length.
-        :return: Encoded input data.
+        :return: Encoded versions of input data (data, data_length, seq_len).
         """
+        return (self._encode(data), data_length, seq_len)
```
fhieber: parentheses not needed

sockeye/encoder.py
```diff
@@ -172,7 +192,7 @@ def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int)
                                      name=self.prefix + 'embed')
         if self.dropout > 0:
             embedding = mx.sym.Dropout(data=embedding, p=self.dropout, name="source_embed_dropout")
-        return embedding
+        return (embedding, data_length, seq_len)
```
fhieber: parentheses not needed

sockeye/encoder.py
```diff
@@ -197,18 +217,21 @@ class EncoderSequence(Encoder):
...
         :param data_length: Vector with sequence lengths.
         :param seq_len: Maximum sequence length.
-        :return: Encoded input data.
+        :return: Encoded versions of input data (data, data_length, seq_len).
         """
         for encoder in self.encoders:
-            data = encoder.encode(data, data_length, seq_len)
-        return data
+            data, data_length, seq_len = encoder.encode(data, data_length, seq_len)
+        return (data, data_length, seq_len)
```
fhieber: parentheses not needed

sockeye/encoder.py
```diff
@@ -247,18 +278,21 @@ def __init__(self,
...
         :param data: Input data.
         :param data_length: Vector with sequence lengths.
         :param seq_len: Maximum sequence length.
-        :return: Encoded input data.
+        :return: Encoded versions of input data (data, data_length, seq_len).
         """
         outputs, _ = self.rnn.unroll(seq_len, inputs=data, merge_outputs=True, layout=self.layout)
 
-        return outputs
+        return (outputs, data_length, seq_len)
```
fhieber: parentheses not needed

sockeye/encoder.py
```diff
@@ -85,14 +87,17 @@ class Encoder:
     Generic encoder interface.
     """
 
-    def encode(self, data: mx.sym.Symbol, data_length: mx.sym.Symbol, seq_len: int) -> mx.sym.Symbol:
+    def encode(self,
+               data: mx.sym.Symbol,
+               data_length: mx.sym.Symbol,
+               seq_len: int) -> (mx.sym.Symbol, mx.sym.Symbol, int):
```
fhieber: is mypy happy with this typing or should we use:
`Tuple[mx.sym.Symbol, mx.sym.Symbol, int]` ?

(see also below)

sockeye/encoder.py
fhieber: please change to check_condition to make David happy for not using assertions to validate user input :)

sockeye/training.py
```diff
@@ -112,10 +112,12 @@ def sym_gen(seq_lens):
             """
             source_seq_len, target_seq_len = seq_lens
 
-            source_encoded = self.encoder.encode(source, source_length, seq_len=source_seq_len)
+            (source_encoded,
```
fhieber: parentheses not needed

sockeye/inference.py
```diff
@@ -97,16 +97,20 @@ def _build_modules(self):
         # Encoder symbol & module
         source = mx.sym.Variable(C.SOURCE_NAME)
         source_length = mx.sym.Variable(C.SOURCE_LENGTH_NAME)
+        source_encoded_length = mx.sym.Variable(C.SOURCE_ENCODED_LENGTH_NAME)
 
         def encoder_sym_gen(source_seq_len: int):
-            source_encoded = self.encoder.encode(source, source_length, seq_len=source_seq_len)
+            nonlocal source_encoded_length
+            (source_encoded,
```
fhieber: parentheses not needed

sockeye/encoder.py
```diff
@@ -413,3 +453,171 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
         Returns a list of RNNCells used by this encoder.
         """
         return self.forward_rnn.get_rnn_cells() + self.reverse_rnn.get_rnn_cells()
+
+
+class ConvolutionalEmbeddingEncoder(Encoder):
```
fhieber: would it be possible to write a simple unit test for this layer? I.e. provide random inputs, bind the symbol after this encoder and check for output shapes given input shapes?

sockeye/inference.py
```diff
@@ -97,16 +97,20 @@ def _build_modules(self):
         # Encoder symbol & module
         source = mx.sym.Variable(C.SOURCE_NAME)
         source_length = mx.sym.Variable(C.SOURCE_LENGTH_NAME)
+        source_encoded_length = mx.sym.Variable(C.SOURCE_ENCODED_LENGTH_NAME)
```
fhieber: I am not sure I understand why you need to create another input to the graph here. You are overwriting source_encoded_length immediately below in the inner function (using nonlocal). As far as I can tell this variable symbol will not be part of the graph anyway so why not just set it to None?

sockeye/encoder.py
mjdenkowski: Updated

mjdenkowski:
The last commit adds a unit test and addresses some of the cosmetic comments.

fhieber:
Thanks for iterating!

PR #70, opened 2017-07-14
Updated return value of get_loss functions for correctness

fhieber:
Previously CrossEntropy and SmoothedCrossEntropy had different return types.
Their return values are grouped in training.py::121, which expects a list as input.

### DIFF
--- a/sockeye/loss.py
+++ b/sockeye/loss.py
@@ -74,7 +74,7 @@ class CrossEntropyLoss(Loss):
 
         :param logits: Shape: (batch_size * target_seq_len, target_vocab_size).
         :param labels: Shape: (batch_size * target_seq_len,).
-        :return: List of loss and softmax output symbols.
+        :return: List of loss symbol.
         """
         if self._normalize:
             normalization = "valid"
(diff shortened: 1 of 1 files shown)


sockeye/loss.py
```diff
@@ -68,24 +68,24 @@ class CrossEntropyLoss(Loss):
...
         :param labels: Shape: (batch_size * target_seq_len,).
-        :return: Loss and softmax output symbols.
+        :return: List of loss and softmax output symbols.
         """
         if self._normalize:
             normalization = "valid"
         else:
             normalization = "null"
-        return mx.sym.SoftmaxOutput(data=logits,
+        return [mx.sym.SoftmaxOutput(data=logits,
```
davvil: I cannot find any clear documentation on this. What does SoftmaxOutput return exactly? If it's a tuple, then here we are creating a list with a single (tuple) element.

fhieber: Its returning a single symbol

davvil: I thought so. Then the :return: specification is not correct, is it?

fhieber: updated

PR #72, opened 2017-07-14
Config

fhieber:
New Pull request with working configuration rewrite.

### DIFF
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.0.3'
+__version__ = '1.1.0'
--- a/sockeye/config.py
+++ b/sockeye/config.py
@@ -1,0 +1,86 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not
+# use this file except in compliance with the License. A copy of the License
+# is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+from typing import Any, Dict
+
+import inspect
+import jsonpickle
+
+
+class Metaconfig(type):
+    def __new__(mcs, name, bases, namespace, **kwargs):
+        new_cls = super(Metaconfig, mcs).__new__(mcs, name, bases, namespace, **kwargs)
+        user_init = new_cls.__init__
+        defaults = {}
+        for name, v in inspect.signature(user_init).parameters.items():
+            if v.default != inspect.Parameter.empty:
+                defaults[name] = v.default
+
+        def __init__(self, *args, **kwargs):
+            for name, value in defaults.items():
+                if name not in kwargs:
+                    kwargs[name] = value
+            for k, v in kwargs.items():
+                setattr(self, k, v)
+            self._frozen = False
+            # Normally we won't be doing anything here, but just to be sure
+            user_init(self, *args, **kwargs)
+
+        setattr(new_cls, '__init__', __init__)
+        return new_cls
+
+
+class Config(metaclass=Metaconfig):
+    """
+    Base configuration object that supports freezing of members and JSON (de-)serialization.
+    Actual Configuration should subclass this object.
+    """
+    def __setattr__(self, key, value):
+        if hasattr(self, '_frozen') and self._frozen:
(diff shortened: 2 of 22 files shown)


sockeye/config.py
```diff
@@ -0,0 +1,82 @@
...
+
+import yaml
+
+
+class Config(yaml.YAMLObject):
+    """
+    Base configuration object that supports freezing of members and JSON (de-)serialization.
+    Actual Configuration should subclass this object.
+    """
+    yaml_tag = u'!Config'
```
davvil: Do we need u here (and derived configs)? I thought strings are unicode by default in python3.

sockeye/config.py
```diff
@@ -0,0 +1,82 @@
...
+        object.__setattr__(self, key, value)
+
+    def freeze(self):
+        """
+        Freezes this Config object, disallowing modification or addition of any parameters.
+        """
+        if hasattr(self, '_frozen') and self._frozen:  # It's ok to freeze an already frozen config
+            return
+        self._frozen = True
+        for k, v in self.__dict__.items():
```
davvil: Minor: Here we may call us recursively. It is ok because of the freezing an already frozen config, but something to be aware of. We may add a k != "self". (I had to add it in the previous PR in one of the dark commits).

sockeye/config.py
```diff
@@ -0,0 +1,82 @@
...
+
+class Config(yaml.YAMLObject):
+    """
+    Base configuration object that supports freezing of members and JSON (de-)serialization.
+    Actual Configuration should subclass this object.
+    """
+    yaml_tag = u'!Config'
+
+    def __init__(self):
+        self._frozen = False
```
davvil: Use __add_frozen()? (See also next comment)

sockeye/config.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __setattr__(self, key, value):
+        if hasattr(self, '_frozen') and self._frozen:
```
davvil: If we use __add_frozen() consitently (see constructor), we may skip the hasattr() test.

sockeye/model.py
```diff
@@ -11,70 +11,50 @@
...
+    yaml_tag = u"!ModelConfig"
+
+    def __init__(self,
+                 max_seq_len: int,
+                 vocab_source_size: int,
+                 vocab_target_size: int,
+                 config_encoder: encoder.RecurrentEncoderConfig,
+                 config_decoder: decoder.RecurrentDecoderConfig,
+                 config_attention: attention.AttentionConfig,
+                 config_loss: loss.LossConfig,
```
davvil: This looks nice now :-)

sockeye/config.py
fhieber: good point. I think we can remove this

sockeye/config.py
fhieber: I saw that but didn't really get that would happen. You mean if a child config has a reference to the parent? Because just by going through __dict__.items() we won't call 'ourselves' here.

sockeye/config.py
fhieber: done

sockeye/config.py
```diff
@@ -18,65 +18,73 @@
...
-    yaml_tag = u'!Config'
+    yaml_tag = '!Config'
 
     def __init__(self):
-        self._frozen = False
+        self.__add_frozen()
 
     def __setattr__(self, key, value):
-        if hasattr(self, '_frozen') and self._frozen:
+        if hasattr(self, '_frozen') and getattr(self, '_frozen'):
```
davvil: It doesn't hurt, but I think we won't need hasattr(self, '_frozen') any more.

fhieber: actually we do :)

sockeye/model.py
```diff
@@ -11,70 +11,50 @@
...
-ModelConfig defines model parameters defined at training time which are relevant to model inference.
-Add new model parameters here. If you want backwards compatibility for models trained with code that did not
-contain these parameters, provide a reasonable default under default_values.
-"""
+
+class ModelConfig(Config):
+    """
+    ModelConfig defines model parameters defined at training time which are relevant to model inference.
+    Add new model parameters here. If you want backwards compatibility for models trained with code that did not
+    contain these parameters, provide a reasonable default under default_values.
```
tdomhan: default_values don't seem to exist anymore. How does one do backwards compatible changes in the future? Can you just add arguments with default values to the config classes?

sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.0.3'
+__version__ = '1.1.0'
```
tdomhan: I guess we'll bump again for the next MXnet version, right?

sockeye/model.py
fhieber: yes you can, but also we should start using versioning more in these cases.

sockeye/__init__.py
fhieber: yes, we will

PR #74, opened 2017-07-17
Added version checks

fhieber:


sockeye/utils.py
```diff
@@ -30,13 +30,57 @@
...
+    :param version: Given version.
+    """
+    code_version = parse_version(__version__)
+    given_version = parse_version(version)
+    check_condition(code_version[0] == given_version[0],
+                    "Given release version (%s) does not match release code version (%s)" % (version, __version__))
+    check_condition(code_version[1] == given_version[1],
+                    "Given major version (%s) does not match major code version (%s)" % (version, __version__))
+    if code_version[2] != given_version[2]:
+        logger.warning("Given minor version (%s) does not match code version (%s)", version, __version__)
```
tdomhan: Maybe we should just do a logger.warning or not even log anything, as a user shouldn't be concerned by a minor version difference.

fhieber: I think some indication in the log is useful. Do you want to reduce to logger.info or what do you mean by "just do a logger.warning" because that is already the case, right?

tdomhan: ah yeah, that's what I meant. I just mean that a warning 'A did not B' sounds concerning, while in reality a mismatch shouldn't be concerning.

We're already logging the current Sockeye version and with the logging statement of the model Sockeye version in inference.py I think it would be ok to leave out this warning or info message here.

fhieber: Ok, will remove

PR #83, opened 2017-07-21
Propagate encoded sequence length during inference

mjdenkowski:
Looks like something may have gotten out of sync.  Querying encoded sequence length here gets the ConvolutionalEmbeddingEncoder working again.

This might be a good time to think about some tiny end-to-end tests.  I'll open an issue for discussion.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -630,12 +630,18 @@ class Translator:
         :return List of lists of word ids, list of attentions, array of accumulated length-normalized
                 negative log-probs.
         """
+        # Length of encoded sequence (may differ from initial input length)
+        encoded_source_length = self.models[0].encoder.get_encoded_seq_len(bucket_key)
+        for m in self.models[1:]:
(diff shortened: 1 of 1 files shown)


sockeye/inference.py
```diff
@@ -630,12 +630,18 @@ def _beam_search(self,
         :return List of lists of word ids, list of attentions, array of accumulated length-normalized
                 negative log-probs.
         """
+        # Length of encoded sequence (may differ from initial input length)
+        encoded_source_length = self.models[0].encoder.get_encoded_seq_len(bucket_key)
+        for m in self.models[1:]:
+            utils.check_condition(m.get_encoded_seq_len(bucket_key) == encoded_source_length,
```
fhieber: you could use all() here:
```python
utils.check_condition(all(encoded_source_length == m.get_encoded_seq_len(bucket_key) for m in self.models), "models must agree...")
```

mjdenkowski: Good way to save a line!

PR #85, opened 2017-07-23
End-to-end testing with various model configurations

mjdenkowski:
This adds some system-level checks to pytest.  The test_train_translate() function allows various model configurations to be evaluated on a simple sequence copying task.  For each test case:
* Create train and dev files where lines are randomly generated sequences of digits.
* Train a model with the specified args.
* Use the model to translate the dev file.
* Check that dev perplexity and BLEU meet the specified thresholds.

The current set of tests runs in a couple of minutes on a single CPU and covers most of the model components in Sockeye.  We can easily add more tests as we add more components.

Open questions:
* These tests are pretty quick, but still slower than the near-instantaneous unit tests.  Do we want to let developers to select unit tests, system tests, or both?
* The perplexity and BLEU thresholds are set to be very forgiving given that we are using randomly generated data.  Do we want to use static test files so we can tighten the bounds on expected perplexity and BLEU?


test/system/test_train_translate.py
```diff
@@ -0,0 +1,121 @@
...
+     " --conv-embed-pool-stride 2 --conv-embed-num-highway-layers 1 --rnn-num-layers 1 --rnn-cell-type lstm"
+     " --rnn-num-hidden 64 --num-embed 8 --attention-num-hidden 32 --batch-size 8 --loss cross-entropy"
+     " --optimized-metric perplexity --max-updates 2500 --checkpoint-frequency 2500 --optimizer adam"
+     " --initial-learning-rate 0.01",
+     "--beam-size 5",
+     2.0,
+     0.65),
+])
+
+def test_train_translate(train_params, translate_params, perplexity_thresh, bleu_thresh):
```
fhieber: can we rename to something like test_integration_seqcopy to denote the task a little bit clearer?
I'd also would like to add a sequence sorting task where source is randomly drawn numbers and target is the sorted sequence.

fhieber:
Thanks Michael!

I'd say the unit tests should be super quick, so this is more of an integration test that should run on travis. However, we could run these integration tests also locally from the commandline by adding a setup.py command. Unit tests can already be run with ``python setup.py test(s)``. Docs are created with ``python setup.py doc``. So we could add something like: ``python setup.py integration_test(s)``.

Sequence copying should be easily achievable by all these architectures and usually you reach a ppl of 1 fairly quickly. I think the question is more on how much time we want to allocate. Imo, an integration test is fine to run for 5-10 minutes, so allowing a few more updates to tighten the bounds should be ok.

mjdenkowski:
I really like the idea of more challenging tests that run on commit.  Here's another option for splitting up tests: we could have the existing unit tests, then integration and system tests.

Integration tests could run alongside unit tests with the sole purpose of making sure components can call each other and pass data without throwing exceptions.  For example, if we ignore perplexity and BLEU, we can train a model for 10 batches and translate 10 sentences on 1 CPU in about a second.  Here we could try to cover command line options exhaustively.

System tests where we do care about perplexity and BLEU could run on commit (and pre-commit.sh).  This could be digit reordering or even very limited natural language translation.  Here we would test specific types of models we want to perform well.

As a developer, I like unit and integration tests to immediately (<10 seconds) answer the question "Did my latest commit break anything?".  I like system tests to make sure the full system works as expected.

I'm going to rework these tests a bit and update the PR.

mjdenkowski:
The last commit divides things into 3 test categories:

**Unit**: Existing unit tests that run with `python setup.py test`

**Integration**: End-to-end tests running a few batches and decoding with configurations that cover many command line options.  These also run with `python setup.py test` and everything takes about 5 seconds on a modern laptop.

**System**: End-to-end tests where we expect perplexity and BLEU score to be near 1.  This currently includes sequence copying and sequence sorting with a small vanilla model.  These run as part of the pre-commit script and Travis CI.  They take about 5 minutes on a laptop and dramatically longer on Travis (1-2 hours?).  We can either trim them down or decide that's fine since developers can run 100% of the tests locally and we don't expect Travis to fail.

test/integration/test_seq_copy_int.py
```diff
@@ -0,0 +1,65 @@
...
+    with TemporaryDirectory(prefix="test_seq_copy") as work_dir:
+        # Simple digits files for train/dev data
+        train_source_path = os.path.join(work_dir, "train.src")
+        train_target_path = os.path.join(work_dir, "train.tgt")
+        dev_source_path = os.path.join(work_dir, "dev.src")
+        dev_target_path = os.path.join(work_dir, "dev.tgt")
+        generate_digits_file(train_source_path, train_target_path, _TRAIN_LINE_COUNT, _LINE_MAX_LENGTH)
+        generate_digits_file(dev_source_path, dev_target_path, _DEV_LINE_COUNT, _LINE_MAX_LENGTH)
+        # Test model configuration
+        _, _ = run_train_translate(train_params,
```
tdomhan: why assign to "_" if the return values are anyway not used? (instead of not assigning to anything)

tdomhan:
Does Travis actually time out or does it just take annoying long?

mjdenkowski:
I stopped it when it was about 33% finished after an hour.  We could just let it run, but then there would be a multiple hour time lag on testing commits.

mjdenkowski:
Any objections to merging the updated version and then iterating?  I'll merge in 24 hours if no one raises any issues.

fhieber:
No blockers from my side. I agree that using the "_"s is a bit awkward but no blocker.

tdomhan:
no blockers from my side either. I agree that we shouldn't run the system tests on Travis if they take that long and that we can iterate in order to make them run quickly enough in the future, or then find a solution other than Travis.

test/common.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    length = random.randint(1, max_len)
+    # Due to the special words, the actual words start at index 3 and go up to vocab_size+2
+    return [random.randint(3, vocab_size + 2) for _ in range(length)]
+
+
+_DIGITS = "0123456789"
+
+
+def generate_digits_file(source_path: str,
```
tdomhan: this method and also run_train_translate seem to be only relevant to the system tests, so I'd consider moving it in the system module.

PR #87, opened 2017-07-25
Minor updates to attention.py

fhieber:
* small refactor of attention methods
* added dot product scaling to DotAttention [see Vaswani et al, 2017].

### DIFF
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.1.0'
+__version__ = '1.1.1'
--- a/sockeye/attention.py
+++ b/sockeye/attention.py
@@ -310,6 +310,9 @@ class DotAttention(Attention):
                                               num_hidden=self.num_hidden,
                                               no_bias=True, name="%squery_hidden_fc" % self.prefix)
 
+            # scale down dot product by sqrt(num_hidden) [Vaswani et al, 17]
+            query *= self.num_hidden ** -0.5
+
             # (batch_size, decoder_num_hidden, 1)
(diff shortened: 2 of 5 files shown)


sockeye/attention.py
```diff
@@ -310,6 +310,9 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
                                               num_hidden=self.num_hidden,
                                               no_bias=True, name="%squery_hidden_fc" % self.prefix)
 
+            # scale down dot product by sqrt(num_hidden) [Vaswani et al, 17]
+            query *= self.num_hidden ** -0.5
```
tdomhan: I wouldn't make this change backwards-incompatible, but rather add another parameter, or represent this as a different attention type like 'dot_scaled' (which could internally map to a parameter passed to DotAttention). Otherwise models that have been trained without the change will just silently suddenly compute a different attention score.

sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.1.0'
+__version__ = '1.1.1'
```
tdomhan: we don't do full semantic versioning (a la http://semver.org/), but the lowest version number should still  be reserved for backwards-compatible changes. Just like with the backwards-incompatible change to the config change we should bump at least the second number if we don't make the change backwards compatible. That said, I'd rather just make the change backwards compatible, in order to not invalidate trained models.

sockeye/attention.py
fhieber: ah yes, you are right I haven't thought about backwards compatibility here. I'll expose another type dot_scaled that maps to a parameter passed to DotAttention.

tdomhan: thanks for making the change!

PR #91, opened 2017-07-26
First stab at weights & tensorboard monitoring. For now, only logs statistics.

fhieber:
MXNet's monitor behavior isn't really straightforward, so I have not integrated logging to tensorboard yet. However, this enables an easy way to track details from the graph during training.

### DIFF
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -14,6 +14,7 @@
 """
 Defines various constants used througout the project
 """
+import mxnet as mx
 
 BOS_SYMBOL = "<s>"
 EOS_SYMBOL = "</s>"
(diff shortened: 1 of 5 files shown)


sockeye/constants.py
```diff
@@ -88,6 +88,10 @@
 
 MEASURE_SPEED_EVERY = 50  # measure speed and metrics every X batches
 
+# Monitor constants
+STAT_FUNC_NORM = "norm"  # default MXNet monitor stat func: |x|/size(x)
+MONITOR_STAT_FUNCS = {STAT_FUNC_NORM: None}
```
tdomhan: why does it map to None?

tdomhan: I guess it's for implicitly using the mxnet default defined inside of Monitor. What if mxnet decides to change the norm? Would it maybe make more sense to all the option "default" or "mx_default"?
```
STAT_FUNC_DEFAULT = "mx_default"  # default MXNet monitor stat func: |x|/size(x)
MONITOR_STAT_FUNCS = {STAT_FUNC_DEFAULT: None, STAT_FUNC_NORM: some_other_func}
```

PR #94, opened 2017-07-31
Added option to build a single joint vocab from source and target.

fhieber:


sockeye/arguments.py
```diff
@@ -160,6 +160,10 @@ def add_model_parameters(params):
                               type=int_greater_or_equal(1),
                               default=1,
                               help='Minimum frequency of words to be included in vocabularies. Default: %(default)s.')
+    model_params.add_argument('--joint-vocab',
```
tdomhan: So the only reason for a joint vocab would be if we want to use a shared embedding, right? How about having a parameter that determines if we tie source and target embeddings and if that parameters is True we internally build the shared vocab. What do you think?

tdomhan: This is what I meant: https://github.com/awslabs/sockeye/pull/96

fhieber:
Closing in favor of #96

PR #96, opened 2017-08-01
Adding different types of weight tying.

tdomhan:
Based on Felix's PR: https://github.com/awslabs/sockeye/pull/94 

Is fully backwards compatible to pervious sockeye versions.

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -261,7 +261,15 @@ def add_model_parameters(params):
 
     model_params.add_argument('--weight-tying',
                               action='store_true',
-                              help='Share target embedding and output layer parameter matrix. Default: %(default)s.')
+                              help='Turn on weight tying. The type of weight sharing is determined through '
+                                   '--weight-tying-type. Default: %(default)s.')
+    model_params.add_argument('--weight-tying-type',
+                              default=C.WEIGHT_TYING_TRG_SOFTMAX,
+                              choices=[C.WEIGHT_TYING_SRC_TRG_SOFTMAX,
+                                       C.WEIGHT_TYING_SRC_TRG,
+                                       C.WEIGHT_TYING_TRG_SOFTMAX],
+                              help='The type of weight tying. source embeddings=src, target embeddings=trg, '
+                                   'target softmax weight matrix=softmax. Default: %(default)s.')
 
     model_params.add_argument('--max-seq-len',
                               type=int_greater_or_equal(1),
--- a/sockeye/config.py
+++ b/sockeye/config.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 
 import copy
+import inspect
 
 import yaml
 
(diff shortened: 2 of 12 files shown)


tdomhan:
I removed the --joint-vocab flag and set it implicitly in train, but if we want to have the flag for some other use case I'm happy to add it back.

sockeye/model.py
```diff
@@ -68,6 +74,10 @@ def __init__(self,
         self.config_loss = config_loss
         self.lexical_bias = lexical_bias
         self.learn_lexical_bias = learn_lexical_bias
+        self.weight_tying = weight_tying
+        if weight_tying and weight_tying_type is None:
```
fhieber: can that ever be the case with the default argument in the constructor above?

sockeye/arguments.py
```diff
@@ -249,7 +249,15 @@ def add_model_parameters(params):
 
     model_params.add_argument('--weight-tying',
```
fhieber: could we rename this to 'weight-sharing' everywhere? I don't like the 'tying' expression that much anymore and its a bit fuzzy on what it means.

If this breaks backward compatibility, lets just bump the version

sockeye/config.py
```diff
@@ -37,6 +38,18 @@ def __setattr__(self, key, value):
             raise AttributeError("Cannot set self as attribute")
         object.__setattr__(self, key, value)
 
+    def __setstate__(self, state):
```
fhieber: can you add a test for that?

sockeye/model.py
```diff
@@ -45,6 +47,8 @@ class ModelConfig(Config):
     :param config_loss: Loss configuration.
     :param lexical_bias: Use lexical biases.
     :param learn_lexical_bias: Learn lexical biases during training.
+    :param weight_tying: Enables weight tying if True.
```
fhieber: I am not too happy with again adding more parameters to the model config here but I guess there is no other way...

tdomhan: and I guess this is the right place to put them no? As it affects both encoder and decoder.

With respect to the name: we already have a parameter called weight_tying in the decoder config. I think it would be confusing to call it weight_tying in one place and weight_sharing in another. At the same time I don't want to necessarily break backwards compatibility.

sockeye/model.py
tdomhan: only if someone specifically passes None, which is actually what I'm doing from train.py. The reason is that it looks slightly confusing when the model/config says:
```
weight_tying: False
weight_tying_type: trg_softmax
```
When setting this to None in case of `weight_tying=False` we get:
```
weight_tying: False
weight_tying_type: null
```
Anyway, just a minor detail. So it can happen.

sockeye/config.py
tdomhan: sure

sockeye/model.py
fhieber: Thats what I meant: also renaming the weight-tying parameter in the decoder. I think this would be the right thing to do and backwards compatibility should not stop us from doing this.

sockeye/arguments.py
tdomhan: what's wrong with tying? After all that's also what's it called in "Using the Output Embedding to Improve Language Models". Is this really a reason to break backwards compatibility?

fhieber: To me 'tying' has a weaker implication (e.g. it could mean that the parameters should be made 'similar'), whereas 'sharing' clearly says that these are the same set of parameters.
but if you feel strongly about backwards compatibility we can leave as is.

PR #98, opened 2017-08-02
Transformer models in Sockeye

fhieber:
This PR adds support for transformer encoders & decoders in Sockeye. 
The transformer architecture was proposed in Vaswani et al 2017, Attention is all you need.

From a high-level, users control the choice of encoder and decoder architectures through the --{encoder,decoder} flags.
Integration tests are added for new architecture types.


tdomhan:
Looking at the Travis log it looks like py3.4 does not support expressions like `[softmax, attention_probs, *states]`, but rather needs something like `[softmax, attention_probs] + states`.

sockeye/lr_scheduler.py
```diff
@@ -128,7 +154,8 @@ def __call__(self, t):
         if self.lr is None:
             assert self.base_lr is not None
             self.lr = self.base_lr
-        return self.lr
+        lr = self._warmup(t) if self.warmup > 0 and t <= self.warmup else self.lr
```
tdomhan: this might lead to confusing logs if we reduce self.lr in the first steps, as we will write to the log that the learning rate got reduce, while we're still ramping up.

tdomhan:
This has become a gigantic feature branch :) How are we going to bring this into master? Is this still squashable, given that the branch contains regular merges from master?

sockeye/arguments.py
```diff
@@ -215,6 +230,28 @@ def add_model_parameters(params):
                               help="Add residual connections to stacked RNNs if --rnn-num-layers > 3. "
                                    "(see Wu ETAL'16). Default: %(default)s.")
 
+    # transformer arguments
+    model_params.add_argument('--transformer-model-size',
+                              type=int_greater_or_equal(1),
+                              default=512,
+                              help='Size of all layers and embeddings when using transformer. Default: %(default)s.')
+    model_params.add_argument('--transformer-num-layers',
```
fhieber: but we wouldn't be able to specify m layers for a transformer encoder, but only n layers for an RNN decoder. I'd like to keep this separate if possible.

sockeye/arguments.py
```diff
@@ -248,7 +285,10 @@ def add_model_parameters(params):
     model_params.add_argument('--attention-coverage-num-hidden',
                               type=int,
                               default=1,
-                              help="Number of hidden units for coverage vectors. Default: %(default)s")
+                              help="Number of hidden units for coverage vectors. Default: %(default)s.")
+    model_params.add_argument('--attention-mhdot-heads',
```
fhieber: because in theory you could have a self-attending transformer encoder with x heads, but an RNN decoder using multi-head attention with y heads.

sockeye/arguments.py
```diff
@@ -420,7 +472,12 @@ def add_training_args(params):
                               type=float,
                               default=10,
                               help="Half-life of learning rate in checkpoints. For 'fixed-rate-*' "
-                                   "learning rate schedulers. Default: 10.")
+                                   "learning rate schedulers. Default: %(default)s.")
+    train_params.add_argument('--learning-rate-warmup',
+                              type=float,
```
fhieber: true, thanks

sockeye/arguments.py
```diff
@@ -420,7 +472,12 @@ def add_training_args(params):
...
+                              default=0,
+                              help="Number of warmup steps. For 'fixed-rate-*' "
```
fhieber: fixed

sockeye/lr_scheduler.py
fhieber: can you give more detail, I am not sure I see what could happen?

fhieber:
We can either merge without squashing, or squash into a single commit (to avoid redoing the merges I had to do before). Lets try to get this in sooner than later to avoid growing this branch even more. 
We can still make incremental updates to master after merging this.

sockeye/arguments.py
davvil: --encoder-num-layers and --decoder-num-layers, then?

sockeye/arguments.py
tdomhan: ahh ok, got it. Could you maybe add the word `self` to the other attention argument or make it clear in the help string that the other argument is about self-attention. It all makes sense now, but initially looking at the parameters I didn't realize that one param was for the attention mechanism from enc to dec and the other for self-attention.  Maybe:
```
model_params.add_argument('--transformer-attention-heads',
                              type=int_greater_or_equal(1),
                              default=8,
                              help='Number of heads for all self-attention when using transformer layers. Default: %(default)s.')
```

tdomhan:
yeah, sure. I'm all for merging this asap :)

sockeye/lr_scheduler.py
tdomhan: if the learning rate is reduced at any step `t < warmup` because the validation score plateaued you get the log
```
%d checkpoints since improvement or rate scaling, lowering learning rate: %1.2e -> %1.2e
```
which technically would not be true as the learning rate would still be coming from self._warmup rather then being self.lr, which could be slightly confusing. That's all I mean.

sockeye/attention.py
```diff
@@ -315,6 +323,7 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
                                               weight=self.t2h_weight,
                                               num_hidden=self.num_hidden,
                                               no_bias=True, name="%squery_hidden_fc" % self.prefix)
+            query *= self.num_hidden ** -0.5
```
mjdenkowski: There's also a statement below to scale the query.  Is this applying scaling twice?

sockeye/arguments.py
fhieber: That would be a possibility. But all of our arguments are currently 'tied' to the type of model. we have attention-num-hidden, rnn-num-hidden and all the conv-seq-embed arguments. On the other hand, the number of layers tends to be a fairly general concept across models, so I will make the change.

sockeye/arguments.py
fhieber: fixed

sockeye/lr_scheduler.py
fhieber: Ah I see. I don't have a good idea how to fix this though.

sockeye/attention.py
fhieber: Good catch! I missed removing this when I added the parameter, thanks

sockeye/lr_scheduler.py
tdomhan: I don't have anything elegant, but a t>warmup around the log would work.

sockeye/lr_scheduler.py
```diff
@@ -34,21 +55,23 @@ class LearningRateSchedulerInvSqrtT(LearningRateScheduler):
...
         # then factor = 3 ./ T, with T = half_life * updates_per_checkpoint
         self.factor = 3. / (half_life * updates_per_checkpoint)
+        self.warmup = warmup
         self.t_last_log = -1
         self.log_every_t = int(half_life * updates_per_checkpoint)
 
     def __call__(self, num_updates: int):
-        lr = self.base_lr / sqrt(1 + num_updates * self.factor)
+        change = min(1.0 / sqrt(1 + num_updates * self.factor), self._warmup(num_updates) if self.warmup > 0 else 99999)
+        lr = self.base_lr * change
```
tdomhan: Somehow can't see my other comment, but warmup should either return a fraction (instead of a learning rate) or the result here shouldn't be multiplied by base_lr. Otherwise you multiply by base_lr twice.

fhieber: Ha, good catch. I changed it

sockeye/lr_scheduler.py
fhieber: not really, since we dont have a t in new_evaluation_result() :)
I can add add a flag 'warmed_up' written to by __call__; its not very elegant though :/

sockeye/arguments.py
```diff
@@ -214,7 +241,30 @@ def add_model_parameters(params):
                               default=False,
                               help="Add residual connections to stacked RNNs if --rnn-num-layers > 3. "
                                    "(see Wu ETAL'16). Default: %(default)s.")
+    model_params.add_argument('--rnn-context-gating', action="store_true",
+                              help="Enables a context gate which adaptively weighs the RNN decoder input against the"
+                                   "source context vector before each update of the decoder hidden state.")
```
tdomhan: space missing between this and the previous sentence ;)

sockeye/arguments.py
```diff
@@ -214,7 +241,30 @@ def add_model_parameters(params):
...
+
+    # transformer arguments
+    model_params.add_argument('--transformer-model-size',
```
tdomhan: I don't have  a strong preference, but this sounds like the equivalent to "--rnn-num-hidden". Would it make sense to call it similarly? Like --transformer-num-hidden. What do you think?

sockeye/attention.py
```diff
@@ -37,21 +38,24 @@ class AttentionConfig(config.Config):
...
     """
     def __init__(self,
                  type: str,
                  num_hidden: int,
                  input_previous_word: bool,
                  rnn_num_hidden: int,
                  layer_normalization: bool,
-                 config_coverage: Optional[coverage.CoverageConfig] = None) -> None:
+                 config_coverage: Optional[coverage.CoverageConfig] = None,
+                 num_heads: int = 1) -> None:
```
tdomhan: should this be an optional parameter, if it's not always used?

sockeye/layers.py
```diff
@@ -58,18 +61,376 @@ def moments(inputs: mx.sym.Symbol) -> Tuple[mx.sym.Symbol, mx.sym.Symbol]:
...
+        logits = mx.sym.broadcast_add(logits, bias)
+
+    probs = mx.sym.softmax(logits, axis=-1)
+    probs = mx.sym.Dropout(probs, p=dropout) if dropout > 0.0 else probs
+
+    # (n, lq, lk) x (n, lk, dv) -> (n, lq, dv)
+    return mx.sym.batch_dot(lhs=probs, rhs=values)
+
+
+class MultiHeadAttentionBase:
```
tdomhan: would it make sense to add the word "self" to this class as well?

sockeye/attention.py
```diff
@@ -336,6 +344,117 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
         return attend
 
 
+class MultiHeadDotAttention(Attention):
+    """
+    Dot product attention with multiple heads as proposed in Vaswani et al, Attention is all you need.
+    Can be used with a RecurrentDecoder.
+
+    :param input_previous_word: Feed the previous target embedding into the attention mechanism.
+    :param depth: Number of hidden units.
```
tdomhan: a comment on terminology: I find the use of "depth" confusing here, as it makes one think of the number of layers, rather than a number of hidden units. I'd rather call it num_hidden or num_hidden_per_head (just a suggestion).

sockeye/attention.py
```diff
@@ -336,6 +344,117 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
...
+        AttentionState = attend(AttentionInput, AttentionState).
+
+        :param source: Shape: (batch_size, seq_len, encoder_num_hidden).
+        :param source_length: Shape: (batch_size,).
+        :param source_seq_len: Maximum length of source sequences.
+        :return: Attention callable.
+        """
+
+        # Combine batch and time dimension
+        # (batch * length, input_depth)
```
tdomhan: thanks for adding all the dimensions, makes it much clearer :)

sockeye/layers.py
```diff
@@ -58,18 +61,376 @@ def moments(inputs: mx.sym.Symbol) -> Tuple[mx.sym.Symbol, mx.sym.Symbol]:
...
+    x = mx.sym.reshape(data=x, shape=(-4, -1, heads, length, 0))
+    # (batch, length, heads, depth_per_head)
+    x = mx.sym.transpose(x, axes=(0, 2, 1, 3))
+    # (batch, length, depth)
+    return mx.sym.reshape(x, shape=(-1, length, -3))
+
+
+def broadcast_lengths(x: mx.sym.Symbol, heads: int) -> mx.sym.Symbol:
+    """
+    Broadcasts the length information of each sequence to multiple heads.
```
tdomhan: I fail to see how this is related to "length". So you basically broadcast according to the number of heads. Where does length come in?

sockeye/arguments.py
fhieber: fixed

sockeye/arguments.py
fhieber: I'd prefer model size. Its closer to what the paper describes and its a little bit more used throughout different blocks in the transformer.

sockeye/attention.py
fhieber: yes, will change

sockeye/attention.py
fhieber: ok I can change it here, but depth is used in the transformer code so I will leave it there

sockeye/layers.py
fhieber: True, the name is a bit misleading. The method is only used where input x is the length information but the method is more general. will rename

sockeye/layers.py
fhieber: no because also used for the non-self attention. Self attention is a special case of MultiHeadAttention where the queries == keys/values

sockeye/arguments.py
tdomhan: that is true. I thought more regarding consistence. Anyway, no strong feelings about this ;)

sockeye/attention.py
tdomhan: I guess if we have it in multiple locations we should stay consistent. anyway, just a comment. feel free to leave it at depth ;)

mjdenkowski:
Have we run end-to-end systems with RNN and transformer for the current state of the branch?  I started a RNN run today but someone else may have already finished one.  Felix can probably comment best on the transformer results.

If we get the expected BLEU scores for RNN and something in the ballpark for transformer, I think we're ready to merge everything into master.  We can continue to iterate on hyper-parameter settings for transformer and training.

fhieber:
I have one finished for RNNs with this branch, produces the same result as master, down to the last digit in ppl.
Still waiting on transformer results. If you are fine with the current state, approve and I will merge once Tobi had a final look and I have transformer results.

sockeye/arguments.py
```diff
@@ -175,7 +175,32 @@ def add_model_parameters(params):
                               choices=C.ENCODERS,
                               default=C.RNN_NAME,
                               help="Type of encoder. Default: %(default)s.")
+    model_params.add_argument('--decoder',
+                              choices=C.DECODERS,
+                              default=C.RNN_NAME,
+                              help="Type of encoder. Default: %(default)s.")
+
+    model_params.add_argument('--num-layers',
```
fhieber: Not a bad idea. We could extend that to the names for encoder/decoder flags as well, e.g. --model rnn:transformer or transformer.
Equally for num_embed: 512:1024, vocab-size etc etc.
I think this would be a nice change but out of scope of this PR.
I will leave as is for now.

sockeye/decoder.py
```diff
@@ -14,85 +14,123 @@
...
 
-    :param vocab_size: Target vocabulary size.
-    :param num_embed: Target word embedding size.
-    :param rnn_config: RNN configuration.
-    :param dropout: Dropout probability for decoder RNN.
-    :param weight_tying: Whether to share embedding and prediction parameter matrices.
-    :param context_gating: Whether to use context gating.
-    :param layer_normalization: Apply layer normalization.
+
+class Decoder:
```
fhieber: Good idea

sockeye/decoder.py
```diff
@@ -14,85 +14,123 @@
...
+    """
+    Generic decoder interface.
+    A decoder needs to provide methods to provide training (decode) and inference symbols (predict).
```
fhieber: I will change the wording to make this a bit clearer, fully agreed that this is not specific to inference/search

sockeye/decoder.py
```diff
@@ -105,51 +143,291 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
         raise NotImplementedError()
 
 
-DecoderState = NamedTuple('DecoderState', [
+class TransformerDecoder(Decoder):
+    """
+    Transformer decoder as in Vaswani et al, 2017: Attention is all you need.
+    In training, computation scores for each position of the known target sequence are compouted in parallel,
+    yielding most of the speedup.
+    At inference time, the decoder block is evaluated again and again over a maximum length input sequence that is
```
fhieber: That was my first idea too. Its not trivial so I want to leave this out of scope of this PR. But obviously there are a couple of things that can be done to speed this up.
Overall its actually not that slow on a GPU :)

sockeye/decoder.py
```diff
@@ -105,51 +143,291 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
+            embed_weight = mx.sym.Variable(C.TARGET_EMBEDDING_PREFIX + "weight")
+
+        self.embedding = encoder.Embedding(num_embed=config.model_size,
+                                           vocab_size=config.vocab_size,
+                                           prefix=C.TARGET_EMBEDDING_PREFIX,
+                                           dropout=config.dropout_residual,
+                                           embed_weight=embed_weight,
+                                           add_positional_encoding=config.positional_encodings)
+        if self.config.weight_tying:
+            logger.info("Tying the target embeddings and prediction matrix.")
```
fhieber: no, as the transformer always uses embedding_size == model_size/num_hidden.

sockeye/transformer.py
```diff
@@ -0,0 +1,298 @@
...
+                 num_hidden: int,
+                 layer_normalization: bool,
+                 dropout: float,
+                 prefix: str) -> None:
+        self.num_hidden = num_hidden
+        self.dropout = dropout
+        self.layer_norm = None
+        self.prefix = prefix
+        if layer_normalization:
+            self.layer_norm = layers.LayerNormalization(num_hidden=self.num_hidden, prefix=self.prefix)
```
fhieber: I don't feel strongly about it. I looked a couple of times at the parameter dicts from the transformer and found this a bit more concise and shorter.

sockeye/transformer.py
```diff
@@ -0,0 +1,298 @@
...
+        self.assign(out_data[0], req[0], self.bias)
+
+    def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
+        pass
+
+
+@mx.operator.register("auto_regressive_bias")
+class AutoRegressiveBiasProp(mx.operator.CustomOpProp):
+
+    def __init__(self, length: str):
```
fhieber: yes, thats one of the quite obscure limitations of passing arguments to python custom ops.

sockeye/encoder.py
```diff
@@ -87,6 +105,32 @@ def get_recurrent_encoder(config: RecurrentEncoderConfig, fused: bool,
...
+                            embed_weight: Optional[mx.sym.Symbol] = None) -> 'Encoder':
+    """
+    Returns a Transformer encoder, consisting of an embedding layer with
+    positional encodings and a TransformerEncoder instance.
+
+    :param config: Configuration for transformer encoder.
+    :param embed_weight: Optionally use an existing embedding matrix instead of creating a new one.
+    :return: Encoder instance.
+    """
+    encoders = list()
```
fhieber: PyCharm prefers "list()" over [] ("i.e. could be written as a list literal")

sockeye/encoder.py
```diff
@@ -458,31 +533,93 @@ def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
...
 
 
+class TransformerEncoder(Encoder):
+    """
+    Non-recurrent encoder based on the transformer architecture in:
+
+    Attention Is All You Need, Figure 1 (left)
+    Vaswani et al. (https://arxiv.org/pdf/1706.03762.pdf)
+
+    :param config: TransformerEncoderConfig
```
fhieber: added

sockeye/transformer.py
tdomhan: interesting. Well I guess that's something we need to live with.

sockeye/arguments.py
tdomhan: sounds good

test/unit/test_callback.py
```diff
@@ -58,12 +63,23 @@ def test_callback(optimized_metric, initial_best, train_metrics, eval_metrics, i
...
-            
+            assert os.path.exists(metrics_fname)
+            metrics = utils.read_metrics_file(metrics_fname)
+            _compare_metrics(metrics, monitor.metrics)
+
+
+def _compare_metrics(a, b):
+    assert len(a) == len(b)
+    for x, y in zip(a, b):
+        for (xk, xv), (yk, yv) in zip(sorted(x.items()), sorted(y.items())):
```
tdomhan: you could add another assert on length of `x.items()` here, just in case.

sockeye/lr_scheduler.py
```diff
@@ -34,22 +55,24 @@ class LearningRateSchedulerInvSqrtT(LearningRateScheduler):
...
-        self.base_lr = None
         # 0.5 base_lr = base_lr * sqrt(1 + T * factor)
         # then factor = 3 ./ T, with T = half_life * updates_per_checkpoint
         self.factor = 3. / (half_life * updates_per_checkpoint)
         self.t_last_log = -1
         self.log_every_t = int(half_life * updates_per_checkpoint)
 
-    def __call__(self, num_updates: int):
-        lr = self.base_lr / sqrt(1 + num_updates * self.factor)
+    def _update_rate(self, num_updates):
```
tdomhan: from where is this method called?

PR #99, opened 2017-08-02
Simplified data iterator. Removed source_length

fhieber:
source_length can be easily computed in the computation graph. This simplifies the data iterator quite a bit. 
Once we merged the transformer branch I will also remove the dependency of having source_length as an input argument to the inference encoder graph.

### DIFF
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -51,7 +51,7 @@ def define_buckets(max_seq_len: int, step=10) -> List[int]:
 def define_parallel_buckets(max_seq_len_source: int,
                             max_seq_len_target: int,
                             bucket_width: int = 10,
-                            length_ratio:float = 1.0) -> List[Tuple[int, int]]:
+                            length_ratio: float = 1.0) -> List[Tuple[int, int]]:
     """
     Returns (source, target) buckets up to (max_seq_len_source, max_seq_len_target).  The longer side of the data uses
     steps of bucket_width while the shorter side uses steps scaled down by the average target/source length ratio.  If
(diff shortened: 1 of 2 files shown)


sockeye/data_io.py
```diff
@@ -50,8 +50,8 @@ def define_buckets(max_seq_len: int, step=10) -> List[int]:
 
 def define_parallel_buckets(max_seq_len_source: int,
                             max_seq_len_target: int,
-                            bucket_width=10,
-                            length_ratio=1.0) -> List[Tuple[int, int]]:
+                            bucket_width: int = 10,
+                            length_ratio:float = 1.0) -> List[Tuple[int, int]]:
```
tdomhan: nit-picky, but one line has a space between the colon and the type and the other doesn't.

sockeye/training.py
```diff
@@ -94,7 +94,7 @@ def _build_module(self,
         Initializes model components, creates training symbol and module, and binds it.
         """
         source = mx.sym.Variable(C.SOURCE_NAME)
-        source_length = mx.sym.Variable(C.SOURCE_LENGTH_NAME)
+        source_length = mx.sym.sum(mx.sym.broadcast_not_equal(source, mx.sym.zeros((1,))), axis=1)
```
tdomhan: how about adding an:
```
assert C.PAD_ID == 0
```
to catch issues if we ever changed the PAD_ID and to make it explicit that this is an assumption.

sockeye/data_io.py
fhieber: fixed

sockeye/training.py
fhieber: yes good point. We have that assertion somewhere in the data iterator as well

fhieber: added

fhieber:
no, I didn't see any performance differences locally. Its a fairly simple operation, compared to the unrolled RNNs its negligible I suppose.

PR #103, opened 2017-08-10
Added option to add attention in the (upper) layers of the decoder RNN

davvil:
It would still need some experimentation before merging to decide on some implementation details (e.g. last FF layer in the decoder).

### DIFF
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -480,7 +480,6 @@ class RecurrentDecoder(Decoder):
                                                           layers=range(1, self.rnn_config.num_layers))
         self.rnn_pre_attention_n_states = len(self.rnn_pre_attention.state_shape)
 
-
         if not self.config.zero_state_init:
             self._create_state_init_parameters()
 
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -16,7 +16,6 @@ Decoders for sequence-to-sequence models.
 """
 import logging
 from abc import ABC, abstractmethod
-import itertools
 from typing import Callable, List, NamedTuple, Tuple
 from typing import Optional
 
@@ -697,7 +696,7 @@ class RecurrentDecoder(Decoder):
                 mx.sym.Variable(C.SOURCE_LENGTH_NAME),
                 mx.sym.Variable(C.HIDDEN_PREVIOUS_NAME)] + \
                [mx.sym.Variable("%senc2decinit_%d" % (self.prefix, i)) for i in
-                range(len(list(itertools.chain(*[rnn.state_info for rnn in self.get_rnn_cells()]))))]
+                range(len(sum([rnn.state_info for rnn in self.get_rnn_cells()], [])))]
 
     def state_shapes(self,
                      batch_size: int,
@@ -727,7 +726,7 @@ class RecurrentDecoder(Decoder):
                [mx.io.DataDesc("%senc2decinit_%d" % (self.prefix, i),
                                (batch_size, num_hidden),
                                layout=C.BATCH_MAJOR) for i, (_, num_hidden) in enumerate(
-                                   itertools.chain(*[rnn.state_shape for rnn in self.get_rnn_cells()])
+                                   sum([rnn.state_shape for rnn in self.get_rnn_cells()], [])
                                )]
 
     def get_rnn_cells(self) -> List[mx.rnn.BaseRNNCell]:
@@ -765,8 +764,7 @@ class RecurrentDecoder(Decoder):
 
         # initial states for each layer
         layer_states = []
-        for state_idx, (_, init_num_hidden) in enumerate(itertools.chain(
-                *[rnn.state_shape for rnn in self.get_rnn_cells()])):
+        for state_idx, (_, init_num_hidden) in enumerate(sum([rnn.state_shape for rnn in self.get_rnn_cells()], [])):
             if self.config.zero_state_init:
                 init = mx.sym.tile(data=zeros, reps=(1, init_num_hidden))
             else:
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -456,6 +456,8 @@ class RecurrentDecoder(Decoder):
         self.num_hidden = self.rnn_config.num_hidden
 
         if self.config.context_gating:
+            utils.check_condition(not self.config.attention_in_upper_layers,
+                                  "Context gating is not supported with attention in upper layers.")
             self.gate_w = mx.sym.Variable("%sgate_weight" % prefix)
             self.gate_b = mx.sym.Variable("%sgate_bias" % prefix)
             self.mapped_rnn_output_w = mx.sym.Variable("%smapped_rnn_output_weight" % prefix)
@@ -818,6 +820,9 @@ class RecurrentDecoder(Decoder):
                                         state.layer_states[self.rnn_pre_attention_n_states:])
             hidden_concat = mx.sym.concat(upper_rnn_output, attention_state.context,
                                           dim=1, name='%shidden_concat_t%d' % (self.prefix, seq_idx))
+            if self.config.hidden_dropout > 0:
+                hidden_concat = mx.sym.Dropout(data=hidden_concat, p=self.config.hidden_dropout,
+                                               name='%shidden_concat_dropout_t%d' % (self.prefix, seq_idx))
             hidden = self._hidden_mlp(hidden_concat, seq_idx)
             # TODO: add context gating?
         else:
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -470,13 +470,14 @@ class RecurrentDecoder(Decoder):
 
         # Stacked RNN
         if self.rnn_config.num_layers == 1 or not self.rnn_config.attention_in_upper_layers:
-            self.attention_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, False)
-            self.upper_layers_rnn = None
+            self.rnn_pre_attention = rnn.get_stacked_rnn(self.rnn_config, self.prefix, parallel_inputs=False)
+            self.rnn_post_attention = None
         else:
-            self.attention_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, False, layers=[0])
-            self.upper_layers_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, True,
-                                                        layers=range(1, self.rnn_config.num_layers))
-        self.attention_rnn_n_states = len(self.attention_rnn.state_shape)
+            self.rnn_pre_attention = rnn.get_stacked_rnn(self.rnn_config, self.prefix, parallel_inputs=False,
+                                                         layers=[0])
+            self.rnn_post_attention = rnn.get_stacked_rnn(self.rnn_config, self.prefix, parallel_inputs=True,
+                                                          layers=range(1, self.rnn_config.num_layers))
+        self.rnn_pre_attention_n_states = len(self.rnn_pre_attention.state_shape)
 
 
         if not self.config.zero_state_init:
@@ -511,9 +512,9 @@ class RecurrentDecoder(Decoder):
         Creates parameters for encoder last state transformation into decoder layer initial states.
         """
         self.init_ws, self.init_bs, self.init_norms = [], [], []
-        state_shapes = self.attention_rnn.state_shape
-        if self.upper_layers_rnn:
-            state_shapes += self.upper_layers_rnn.state_shape
+        state_shapes = self.rnn_pre_attention.state_shape
+        if self.rnn_post_attention:
+            state_shapes += self.rnn_post_attention.state_shape
         for state_idx, (_, init_num_hidden) in enumerate(state_shapes):
             self.init_ws.append(mx.sym.Variable("%senc2decinit_%d_weight" % (self.prefix, state_idx)))
             self.init_bs.append(mx.sym.Variable("%senc2decinit_%d_bias" % (self.prefix, state_idx)))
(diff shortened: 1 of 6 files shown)


sockeye/rnn.py
```diff
@@ -48,20 +48,54 @@ def __init__(self,
...
+        # Adapted copy of mx.rnn.SequentialRNNCell.__call__()
+        self._counter += 1
+        next_states = []
+        pos = 0
+        for cell in self._cells:
+            assert not isinstance(cell, mx.rnn.BidirectionalCell)
+            length = len(cell.state_info)
+            state = states[pos:pos+length]
+            pos += length
+            concat_inputs = mx.sym.concat(inputs, parallel_inputs)
```
fhieber: on what axis does this concatenate? What are the shapes of inputs / parallel_inputs?

sockeye/rnn.py
```diff
@@ -48,20 +48,54 @@ def __init__(self,
...
+            inputs, state = cell(concat_inputs, state)
+            next_states.append(state)
+        return inputs, sum(next_states, [])
+
+    def add(self, cell):
+        super().add(cell)
+        print(self._cells)
```
fhieber: you probably can remove this

sockeye/rnn.py
```diff
@@ -48,20 +48,54 @@ def __init__(self,
...
-    rnn = mx.rnn.SequentialRNNCell()
-    for layer in range(config.num_layers):
+    rnn = mx.rnn.SequentialRNNCell() if not parallel_inputs else SequentialRNNCellParallelInput()
+    if not layers:
+        layers = range(config.num_layers)
+    for layer in layers:
         # fhieber: the 'l' in the prefix does NOT stand for 'layer' but for the direction 'l' as in mx.rnn.rnn_cell::517
         # this ensures parameter name compatibility of training w/ FusedRNN and decoding with 'unfused' RNN.
         cell_prefix = "%sl%d_" % (prefix, layer)
+        print(">>>>> cell_prefix:", cell_prefix)
```
fhieber: print

sockeye/decoder.py
```diff
@@ -375,10 +425,14 @@ def decode(self,
 
         lexical_biases = []
 
-        self.rnn.reset()
+        self.attention_rnn.reset()
         # TODO remove this once mxnet.rnn.SequentialRNNCell.reset() invokes recursive calls on layer cells
-        for cell in self.rnn._cells:
+        for cell in self.attention_rnn._cells:
```
fhieber: can we outsource the whole reset business in a separate private function to make it more concise here?

sockeye/decoder.py
```diff
@@ -158,7 +162,15 @@ def __init__(self,
             self.mapped_context_b = mx.sym.Variable("%smapped_context_bias" % prefix)
 
         # Stacked RNN
-        self.rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix)
+        if self.rnn_config.num_layers == 1 or not self.attention_in_upper_layers:
+            self.attention_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, False)
```
fhieber: Can we rename this a little bit? Its true that this is the rnn layer to compute the attention, but it might be easier to understand if we name it "rnn" and the upper layers just "upper_rnn".

sockeye/decoder.py
```diff
@@ -274,48 +299,30 @@ def _step(self,
...
                                   name="%sconcat_target_context_t%d" % (self.prefix, seq_idx))
         # rnn_output: (batch_size, rnn_num_hidden)
         # next_layer_states: num_layers * [batch_size, rnn_num_hidden]
-        rnn_output, layer_states = self.rnn(rnn_input, state.layer_states)
+        attention_rnn_output, attention_rnn_layer_states = \
+            self.attention_rnn(rnn_input, state.layer_states[:self.attention_rnn_n_states])
 
         # (2) Attention step
-        attention_input = self.attention.make_input(seq_idx, word_vec_prev, rnn_output)
+        attention_input = self.attention.make_input(seq_idx, word_vec_prev, attention_rnn_output)
```
fhieber: same comment here: its a bit confusing to read attention_input = .... attention_rnn_output
The lowest layer rnn isn't necessarily related to attention, its just the attention that uses it.

sockeye/decoder.py
```diff
@@ -274,48 +299,30 @@ def _step(self,
...
+        # (3) Upper layers if present or combine with attention
+        if self.upper_layers_rnn:
+            # TODO: context gating for multiple layers (with attention)
+            upper_rnn_output, upper_rnn_layer_states = \
+                self.upper_layers_rnn(attention_rnn_output, attention_state.context,
+                                      state.layer_states[self.attention_rnn_n_states:])
+            # TODO: if we do not include a last layer here, inference complains
+            #       about duplicate variable output names (??). Have to investigate
+            #       more.
+            # Do we want to include attention another time here?
```
fhieber: I think this MLP is fairly important, so I see no harm on including context here again.

sockeye/decoder.py
fhieber: it would make sense to explicitly name the arguments you are passing here, e.g.:
rnn.get_stacked_rnn(self.rnn_config, prefix=self.prefix, att_in_upper_layers=False)

sockeye/decoder.py
```diff
@@ -158,7 +162,15 @@ def __init__(self,
             self.mapped_context_b = mx.sym.Variable("%smapped_context_bias" % prefix)
 
         # Stacked RNN
-        self.rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix)
+        if self.rnn_config.num_layers == 1 or not self.attention_in_upper_layers:
+            self.attention_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, False)
+            self.upper_layers_rnn = None
+        else:
+            self.attention_rnn = rnn.get_stacked_rnn(self.rnn_config, self.prefix, False, layers=[0])
```
fhieber: same here

davvil:
For the global comments:
* Yes, this still needs some testing, so I will do the rebase.
* Inference does work.
* It has to do with attention, but to how the decoder uses it. The attention code is not changed at all for this modification. So I would keep it as a decoder option.

sockeye/decoder.py
davvil: I was considering "first_layer_rnn" or something similar. In the end I found clearer as attention_rnn, but I don't have a strong argument for it.

davvil: Will do.

sockeye/decoder.py
davvil: This part is the one where I still have to tweak a little bit. If we leave attention in, this is effectively the same as below if no context gating is used, so it can be simplified.

fhieber:
You are right, makes sense to leave flag as RNNDecoder parameter.

sockeye/decoder.py
davvil: It would be clearer, yes.

sockeye/rnn.py
davvil: On axis 1 (default for mx.sym.concat). In our use case inputs is (num_hidden, batch_size) and parallel_inputs is (attention_num_hidden, batch_size), resulting in concat_inputs (num_hidden + attention_num_hidden, batch_size).

sockeye/rnn.py
davvil: Ooops!

sockeye/rnn.py
davvil: Ooops! Ooops!

sockeye/decoder.py
fhieber: maybe its easiest to read if "rnn" is the prefix of these vairables. Something like rnn_layer_1 ?
rnn_attention vs. rnn_upper?

davvil: How about rnn_pre_attention and rnn_post_attention?

sockeye/rnn.py
```diff
@@ -88,8 +137,13 @@ def get_stacked_rnn(config: RNNConfig, prefix: str) -> mx.rnn.SequentialRNNCell:
                                           dropout_states=config.dropout)
 
         # layer_idx is 0 based, whereas first_residual_layer is 1-based
+        assert not (parallel_inputs and config.residual and config.first_residual_layer > 2), \
```
davvil: Still have to fix this.

sockeye/rnn.py
```diff
@@ -50,19 +51,67 @@ def __init__(self,
...
+    """
+    def __call__(self, inputs, parallel_inputs, states):
+        concat_inputs = mx.sym.concat(inputs, parallel_inputs)
+        output, states = self.base_cell(concat_inputs, states)
+        output = mx.symbol.elemwise_add(output, inputs, name="%s_plus_residual" % output.name)
+        return output, states
+
+
+def get_stacked_rnn(config: RNNConfig, prefix: str,
+                    parallel_inputs: bool = False, layers: Optional[Iterable[int]] = None) -> mx.rnn.SequentialRNNCell:
```
fhieber: missing docstring for the layers argument. Also, I am not entirely sure what it means :)

sockeye/rnn.py
fhieber: change to check_condition

sockeye/rnn.py
```diff
@@ -88,8 +137,13 @@ def get_stacked_rnn(config: RNNConfig, prefix: str) -> mx.rnn.SequentialRNNCell:
                                           dropout_states=config.dropout)
 
         # layer_idx is 0 based, whereas first_residual_layer is 1-based
+        assert not (parallel_inputs and config.residual and config.first_residual_layer > 2), \
+            "Parallel inputs with first residual layer > 2 not implemented yet"
         if config.residual and layer_idx + 1 >= config.first_residual_layer:
-            cell = mx.rnn.ResidualCell(cell)
+            if not parallel_inputs:
+                cell = mx.rnn.ResidualCell(cell)
```
fhieber: maybe also one-line this if/else?

sockeye/rnn.py
```diff
@@ -50,19 +51,67 @@ def __init__(self,
...
+
+
+def get_stacked_rnn(config: RNNConfig, prefix: str,
+                    parallel_inputs: bool = False, layers: Optional[Iterable[int]] = None) -> mx.rnn.SequentialRNNCell:
     """
     Returns (stacked) RNN cell given parameters.
 
     :param config: rnn configuration.
     :param prefix: Symbol prefix for RNN.
+    :param parallel_inputs: Support parallel inputs for the stacked RNN cells
```
fhieber: nitpick: missing period

sockeye/rnn.py
```diff
@@ -50,19 +51,67 @@ def __init__(self,
...
+            if self.concat_inputs:
+                concat_inputs = mx.sym.concat(inputs, parallel_inputs)
+                inputs, state = cell(concat_inputs, state)
+            else:
+                inputs, state = cell(inputs, parallel_inputs, state)
+            next_states.append(state)
+        return inputs, sum(next_states, [])
+
+
+class ResidualCellParallelInput(mx.rnn.ResidualCell):
```
fhieber: it would be good to have a test for this

sockeye/rnn.py
```diff
@@ -50,19 +51,67 @@ def __init__(self,
         self.residual = residual
         self.first_residual_layer = first_residual_layer
         self.forget_bias = forget_bias
+        self.attention_in_upper_layers = attention_in_upper_layers
 
 
-def get_stacked_rnn(config: RNNConfig, prefix: str) -> mx.rnn.SequentialRNNCell:
+class SequentialRNNCellParallelInput(mx.rnn.SequentialRNNCell):
```
fhieber: would be good to have a test

sockeye/decoder.py
```diff
@@ -740,7 +765,8 @@ def get_initial_state(self,
 
         # initial states for each layer
         layer_states = []
-        for state_idx, (_, init_num_hidden) in enumerate(self.rnn.state_shape):
+        for state_idx, (_, init_num_hidden) in enumerate(itertools.chain(
```
fhieber: mxnet does a similar list flattening using sum() in mx.rnn

sockeye/rnn.py
```diff
@@ -12,7 +12,7 @@
 # permissions and limitations under the License.
 
 # List is needed for mypy, but not used in the code, only in special comments
-from typing import Optional, List
+from typing import Optional, List, Iterable # NOQA pylint: disable=unused-import
```
fhieber: why this?

davvil: Because List is only used in a comment (intended for mypy, if I'm not mistaken), and thus pylint and flake8 (->NOQA) complain about an unused import.

sockeye/rnn.py
davvil: Right, I forgot. The basic idea is to specify which layers to create. It cannot be just the number of layers as we create identifiers with the layer index (it could be a number of layers and a first layer index, though).

sockeye/rnn.py
davvil: I missed these comments ;-)

sockeye/rnn.py
davvil: I put is as an assert because it is now a problem with the code. I hope to fix it in the next update.

sockeye/rnn.py
fhieber: ok makes sense

sockeye/rnn.py
davvil: Done

sockeye/rnn.py
davvil: Done

sockeye/rnn.py
davvil: Done, although not detected by github.

test/unit/test_rnn.py
```diff
@@ -77,3 +77,96 @@ def test_get_stacked_rnn(config, expected_cell):
...
+def test_residual_cell_parallel_input():
+    num_hidden = 128
+    batch_size = 256
+    parallel_size = 64
+
+    input_shape = (batch_size, num_hidden)
+    states_shape = (batch_size, num_hidden)
+    parallel_shape = (batch_size, parallel_size)
+
+    input = mx.sym.Variable("input")
```
fhieber: fine with me if the code checkers don't complain shadowing a reserved python word. I usually go for 'inp' :D

test/unit/test_rnn.py
```diff
@@ -77,3 +77,96 @@ def test_get_stacked_rnn(config, expected_cell):
         cell = cell.base_cell
     assert isinstance(cell, expected_cell)
     assert cell._num_hidden, config.num_hidden
+
+def test_residual_cell_parallel_input():
```
fhieber: nice test, thanks!

PR #107, opened 2017-08-15
Recombined arguments for source and target, e.g. support for things l…

fhieber:
…ike --num-words <src>:<trg>

This is a proposal how to reduce number of arguments a little bit as this is getting quickly out of hand :)
We can also think about more compact model definitions, such as
--encoder <type>:<num_hidden>:<layers> but this might be not common enough to all architecture types.


sockeye/arguments.py
```diff
@@ -52,6 +52,27 @@ def parse(schedule_str):
...
             raise argparse.ArgumentTypeError("Learning rate schedule string should have form rate1:num_updates1[,rate2:num_updates2,...]")
         return schedule
+
+    return parse
+
+
+def two_ints(greater_or_equal: Optional[int] = None) -> Callable:
+
+    def parse(value_to_check):
+        try:
```
tdomhan: why not just do:
```
if ':' in value_to_check:
  value1, value2 = value_to_check.split(C.ARG_SEPARATOR, 1)
else:
  value1 = value2 = int(value)
```
?

sockeye/arguments.py
```diff
@@ -173,20 +194,13 @@ def add_model_parameters(params):
...
-    model_params.add_argument('--num-words-source',
-                              type=int_greater_or_equal(0),
-                              default=None,
-                              help='Maximum source vocabulary size. Overrides --num-words. Default: %(default)s')
-    model_params.add_argument('--num-words-target',
-                              type=int_greater_or_equal(0),
-                              default=None,
-                              help='Maximum target vocabulary size. Overrides --num-words. Default: %(default)s')
+                              type=two_ints(greater_or_equal=0),
+                              default="50000:50000",
```
tdomhan: this comment applies to all the defaults: why not leave the default values at containing a single value (in order to use the same for source and target)?

sockeye/arguments.py
fhieber: fixed

sockeye/arguments.py
```diff
@@ -52,6 +52,22 @@ def parse(schedule_str):
         except ValueError:
             raise argparse.ArgumentTypeError("Learning rate schedule string should have form rate1:num_updates1[,rate2:num_updates2,...]")
         return schedule
+
+    return parse
+
+
+def two_ints(greater_or_equal: Optional[int] = None) -> Callable:
```
mjdenkowski: If we made this more general (int_list or similar), we could fold in things like number of filters for the convolutional embedding encoder:

    --conv-embed-num-filters 200:200:250:250:300:300:300:300

fhieber: makes sense, I generalized the function to make this possible.

sockeye/arguments.py
```diff
@@ -52,6 +52,33 @@ def parse(schedule_str):
...
+
+    return parse
+
+
+def multiple_values(num_values: Optional[int] = 0,
+                    greater_or_equal: Optional[float] = None,
+                    data_type: Callable = int) -> Callable:
+    """
+    Returns a method to be used in argument parsing to parse a string of the form "<val>:<val>[:<val>...]" into
+    a tuple of ints.
```
tdomhan: technically it's parsed into a tuple of `data_type`

sockeye/arguments.py
```diff
@@ -52,6 +52,33 @@ def parse(schedule_str):
...
+
+    :param num_values: Optional number of ints required.
+    :param greater_or_equal: Optional constraint that all values should be greater or equal to this value.
+    :param data_type: Type of values. Default: int.
+    :return: Method for parsing.
+    """
+
+    def parse(value_to_check):
+        if ':' in value_to_check:
+            values = tuple(map(data_type, value_to_check.split(C.ARG_SEPARATOR, num_values - 1)))
```
tdomhan: I'd add a check with a nice error message here. Something like:
```
expected_num_seperators = num_values - 1
check_condition(value_to_check.count(':') == expected_num_seperators, "Expected either a single value or %d values separated by %s" % (num_values, C.ARG_SEPARATOR))
```
(disclaimer: untested ;) )

PR #109, opened 2017-08-17
Benchmark translation output handler

mjdenkowski:
This adds an output handler for detailed benchmarking.  It records inputs/outputs/wall-clock time for each sentence using the standard tab-delimited format.

    [INFO:__main__] Translating...
    1 2 3 4 5
    input=1 2 3 4 5	output=1 2 3 4 5	input_tokens=5	output_tokens=6	translation_time=0.0856

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -587,7 +587,11 @@ def add_inference_args(params):
 
     decode_params.add_argument('--output-type',
                                default='translation',
-                               choices=["translation", "translation_with_alignments", "align_plot", "align_text"],
+                               choices=["translation",
+                                        "translation_with_alignments",
+                                        "benchmark",
+                                        "align_plot",
+                                        "align_text"],
                                help='Output type. Choices: [translation, translation_with_alignments, '
                                     'align_plot, align_text]. Default: %(default)s.')
     decode_params.add_argument('--sure-align-threshold',
--- a/sockeye/output_handler.py
+++ b/sockeye/output_handler.py
@@ -35,6 +35,8 @@ def get_output_handler(output_type: str,
         return StringOutputHandler(output_stream)
     elif output_type == "translation_with_alignments":
         return StringWithAlignmentsOutputHandler(output_stream, sure_align_threshold)
+    elif output_type == "benchmark":
+        return BenchmarkOutputHandler(output_stream)
     elif output_type == "align_plot":
         return AlignPlotHandler(plot_prefix="align" if output_fname is None else output_fname)
     elif output_type == "align_text":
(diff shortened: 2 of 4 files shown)


sockeye/arguments.py
```diff
@@ -587,7 +587,11 @@ def add_inference_args(params):
 
     decode_params.add_argument('--output-type',
                                default='translation',
-                               choices=["translation", "translation_with_alignments", "align_plot", "align_text"],
+                               choices=["translation",
```
fhieber: with the growing number of options here it looks like its time to adopt the pattern of defining a C.OUTPUT_HANDLER_CHOICES defined in constants.py as well as setting default to a defined constant.

sockeye/output_handler.py
```diff
@@ -35,6 +35,8 @@ def get_output_handler(output_type: str,
         return StringOutputHandler(output_stream)
     elif output_type == "translation_with_alignments":
         return StringWithAlignmentsOutputHandler(output_stream, sure_align_threshold)
+    elif output_type == "benchmark":
```
fhieber: these constants should be reused here then.

sockeye/output_handler.py
```diff
@@ -48,11 +50,15 @@ class OutputHandler:
     Abstract output handler interface
```
fhieber: lets slowly adopt the pattern of decoder.py of defining abstract classes.

sockeye/output_handler.py
```diff
@@ -48,11 +50,15 @@ class OutputHandler:
...
     """
 
-    def handle(self, t_input: sockeye.inference.TranslatorInput, t_output: sockeye.inference.TranslatorOutput):
+    def handle(self,
```
fhieber: in order to keep this interface simple we added the struct-like objects TranslatorInput & TranslatorOutput. Would it make sense to add a walltime to TranslatourOutput to keep the outputhandler's method signatures clean?

mjdenkowski: The idea is that we want to record as much of the total translation time as possible, so time() is the last thing we call before translate() and the first thing we call after.  To move timing inside TranslatorOutput, we could add time() calls a few methods down in the inference code (but not capture absolutely everything) or split timing across methods by tracking start and end times.  I think under the circumstances keeping timing outside is the simplest and most accurate.

fhieber: why not add a member to translation output and set it in the translate CLI itself:
t_output.walltime = tic - toc

mjdenkowski: Right now TranslationOutput is a NamedTuple and I like the idea of translator input/output objects being atomic/read-only.  The subtle issue is that creating the TranslationOutput object is part of translation and we want to record the total time of all method calls for translation.  In fact, looking at the code again I would consider moving the "tic" call to before creating TranslatorInput as well so we're measuring the total time from input string to output string.  Thoughts?

fhieber: Makes sense. Its better to have these translator objects as immutable data objects. 
Yes tic should also include generation of the input object.

PR #111, opened 2017-08-17
Disable capturing in pytest for easier debugging

abdelrahmanbadawy:


tdomhan:
why was the pull request closed?

abdelrahmanbadawy:
It failed the travis check for Python36. I guess the adopt should be --capture=no instead.

PR #112, opened 2017-08-17
Reworked dropout on RNN models

fhieber:
This PR adds more control over dropout at various places in the RNN architectures. Not backwards compatible.
Specifically it adds 
--embed dropout source:target
--rnn-dropout encoder:decoder [NOTE: this is now variational dropout]
--rnn-decoder-hidden-dropout
--conv-embed-dropout Dropout probability for ConvolutionalEmbeddingEncoder


sockeye/train.py
```diff
@@ -228,14 +228,23 @@ def main():
         num_embed_source, num_embed_target = args.num_embed
         encoder_num_layers, decoder_num_layers = args.num_layers
 
+        encoder_embed_dropout, decoder_embed_dropout = args.embed_dropout
+        encoder_rnn_dropout, decoder_rnn_dropout = args.rnn_dropout
+        if encoder_embed_dropout > 0 and encoder_rnn_dropout > 0:
+            logger.warning("Strange encoder dropout configuration: "
```
tdomhan: thanks for adding the error message. I'm not sure it would be clear to the user what's going on though. Maybe: "WARNING: Turning both the embedding dropout and the RNN input dropout on will lead to two dropout layers on top of each other. Are you sure you want this?" 
Note sure it's clearer :P

sockeye/arguments.py
```diff
@@ -430,11 +430,21 @@ def add_training_args(params):
...
+                              help='RNN variational dropout probability for encoder & decoder RNNs.'
+                                   'Use <val>:<val> to specify separate values. Default: %(default)s.')
+
+    train_params.add_argument('--rnn-decoder-hidden-dropout',
                               type=float,
-                              default=0.,
-                              help='Dropout probability for source embedding and source and target RNNs. '
-                                   'Default: %(default)s.')
+                              default=.0,
+                              help='Dropout probability for hidden state in RNN decoder. Default: %(default)s.')
```
tdomhan: 'hidden state' + the "--rnn" prefix make it sound like it's related to the decoder RNN, how could we make it clear that this is for the hidden state that combines context with the upper hidden state?
For the description one could write "Dropout probability for hidden state that combines the context with the RNN hidden state in the decoder".

sockeye/decoder.py
```diff
@@ -745,44 +769,65 @@ def _step(self,
...
+                    seq_idx: int) -> mx.sym.Symbol:
+        # hidden: (batch_size, rnn_num_hidden)
+        hidden = mx.sym.FullyConnected(data=mx.sym.concat(rnn_output, attention_state.context, dim=1),
+                                       # use same number of hidden states as RNN
+                                       num_hidden=self.num_hidden,
+                                       weight=self.hidden_w,
+                                       bias=self.hidden_b)
+        if self.config.layer_normalization:
+            hidden = self.hidden_norm.normalize(hidden)
+        if self.config.hidden_dropout > 0:
```
tdomhan: wouldn't we want dropout on the input here? Namely on `concat(rnn_output, attention_state..)`

sockeye/decoder.py
```diff
@@ -745,44 +769,65 @@ def _step(self,
...
+        # mapped_context: (batch_size, rnn_num_hidden)
+        mapped_context = mx.sym.FullyConnected(data=attention_state.context,
+                                               num_hidden=self.num_hidden,
+                                               weight=self.mapped_context_w,
+                                               bias=self.mapped_context_b,
+                                               name="%smapped_context_fc_t%d" % (self.prefix, seq_idx))
+
+        hidden = gate * mapped_rnn_output + (1 - gate) * mapped_context
+
+        if self.config.hidden_dropout > 0:
```
tdomhan: also here, should the dropout not be applied to the input rather than applying it before the activation?

sockeye/rnn.py
```diff
@@ -362,3 +365,42 @@ def __call__(self, inputs, states):
                                         name='%sout' % name)
 
         return next_h, [next_h]
+
+
+class VariationalDropoutCell(mx.rnn.ModifierCell):
+    """
+    Apply Bayesian Dropout on input and states separately. The dropout mask does not change when applied sequentially.
+
+    :param dropout_inputs: Dropout probability for inputs.
```
tdomhan: missing the base cell ;)

sockeye/train.py
fhieber: Right :) will change

sockeye/arguments.py
fhieber: I will change to the updated description.
I don't have a better idea for the argument name but think its good enough. 'rnn' stands for the attentional rnn model anyway now in contrast to 'transformer' so that should be fine.

sockeye/model.py
```diff
@@ -56,8 +56,8 @@ def __init__(self,
                  max_seq_len_target: int,
                  vocab_source_size: int,
                  vocab_target_size: int,
-                 config_encoder: encoder.EncoderConfig,
-                 config_decoder: decoder.RecurrentDecoderConfig,
+                 config_encoder: Config,
```
tdomhan: why did this change?

fhieber: EncoderConfig was too narrowly typed so I removed it.
RecurrentDecoderConfig was just wrong as it might also be a TransformerConfig (which is the same for both encoder and decoder).
So the next common type is Config which is fine I think.

PR #113, opened 2017-08-21
Add args for zero initialization and input sequence reversal

fhieber:
Also fixes a bug with multi-rnn layers (thanks @tdomhan) and does some mypy-related improvements.


sockeye/decoder.py
```diff
@@ -489,8 +491,9 @@ def _create_layer_parameters(self):
         """
         Creates parameters for encoder last state transformation into decoder layer initial states.
         """
-        self.init_ws, self.init_bs = [], []
-        self.init_norms = []
+        if self.config.zero_state_init:
+            return
```
fhieber: makes sense, changed

sockeye/decoder.py
```diff
@@ -712,29 +715,42 @@ def get_initial_state(self,
...
         :return: Decoder state.
         """
-        # initial decoder hidden state
-        hidden = mx.sym.tile(data=mx.sym.expand_dims(data=source_encoded_length * 0, axis=1), reps=(1, self.num_hidden))
+        # we derive the shape of hidden and layer_states from some input to enable
+        # shape inference for the batch dimension during inference.
+        # (batch_size, 1)
+        zeros = mx.sym.expand_dims(mx.sym.zeros_like(source_encoded_length), axis=1)
+        # last encoder state
+        source_encoded_last = mx.sym.SequenceLast(data=source_encoded,
```
davvil: Agreed, but it makes the code easier to understand.

sockeye/encoder.py
```diff
@@ -181,23 +191,23 @@ def encode(self,
...
-        Returns a list of RNNCells used by this encoder.
-        """
-        return []
+class ReverseSequence(Encoder):
+    """
+    Reverses the input sequence.
+    """
+
+    def encode(self,
+               data: mx.sym.Symbol,
```
tdomhan: could you add a comment on the fact that we expect batch major data?

sockeye/encoder.py
```diff
@@ -74,20 +79,24 @@ def get_recurrent_encoder(config: RecurrentEncoderConfig, fused: bool,
...
-    encoders = list()
+    encoders = list()  # type: List[Encoder]
 
     encoders.append(Embedding(num_embed=config.num_embed,
                               vocab_size=config.vocab_size,
                               prefix=C.SOURCE_EMBEDDING_PREFIX,
                               dropout=config.embed_dropout,
                               embed_weight=embed_weight))
+
+    if config.reverse_input:
```
tdomhan: would it make sense to run this after BatchMajor2TimeMajor in case the convolutional encoder is not present in order to save the additional swapaxes calls?

fhieber: good point. changed

PR #115, opened 2017-08-21
Parametrized length penalty for beam search.

tdomhan:


sockeye/inference.py
```diff
@@ -54,6 +54,10 @@ def __init__(self,
                  beam_size: int,
                  checkpoint: Optional[int] = None,
                  softmax_temperature: Optional[float] = None):
+        self.model_version = utils.load_version(os.path.join(model_folder, C.VERSION_NAME))
```
fhieber: specific reason why you moved this up?
Totally fine with me, but just curious.

sockeye/inference.py
```diff
@@ -315,6 +315,39 @@ def sort_state(self, best_hyp_indices: mx.nd.NDArray, best_word_indices: mx.nd.N
...
+    Calculates the length penalty as:
+    (beta + len(Y))**alpha / (beta + 1)**alpha
+
+    See Wu et al. 2016.
+
+    :param alpha: The alpha factor for the length penalty (see above).
+    :param beta: The beta factor for the length penalty (see above).
+    """
+
+    def __init__(self, alpha: float=1.0, beta: float=0.0):
```
fhieber: could you add "-> None" to make code analyzers happy?

sockeye/inference.py
```diff
@@ -315,6 +315,39 @@ def sort_state(self, best_hyp_indices: mx.nd.NDArray, best_word_indices: mx.nd.N
...
+        self.alpha = alpha
+        self.beta = beta
+        self.denominator = (self.beta + 1)**self.alpha
+
+    def __call__(self, lengths: mx.nd.NDArray):
+        """
+        Calculate the length penalty for the given vector of lengths:.
```
fhieber: ":." -> "."

sockeye/inference.py
tdomhan: I had an old model that failed because it didn't have this max_len attribute, rather than failing because of the version check. So I guess it makes sense to do the version check as the first check, before accessing any config attributes.

sockeye/inference.py
tdomhan: sure

sockeye/inference.py
tdomhan: changed

fhieber:
will merge after error with mypy is fixed

tdomhan:
fixed :)

PR #119, opened 2017-08-24
Made evaluate CLI a bit more forgiving if uneven number of hypotheses…

fhieber:
… and references. Also allow reading of hypothesis from stdin to enable piping from sockeye-translate.

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -665,3 +665,7 @@ def add_evaluate_args(params):
                              type=float,
                              default=0.01,
                              help="Numerical value of the offset of zero n-gram counts. Default: %(default)s.")
+    eval_params.add_argument('--not-strict', '-n',
+                             action="store_true",
+                             help="Do not fail if number of hypotheses does not match number of references. "
+                                  "Default: %(default)s.")
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -15,10 +15,12 @@
 Defines commandline arguments for the main CLIs with reasonable defaults.
 """
 import argparse
+import sys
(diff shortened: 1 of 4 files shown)


davvil:
Not sure how I feel about this. If you have a different number of hyps and refs there is something wrong with your files and I think it is ok to fail (and we even report the problem). Also, you are making the optimistic assumption that the missing lines are at the end of the file. If not, you are comparing unrelated sentences. If this happens in the middle or towards the beginning of the file, then probably you would notice because you have catastrophically low scores, but if it is towards the end of the file (but not **at the end**) you may get reasonable but pessimistic scores.

We could include this as an option, but I would be against making it the default behaviour.

fhieber:
For me its a common use case to quickly check the bleu score on a subset of the references, e.g.
`head inputs | translate | evaluate -r refs`. This does not work with the current version but with this pull request. Also, we are returning a warning so I think its clear enough. In any case users should now that translations and references should be parallel :)

tdomhan:
how about the suggestion to turn the strictness off with an argument, but leave it on by default? (`--not-strict`, `--allow-diff-len`, `--allow-different-lengths`). I also think that by default checking for some length is a good sanity check.

fhieber:
I still don't get the argument of failing here. Even with the version in this PR the CLI is warning the user that there is a non-equal amount of hyps & refs, but it still returns a score (which might be useful for advanced users).

davvil:
In my experience warnings are normally not enough, specially if they are part of a pipeline. Either they get lost in a big pile of log messages or they are even unseen as the result is handled by another tool. I would set the default behavior to fail.

PR #124, opened 2017-08-28
Added option for recurrent dropout without memory loss for LSTMs

fhieber:
Implements http://aclanthology.coli.uni-saarland.de/pdf/C/C16/C16-1165.pdf

backwards compatible.

### DIFF
--- a/sockeye/rnn.py
+++ b/sockeye/rnn.py
@@ -12,13 +12,14 @@
 # permissions and limitations under the License.
 
 # List is needed for mypy, but not used in the code, only in special comments
-from typing import Optional, List, Iterable # NOQA pylint: disable=unused-import
+from typing import Optional, List, Iterable  # NOQA pylint: disable=unused-import
 
 import mxnet as mx
 
 from sockeye.config import Config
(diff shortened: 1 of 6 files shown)


sockeye/rnn.py
```diff
@@ -295,6 +302,47 @@ def __call__(self, inputs, states):
...
+        gates = i2h + h2h
+        slice_gates = mx.sym.SliceChannel(gates, num_outputs=4,
+                                          name="%sslice" % name)
+        in_gate = mx.sym.Activation(slice_gates[0], act_type="sigmoid",
+                                    name='%si' % name)
+        forget_gate = mx.sym.Activation(slice_gates[1], act_type="sigmoid",
+                                        name='%sf' % name)
+        in_transform = mx.sym.Activation(slice_gates[2], act_type="tanh",
+                                         name='%sc' % name)
+        if self.dropout > 0.0:
```
davvil: As this is a cell explicitly including dropout (and the test for >0.0 is also done outside), it may make sense to make this a precondition in the constructor.

fhieber: good point, done

PR #128, opened 2017-08-31
Fixed typos and missing variables

lefterav:
General fixes of mistakes and typos
- Removed a leftover encoder declaration due to a bad merge
- Added missing imports
- Fixed typos in variables
- seq_length is in the end needed for the ConvolutionGluBlock, so I added it
- max_seq_length added as a config param and argument
- reordered new imports to be alphabetical


tdomhan:
sorry, saw this too late. This should also be fixed in the convseq2seq branch now.

PR #131, opened 2017-09-05
Added average encoder state RNN decoder initialization.

fhieber:
This should be backwards compatible which makes it a but ugly. 

Any opinions whether we should do it like this or create a new version?

### DIFF
--- a/test/unit/test_arguments.py
+++ b/test/unit/test_arguments.py
@@ -86,7 +86,7 @@ def test_device_args(test_params, expected_params):
               transformer_model_size=512,
               transformer_no_positional_encodings=False,
               attention_use_prev_word=False,
-              rnn_decoder_zero_init=False,
+              rnn_decoder_state_init="last",
               rnn_encoder_reverse_input=False,
               rnn_context_gating=False,
               rnn_cell_type=C.LSTM_TYPE,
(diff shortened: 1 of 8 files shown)


test/unit/test_arguments.py
```diff
@@ -87,6 +87,7 @@ def test_device_args(test_params, expected_params):
               transformer_no_positional_encodings=False,
               attention_use_prev_word=False,
               rnn_decoder_zero_init=False,
+              rnn_decoder_state_init="last",
```
davvil: C.RNN_DEC_INIT_LAST

davvil:
I wouldn't start deprecating arguments yet. I would go for a new version, even if the change is relatively small.

test/unit/test_arguments.py
fhieber: not using the defined constants is intentional here (also done for other args in the same test) so that we get a failure if we change the constant definition.

fhieber:
I think I agree. Since this is not an urgent change in any way, I will change it to bump the version but lets hold off with merging it and see if we have other non-backwards compatible changes coming in in the next couple of days.

test/unit/test_arguments.py
davvil: Right, we are in the test. Silly me :-)

PR #132, opened 2017-09-06
Added options to vary processing of data between transformer layers.

fhieber:
Also renamed prefixes a little bit to be more consistent.
Not backwards compatible.


sockeye/arguments.py
```diff
@@ -317,6 +317,20 @@ def add_model_parameters(params):
     model_params.add_argument('--transformer-no-positional-encodings',
                               action='store_true',
                               help='Do not use positional encodings.')
+    model_params.add_argument('--transformer-preprocess',
+                              type=multiple_values(num_values=2, greater_or_equal=None, data_type=str),
+                              default=('', ''),
+                              help='Transformer preprocess sequence for encoder and decoder.'
+                                   'For example ":drn" would refer to no preprocessing on the encoder and (d)ropout, '
```
davvil: The example is a bit confusing because of the empty part for the encoder. You could have an example like "d:drn" and some note like "leave a part empty for no preprocessing". Also, it would be good to list all the possible options. If it's only 'drn', then state this explicitely.

sockeye/transformer.py
```diff
@@ -139,57 +158,89 @@ def __call__(self,
...
-        if self.layer_norm is not None:
-            z = self._reshape_and_normalize(z, length)
-        return z
+        if not self.sequence:
+            return data
+
+        if prev is None:
+            assert 'r' not in self.sequence, "Residual connection not allowed if no previous value given."
+
+        for step in self.sequence:
```
davvil: Should we check for repetitions?

sockeye/transformer.py
```diff
@@ -139,57 +158,89 @@ def __call__(self,
...
+
+            if step == "r":
+                data = mx.sym._internal._plus(data, prev, name="%sresidual" % self.prefix)
```
davvil: Can't we use mxnet's residual implementation?

fhieber: What residual implementation are you referring to, thats not an RNN cell :)
Its just the '+' operator but I am using the internal plus to be able to name it.

sockeye/arguments.py
fhieber: This gets a bit long in my opinion if I state all combinations of three letters (also it could be longer than three). I will update to make it a bit clearer. But this is an expert setting. One has to go into the model architecture details to understand what you want to do :)

sockeye/transformer.py
davvil: Right, residual in mxnet is for RNNs.

sockeye/arguments.py
davvil: By options I meant 'processing' options. Something like "operations can be chosen from the set {d,r,n}".

fhieber:
I realized there was a bug in how layer inputs/outputs were combined:
Before:
y = layer(pre(x))
x = post(x, y)
Now (which is correct):
y = layer(pre(x))
x = post(y, x)

@tdomhan  please take another look if possible

PR #133, opened 2017-09-06
Fix: correctly passing the state dropout param.

tdomhan:


tdomhan:
totally missed this in the code review..

PR #134, opened 2017-09-07
Log exception and traceback for uncaught exceptions

fhieber:
This change ensures that uncaught exceptions are logged (to file or stdout).
See: https://stackoverflow.com/questions/1508467/log-exception-with-traceback

### DIFF
--- a/sockeye/log.py
+++ b/sockeye/log.py
@@ -13,8 +13,8 @@
 
 import logging
 import logging.config
+import sys
 from typing import Optional
-
 
 FORMATTERS = {
     'verbose': {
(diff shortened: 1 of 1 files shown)


davvil:
Really useful, thanks! One question, though: what happens once the exception has been logged? Does it still abort or does it try to continue because the exception has been caught (in stackoverflow they raise an exception again in the except block)?

fhieber:
From how I understand the documentation (https://docs.python.org/3/library/sys.html) there is no need to re-raise because you are not actually catching the exception. We are just overwriting the hook that would print the result to stderr and log it instead.

davvil:
Thanks, that wasn't clear to me.

PR #135, opened 2017-09-07
Alignment matrix output handler.

tdomhan:


tdomhan:
This is to support this alignment visualization tool:
https://github.com/M4t1ss/SoftAlignments 
which was presented at the MT marathon.

tdomhan:
@M4t1ss does look good?

tdomhan:
Also: I'm not super happy about the name of the output handler. Suggestions are welcome ;)

sockeye/output_handler.py
```diff
@@ -119,6 +121,54 @@ def handle(self,
         self.stream.flush()
 
 
+class NematusStringWithAlignmentsOutputHandler(StringOutputHandler):
+    """
+    Output handler to write translations and alignments to a stream. The format is compatible with the output
+    generated by Nematus. Note that unlike other output handlers each input sentence will result in an output
```
fhieber: Lets not use "Nematus" anywhere because its not really Nematus-related but rather compatible to another tool.
The docstring here could mention the link to the SoftAlignments project.
So maybe "SoftAlignmentsVisualizerOutputHandler"

sockeye/inference.py
```diff
@@ -551,9 +551,13 @@ def _make_result(self,
         :param attention_matrix: Attention matrix.
         :return: TranslatorOutput.
         """
+        # remove special sentence start symbol (<s>) from the output:
+        target_ids = target_ids[1:]
+        attention_matrix = attention_matrix[1:,:]
```
fhieber: why this? have you checked that this is compatible with the other output handlers?

sockeye/output_handler.py
```diff
@@ -119,6 +121,54 @@ def handle(self,
...
+    def handle(self,
+               t_input: sockeye.inference.TranslatorInput,
+               t_output: sockeye.inference.TranslatorOutput,
+               t_walltime: float = 0.):
+        """
+        :param t_input: Translator input.
+        :param t_output: Translator output.
+        :param t_walltime: Total wall-clock time for translation.
+        """
+        self.stream.write("%d ||| %s ||| %f ||| %s ||| %d ||| %d\n" % (t_input.id,
```
davvil: I would prefer here a .format() call with more descriptive names. My rule of thumb is "more than 3 substitutions, use format()". But I won't force you, of course :-)

davvil:
OutputFormatter?

sockeye/inference.py
tdomhan: no strong feelings, but I thought it was more consistent to have the target tokens match the target string. In the attention matrix we anyway don't have any entries for the beginning of sentence symbol so it doesn't add any value. I checked with the alignment plotting output handler. As the attention_matrix is consistent with the source and target tokens I don't expect this to break any of the handlers.

sockeye/output_handler.py
tdomhan: Yeah, it's just the format used by nematus, but I completely agree that it would make sense to leave "nematus" out as it is just a different output format that happens to be compatible with nematus. mhh on the other hand it's also not SoftAlignments specific. I was also considering "translation_with_details", but I'm also not super happy about that. "translation_with_alignment_matrix" and StringWithAlignmentMatrixOutputHandler, how's that?

sockeye/output_handler.py
tdomhan: sure, I can change this.

sockeye/inference.py
fhieber: makes sense.

sockeye/output_handler.py
fhieber: Sounds good!

fhieber:
OutputFormatter is too broad I would say. I like StringWithAlignmentMatrixOutputHandler

PR #136, opened 2017-09-09
Size batches by number of words

mjdenkowski:
This adds support for batching by number of words via the --batch-type option.  For each bucket, a custom batch size (number of sentences) is set such that the average number of non-padding words per batch is (1) as close as possible to the requested size, (2) subject to being evenly divisible by the number of devices.  While the actual number of words per batch will vary (impossible to avoid without truncating sentences), reasonable batch sizes (~4096) and bucket widths (~10) keep the numbers pretty close to the average.


sockeye/data_io.py
```diff
@@ -461,15 +497,33 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
...
-        for bkt, buck in zip(self.buckets, self.data_source):
-            logger.info("bucket of {0} : {1} samples".format(bkt, len(buck)))
+        for bkt, buck, (batch_size_seq, _), average_seq_len in zip(
+                self.buckets, self.data_source, self.bucket_batch_sizes, self.data_label_averge_len):
+            logger.info("Bucket of %s : %d samples in %d batches of %d, approx %0.1f words/batch",
+                        bkt,
+                        len(buck),
+                        math.ceil(len(buck) / batch_size_seq),
+                        batch_size_seq,
+                        batch_size_seq * average_seq_len)
```
mjdenkowski: We now report both sentences and average words per batch.  This really clarifies how unbalanced sentence-based batches can be.

sockeye/data_io.py
```diff
@@ -479,24 +533,63 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
...
+    def _populate_bucket_batch_sizes(self):
+        """
+        Number of sentences and words per batch for each bucket
+        """
+        # Pre-defined bucket batch sizes
+        if self.bucket_batch_sizes is not None:
+            return
+        # Otherwise compute here
+        self.bucket_batch_sizes = [[] for _ in self.buckets]
+        for buck_idx in range(len(self.buckets)):
```
fhieber: can we use enumerate here to get buck_idx and bucket_shape / padded_seq_len jointly?
Probably also average_seq_len

sockeye/data_io.py
```diff
@@ -479,24 +533,63 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
...
+            if self.batch_by_words:
+                # Multiple of number of devices (int) closest to target number of words, assuming each sentence is of
+                # average length
+                batch_size_seq = self.batch_num_devices * round((self.batch_size / average_seq_len)
+                                                                / self.batch_num_devices)
+                batch_size_word = batch_size_seq * average_seq_len
+            else:
+                batch_size_seq = self.batch_size
+                batch_size_word = batch_size_seq * average_seq_len
+            self.bucket_batch_sizes[buck_idx] = [batch_size_seq, batch_size_word]
```
fhieber: can we make this a tuple to make clear its non-modifiable?

sockeye/data_io.py
```diff
@@ -479,24 +533,63 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
         logger.info("fill up mode: %s", self.fill_up)
         logger.info("")
 
+    def _populate_bucket_batch_sizes(self):
+        """
+        Number of sentences and words per batch for each bucket
```
fhieber: It would be great to a little bit more documentation for this function, e.g. explicitly state what member it sets (on a related note: would it be possible to make this function return bucket_batch_sizes and set it explicitly where called?) and what it does for both batching cases: (1) Sentence-based batch size: all bucket batch sizes are the same, namely set to self.batch_size. (2) ?

sockeye/data_io.py
```diff
@@ -361,11 +372,14 @@ class ParallelBucketSentenceIter(mx.io.DataIter):
     :param buckets: List of buckets.
     :param batch_size: Batch_size of generated data batches.
            Incomplete batches are discarded if fill_up == None, or filled up according to the fill_up strategy.
+    :param batch_by_words: Size batches by words rather than sentences.
+    :param batch_num_devices: Number of devices batches will be parallelized across.
```
fhieber: I don't think I yet understand why we need this. 
Currently, the (sentence-based) batch size we get from the commandline is the 'total' batch size which then gets divided by n if we use n devices for training. Is this different for the word-based batch size? It'd be great to a have a minimum example to better understand this.

sockeye/data_io.py
```diff
@@ -401,29 +420,44 @@ def __init__(self,
...
+                           shape=(self.default_bucket_batch_size, self.default_bucket_key[0]),
+                           layout=C.BATCH_MAJOR),
+            mx.io.DataDesc(name=target_data_name,
+                           shape=(self.default_bucket_batch_size, self.default_bucket_key[1]),
+                           layout=C.BATCH_MAJOR)]
         self.provide_label = [
-            mx.io.DataDesc(name=label_name, shape=(self.batch_size, self.default_bucket_key[1]), layout=C.BATCH_MAJOR)]
+            mx.io.DataDesc(name=label_name,
+                           shape=(self.default_bucket_batch_size, self.default_bucket_key[1]),
+                           layout=C.BATCH_MAJOR)]
```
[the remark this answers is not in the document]
fhieber: Yes the metrics will report the same because we collapse batch and time dimension in the softmax output.
I wasn't aware of MXNet's behavior with different batch sizes for batches passed to a module. I think I need to better understand this a bit to be able to completely follow this PR :)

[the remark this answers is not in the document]
mjdenkowski: I couldn't find concrete documentation for this anywhere, but the working idea is that:
1. Each DataBatch includes a DataDesc with its own provide_data shape.
2. That means in principle MXNet should be able to handle batches of different sizes correctly.
3. We need a single provide_data shape for the DataIter so that memory can be reserved for staging batches.
4. If we make this area large enough to fit the largest batch, MXNet appears to do the right thing.

sockeye/data_io.py
mjdenkowski: We can add a comment about this to the code.

If we're only using whole sentences, batches will almost never have the right number of words.  If our batch size is 4096, we can add sentences until either we exceed 4096 or the next sentence would exceed 4096.  If all of our sentences are length 100, we could choose a batch size of 4000 or 4100.

When we have multiple devices, we also want the number of sentences to be evenly divisible so we won't waste computation (some devices processing N examples per batch, some N+1).

The two most direct solutions are:
1. Set number of sentences to the multiple of number of devices that gets us closest to the target batch size.  (Num devices can influence batch size)
2. Set number of sentences to be as close as possible to the target batch size and add padding sentences to make the batch size divisible by number of devices.  (Wasted computation for majority of batches)

This implementation goes with option 1.

sockeye/data_io.py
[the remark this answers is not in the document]
fhieber: Sounds good! This is in line with what we already to with bucketing:
dataIter.provide_data should use the default/largest bucket key so that parameter and memory sharing for different graphs is working properly. Every batch however, has its own bucket_key member that determines which graph the batch should be routed to.

In your case, I would assume that MXNet requires to know the maximum (batch) size once for each graph, allocates the respective memory, and then dynamically reshapes the graph based on the batch size of each batch, reusing already allocated memory.

Regarding bucketing, I think @tdomhan did an investigation once that if one doesn't use the right default bucket key when creating the module, MXNet does not complain but just silently does the wrong thing :)
We should make sure nothing like this can happen here, either by some empirical evidence that this is doing the right thing or by looking at the module code and try to understand the logic. On a related note: does this affect speed at all?

sockeye/data_io.py
fhieber: Thanks for the explanation! I agree that option 1 is the better choice.

sockeye/data_io.py
```diff
@@ -401,29 +420,44 @@ def __init__(self,
...
 
         # convert to single numpy array for each bucket
         self._convert_to_array()
 
+        # "Staging area" that needs to fit any size batch we're using by total number of elements
         self.provide_data = [
-            mx.io.DataDesc(name=source_data_name, shape=(batch_size, self.default_bucket_key[0]), layout=C.BATCH_MAJOR),
-            mx.io.DataDesc(name=target_data_name, shape=(batch_size, self.default_bucket_key[1]), layout=C.BATCH_MAJOR)]
+            mx.io.DataDesc(name=source_data_name,
+                           shape=(self.default_bucket_batch_size, self.default_bucket_key[0]),
```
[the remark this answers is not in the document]
mjdenkowski: Right.  It gets set in _populate_bucket_batch_sizes() which is called by _assign_to_buckets().  It's a little convoluted, but pulling it out and still only iterating over the data once would require substantial refactoring.  I'll add a comment about where it's set.

sockeye/data_io.py
mjdenkowski: The method is pretty heavily commented but I can add a few notes to the docstring as well.

sockeye/train.py
```diff
@@ -188,6 +188,7 @@ def main():
 
         # create data iterators
         max_seq_len_source, max_seq_len_target = args.max_seq_len
+        batch_num_devices = 1 if args.use_cpu else sum(-di if di < 0 else 1 for di in args.device_ids)
```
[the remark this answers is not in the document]
mjdenkowski: I looked there first and it appeared that the GPU acquisition logic is pretty tightly coupled to identifying and locking actual devices rather than just finding a number.

sockeye/data_io.py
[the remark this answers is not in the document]
tdomhan: memory sharing isn't well documented in mxnet. The link only is part of how memory sharing is implemented. Essentially this is for all ndarrays that are in python space. Additionally there is logic for sharing ndarrays allocated in C++ land between the C++ executor objects. The executor of the default bucket key is where ndarrays for all other executors come from. The logic is sort of like this pseudo code:
```
unused_ndarrays = default_executor.ndarrays
for ndarray_size in ndarrays_needed_for_executor:
  ndarray = unused_ndarrays.pop_smallest_that_fits(ndarray_size)
  if ndarray == None:
     ndarray = new NDarray()

```
This is happening somewhere here:
https://github.com/apache/incubator-mxnet/blob/master/src/executor/graph_executor.cc#L917 


Bottom line is that the ndarrays of the executor for the default bucket key need to be large enough to fit ndarrays of all other executors. If so no additional memory is required.

sockeye/train.py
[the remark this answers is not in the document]
tdomhan: it should be possible to use `1 if args.use_cpu else len(expand_requested_device_ids(args.device_ids))`

sockeye/data_io.py
```diff
@@ -361,11 +372,14 @@ class ParallelBucketSentenceIter(mx.io.DataIter):
...
     :param batch_size: Batch_size of generated data batches.
            Incomplete batches are discarded if fill_up == None, or filled up according to the fill_up strategy.
+    :param batch_by_words: Size batches by words rather than sentences.
+    :param batch_num_devices: Number of devices batches will be parallelized across.
     :param fill_up: If not None, fill up bucket data to a multiple of batch_size to avoid discarding incomplete batches.
            for each bucket. If set to 'replicate', sample examples from the bucket and use them to fill up.
     :param eos_id: Word id for end-of-sentence.
     :param pad_id: Word id for padding symbols.
     :param unk_id: Word id for unknown symbols.
+    :param bucket_batch_sizes: Pre-computed bucket batch sizes (used to keep iterators consistent for train/validation)
```
tdomhan: nit-picky: period missing. More importantly as it's a list of lists of ints could you briefly describe the contents of these.

sockeye/data_io.py
```diff
@@ -374,9 +388,12 @@ def __init__(self,
                  target_sentences: List[List[int]],
                  buckets: List[Tuple[int, int]],
                  batch_size: int,
+                 batch_by_words: bool,
+                 batch_num_devices: int,
                  eos_id: int,
                  pad_id: int,
                  unk_id: int,
+                 bucket_batch_sizes: Optional[List[List[int]]] = None,
```
tdomhan: these are always lists of tuples with 2 elements, right? You could either use: `Optional[List[Tuple[int, int]]]` or potentially even put `(batch_size_seq, batch_size_word)` into a namedtuple.

sockeye/data_io.py
```diff
@@ -479,24 +534,76 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
...
+            self.bucket_batch_sizes[buck_idx] = (batch_size_seq, batch_size_word)
+        # Default bucket batch size: batch size of largest bucket batch by total elements, scaled for default bucket key
+        # Effectively, set default values to reserve enough space to provide a batch from any bucket
+        largest_total_batch_size = 0
+        for buck_shape, (batch_size_seq, _) in zip(self.buckets, self.bucket_batch_sizes):
+            seq_len = max(*buck_shape)
+            buck_batch_size = seq_len * batch_size_seq
+            if buck_batch_size > largest_total_batch_size:
+                largest_total_batch_size = buck_batch_size
+                self.default_bucket_batch_size = math.ceil((batch_size_seq * seq_len) / max(*self.default_bucket_key))
```
tdomhan: you could reuse `buck_batch_size `
```
self.default_bucket_batch_size = math.ceil(buck_batch_size / max(*self.default_bucket_key))
```

sockeye/data_io.py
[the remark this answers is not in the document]
tdomhan: In any case this is something that has to be tried out. MXNet will only complain if the batch size of the default bucket is too small. However, if it can't reuse NDArrays internally in the graph executor there won't be any warnings, but rather it will just silently allocate new arrays. So in practice running sockeye with and without the `--no-bucketing` arguments is the easiest way of telling if the memory sharing is working correctly (memory consumption should not change much). A safe option might be to use the largest batch size of any of the buckets, but that might use too much memory.

sockeye/data_io.py
```diff
@@ -479,24 +534,76 @@ def _assign_to_buckets(self, source_sentences, target_sentences):
...
+                                                                / self.batch_num_devices)
+                batch_size_word = batch_size_seq * average_seq_len
+            else:
+                batch_size_seq = self.batch_size
+                batch_size_word = batch_size_seq * average_seq_len
+            self.bucket_batch_sizes[buck_idx] = (batch_size_seq, batch_size_word)
+        # Default bucket batch size: batch size of largest bucket batch by total elements, scaled for default bucket key
+        # Effectively, set default values to reserve enough space to provide a batch from any bucket
+        largest_total_batch_size = 0
+        for buck_shape, (batch_size_seq, _) in zip(self.buckets, self.bucket_batch_sizes):
```
tdomhan: for sentence based batch sizes this is essentially a noop, no? could you make this explicit? something like:
```
if self.batch_by_words:
  for ...
else:
  self.default_bucket_batch_size = self.batch_size
```

mjdenkowski:
As of the last commit, things should be in shape for a merge.  Anyone want to sign off on this?

fhieber:
I think this looks good, but it would be great to have one of the system checks use the new option and assert on bleu for seq_copy/seq_sort.

Did Tobis suggestion with `expand_requested_device_ids()` not work?

mjdenkowski:
`expand_requested_device_ids()` also writes log messages about acquiring GPUs, which we aren't actually doing here.  We could refactor it into two functions and re-use one, though I'm not sure that's worth it to replace one line of code.

fhieber:
yeah thats not worth it. I will approve & merge, but could you add `--batch-type word` in at least one of the system/integration tests so that we have the test suite run this code path at least once. Thanks!

mjdenkowski:
Sounds good.  I added a test and started everything running.  I'll check in the commit once everything passes.

PR #144, opened 2017-09-17
Eve Optimizer

mjdenkowski:
This is an implementation of the Eve optimizer (https://arxiv.org/abs/1611.01505).

Eve is an extension of the widely used Adam optimizer that incorporates feedback from the objective function. In addition to applying Adam's momentum, the optimizer divides the learning rate for each step by a running average of the relative change in the objective function. This is presented as a technique to better handle saddle points. From the paper: "When the average relative change is high, the learning rate is reduced. This can improve the progress if, for example, the algorithm is bouncing around the walls of the objective function. Conversely, when the relative change is low, the learning rate is increased. This can help the algorithm accelerate through flat areas in the loss surface."

This PR includes:

An implementation of the optimizer that can use either batch or checkpoint loss (or both).
Extensions to training to make the value of the objective function (aka loss/metrics) visible to the optimizer.
A unit test for optimizers, including MXNet Adam as a check for the test itself and Eve with two kinds of loss.
Default settings and behavior are unchanged.


sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+BatchState = namedtuple("BatchState", ["metric_val"])
+CheckpointState = namedtuple("CheckpointState", ["checkpoint", "metric_val"])
+
+
+class SockeyeOptimizer(Optimizer):
+    """
+    Optimizer that has access to additional information from the last batch and the last chekpoint
+    when updating weights.
+    """
+    def __init__(self, **kwargs):
```
fhieber: could you try adding optimizers.py to typechecked-files in the package root?
You will need to annotate the return type of __init__ methods with "-> None" to avoid an error.

sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+class Eve(SockeyeOptimizer):
+    """
+    The Eve optimizer is an extended version of Adam that incorporates feedback from the objective
+    function to further adapt the learning rate.
+        * "Improving Stochastic Gradient Descent with Feedback"
+          Jayanth Koushik; Hiroaki Hayashi (https://arxiv.org/abs/1611.01505)
+
+    This version allows using validation checkpoint loss in addition to training batch loss.
+
+    Eve does not currently support rescaling gradients, clipping gradients, or weight decay.
```
fhieber: lets log a warning about this in the Eve constructor, i.e. if rescale_gradient etc is present in kwargs.

Also, it seems like Eve is generally not supporting l1/l2 regularizations (Section 3.1), so using weight decay should probably throw an error to indicate this.

sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+        self.beta3 = beta3
+        self.beta4 = beta4
+        self.epsilon = epsilon
+        self.k_lo = k_lo
+        self.k_hi = k_hi
+        self.use_batch_objective = use_batch_objective
+        self.use_checkpoint_objective = use_checkpoint_objective
+
+    def create_state(self, index: int, weight: NDArray):
+        return (zeros(weight.shape, weight.context, dtype=weight.dtype),  # mean
```
fhieber: you can use nd.zeros_like(weight, ctx=weight.context) here.
(https://mxnet.incubator.apache.org/api/python/ndarray.html#mxnet.ndarray.zeros_like)

sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+                zeros(weight.shape, weight.context, dtype=weight.dtype),  # variance
+                [0, 1],     # batch previous objective "hat", previous d
```
fhieber: these lists seem to be of fixed size, would it make sense to define them as tuples?

sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+            else:
+                d_checkpoint = d_prev
+            d += d_checkpoint
+
+        # Batch and checkpoint contribute equally when both are used
+        if self.use_batch_objective and self.use_checkpoint_objective:
+            d /= 2.
+
+        # Final weight update rule (Adam rule with extra d term)
+        weight[:] = weight - lr * mean / (d * sqrt(var) + self.epsilon)
```
fhieber: MXNet's standard Adam implements the update rule in C++, afaik (https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer.py#L649). Could we reuse this? It seems that you are just scaling and shifting the variance term with d and eps here, so we could fold this into the var variable passed to adam_update and leave the update to the C++ backend.

sockeye/training.py
```diff
@@ -280,6 +281,7 @@ def _fit(self,
         :param mxmonitor: Optional MXNet monitor instance.
         """
         metric_train = self._create_eval_metric(metrics)
+        metric_batch = self._create_eval_metric(metrics)
```
fhieber: General question about what to pass to the extended optimizers: we mostly track perplexity on training and validation data sets but actually optimize cross-entropy. While strongly related, have you tried passing cross-entropy values to Eve instead? Also, since using SoftmaxOutput gradient already entails cross-entropy loss, we could directly use the CE metric here and would not need to create other metrics (e.g. Accuracy). This would also allow you to get rid of these repeated for loops over the metrics below, by specifically creating an 'eve metric'.

We should also make sure that this won't interact strangely with any sequence losses from the MRT branch, or maybe more importantly, how we normalize/scale the loss. This will affect the size of the gradients you get in Eve.update() but won't be reflected in the batch metric here.

sockeye/optimizers.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Now compute Eve's d term
+        d = 0.
+
+        # Eve rules from paper: compute f_hat and d based on last training batch objective
+        if self.use_batch_objective:
+            f_hat_prev, d_prev = prev_batch
+            f = self.batch_state.metric_val
+            if t > 1:
+                # The original paper has a typo in the algorithm here.  The following lines are re-
```
fhieber: computation and update of these intermediate values is duplicated for batch and checkpoint objective. Could you consider factoring this out into a common method?

sockeye/training.py
mjdenkowski: Right now Eve uses whatever we specify as `optimized_metric`.  If the optimizer is making updates based on a different loss function, we should be able to use that as well.  Do we know if/where batch loss is stored in a form we can access?

sockeye/optimizers.py
mjdenkowski: That would be ideal, though unfortunately it combines updating mean/var with the actual weight update.  We need to update var as it is, then scale by d for the weight update.

sockeye/optimizers.py
mjdenkowski: That's probably a better way to do things for now.  Some of the extensions we've been looking at can treat batch/checkpoint values differently so these may not always be identical.

sockeye/optimizers.py
mjdenkowski: Nice!  This code is based on MXNet's Adam implementation which may not be totally up to date with best practice.

sockeye/optimizers.py
mjdenkowski: I'm actually thinking of making an EveState class.  MXNet prefers a "keep it simple" approach but now we're tracking a bunch of labeled values.

sockeye/optimizers.py
mjdenkowski: Good idea to warn early about that.  I'll add some `check_condition` statements.

sockeye/training.py
fhieber: Just to make sure this is clear: optimized_metric does not define the actual loss function that is optimized by MXNet (this refers to the metric that is used to compute when to do early stopping w.r.t. validation scores). That said, we currently always minimize cross-entropy on the training set as given by the symbol created in loss.py. If you use smoothed-cross-entropy you should get a grouped symbol of logits and the loss, whereas for the standard cross-entropy loss we use the SoftmaxOutput symbol which returns logits in the forward, and backprobs cross-entropy gradients, so you don't have direct access to the loss function value for a batch.

So what I would suggest is to create a cross-entropy metric (``mx.metric.create('ce')``) to be used with the Eve optimizer so that it gets the correct loss function values for scaling.

mjdenkowski:
In the last commit, I added support for cross-entropy and smoothed-cross-entropy.  The first uses MXNet's CE metric while the second uses a helper metric to sum pre-computed SCE values.  We automatically select the appropriate loss and (for SCE) check to make sure we're only normalizing once.

Validation loss still uses `optimized_metric`.

mjdenkowski:
This last commit also adds a command line option for `--optimizer-params`, passed as a JSON dictionary.  This will let us experiment with the various hyper-parameters of our optimizers without endless specialized CLI args.

sockeye/train.py
```diff
@@ -399,6 +400,16 @@ def main():
...
             # Making MXNet module API's default scaling factor explicit
             optimizer_params["rescale_grad"] = 1.0 / args.batch_size
+        if optimizer == C.OPTIMIZER_EVE:
+            check_condition(args.normalize_loss, "Eve optimizer requires loss normalization (use --normalize-loss)")
+            check_condition(not clip_gradient,
+                            "Eve optimizer does not support gradient clipping (use --clip-gradient -1)")
+            optimizer_params["use_batch_objective"] = args.eve_loss in(C.EVE_LOSS_BATCH, C.EVE_LOSS_BOTH)
+            optimizer_params["use_checkpoint_objective"] = args.eve_loss in(C.EVE_LOSS_CHECKPOINT, C.EVE_LOSS_BOTH)
+        # Manually specified params
+        if args.optimizer_params:
```
fhieber: args.optimizer_params is just a dictionary, right?
You can to optimizer_params.update(args.optimizer_params) then

sockeye/training.py
```diff
@@ -281,6 +282,12 @@ def _fit(self,
         """
         metric_train = self._create_eval_metric(metrics)
         metric_val = self._create_eval_metric(metrics)
+        # Track loss for optimizer if needed
+        if isinstance(self.module._curr_module._optimizer, SockeyeOptimizer):
+            if self.config.config_loss.type == C.CROSS_ENTROPY:
+                metric_loss = mx.metric.create("ce")
```
fhieber: is the 'ce' metric normalized ?

sockeye/loss.py
```diff
@@ -168,3 +169,38 @@ def get_loss(self, logits: mx.sym.Symbol, labels: mx.sym.Symbol) -> List[mx.sym.
...
+    optimizers during training.
+
+    :param normalized: Whether loss is pre-normalized by the number of examples.
+    :param name: Name of this metric instance for display.
+    :param output_names: Name of predictions that should be used when updating with update_dict.
+    :param output_labels: Name of labels that should be used when updating with update_dict.
+    """
+    def __init__(self,
+                 normalized: bool = False,
+                 name: str = "smoothed-cross-entropy",
```
fhieber: could use the constant here to replace the string

sockeye/loss.py
```diff
@@ -168,3 +169,38 @@ def get_loss(self, logits: mx.sym.Symbol, labels: mx.sym.Symbol) -> List[mx.sym.
...
+    def update(self, labels, preds):
+        """Sum pre-computed smoothed cross entropy values"""
+        # Take cross entropy values only
+        sces = preds[::2]
+        for label, sce in zip(labels, sces):
+            label = label.asnumpy()
+            label = label.ravel()
+            assert label.shape[0] == sce.shape[0]
+            # SCE is pre-computed, so just sum
+            self.sum_metric += sce.asnumpy().sum()
```
fhieber: we could do the sum in mx.nd space and only copy to numpy afterwards. If the sces live on the gpu this is cheaper and faster.
Same above with the ravel. You can ask ndarrays for their shape as well and this is faster than copying to numpy before.

sockeye/training.py
mjdenkowski: For CE, the stats are not normalized, but the `get()` calls are.  The update rule is:
```
prob = pred[numpy.arange(label.shape[0]), numpy.int64(label)]
self.sum_metric += (-numpy.log(prob + self.eps)).sum()
self.num_inst += label.shape[0]
```
A call to `get_name_value()` returns `sum_metric / num_inst`.

For SCE with loss normalization, the values summed for `sum_metric` are already divided by batch instances as part of the loss calculation, so we just use `num_inst += 1`.

sockeye/loss.py
mjdenkowski: Another case where code borrowed from MXNet's implementation isn't as efficient as it could be.

mjdenkowski:
I moved array reshaping to ndarray operations.  Moving the sum produced subtly different values that added up after awhile, presumably due to conversion/rounding.  For now I would keep that in numpy to be consistent with what MXNet's CE implementation does.

On an unrelated note, mxnet==0.10.0 seems to have disappeared from pip for Mac, with only 0.9.5 or 0.11.0 available.  Upgrading to 0.11.0 generally works but some tests fail with errors about binding.  Other than that, it looks like a few things were removed such as the `*=` operator.  Now might be a good time to go through and update everything for 0.11.0.

Finally I had to take optimizers.py out of typechecked-files.  mypy doesn't like assigning typed objects to None and Python < 3.6 doesn't like inline type annotations on None, so there doesn't seem to be a combination that will pass on 3.4, 3.5, and 3.6.

sockeye/training.py
fhieber: I see, so normalization for 'ce' is with respect to the number of instances (the size of the batch dimension), never w.r.t to the number of valid tokens in the batch (i.e. the number of non-pad tokens in labels). This is consistent to what SoftmaxOutput backpropagates if args.normalize_loss is off (i.e. batch normalization of loss, using rescale_grad=1/batch_size).

If args.normalize_loss is on, SoftmaxOutput from the CrossEntropyLoss is using 'valid' normalization, effectively normalizing by the number of valid label tokens in the batch.

For these Eve metrics, we could use batch normalization as default ( `num_inst += label.shape[0]`) if args.normalize_loss is off, and `num_inst += mx.nd.sum(labels != C.PAD_ID)` if args.normalize_loss is on. 
However, doing that consistent for both CE and SCE metrics would also require overwriting the CE metric. Not sure if we want to do this... In any case, it might be hard to exactly get the same loss from the metric as is used to calculate gradients in the graph.

sockeye/optimizers.py
```diff
@@ -35,8 +35,8 @@ class SockeyeOptimizer(Optimizer):
     when updating weights.
     """
     def __init__(self, **kwargs) -> None:
-        self.batch_state: BatchState = None
-        self.checkpoint_state: CheckpointState = None
+        self.batch_state = None
```
fhieber: You can leave the type annotations but change them to Optional[BatchState] to allow None (then typechecking will also work :))

fhieber:
This disappearance of 0.10 is interesting... I would prefer having a separate mxnet update change though to be clear what needs to be modified in the codebase (i.e. removal of the symbolic inplace operations). But we also need good testing first w.r.t performance and speed before making the change.

sockeye/training.py
mjdenkowski: Good point about valid tokens.  CE does call `ravel()` so it normalizes by the total number of tokens in the batch (`shape[0]`), but it doesn't check for padding.  The logic is about 10 lines so it's not terrible to keep our own version.  In fact, it's probably a good idea to keep a Metric version for every Loss we have.

sockeye/optimizers.py
mjdenkowski: Good to know!

sockeye/training.py
fhieber: Agreed, this is however only necessary for losses that are not returned by the symbol themselve (as is the case for CrossEntropy). For all others, e.g. SmoothedCrossEntropy, which also returns a symbol of the loss value, you can use the `Loss` metric from MXNet: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/metric.py#L976

fhieber:
Nice change, looks good to me now. Thanks a lot for iterating.
Once you fix the remaining failing unit test, we can merge this I think.

szha:
Would you be interested in committing the Eve optimizer to MXNet?

sockeye/training.py
```diff
@@ -281,6 +282,10 @@ def _fit(self,
         """
         metric_train = self._create_eval_metric(metrics)
         metric_val = self._create_eval_metric(metrics)
+        # Track loss for optimizer if needed
+        if isinstance(self.module._curr_module._optimizer, SockeyeOptimizer):
+            # Create metric corresponding to training loss
+            metric_loss = mx.metric.create(self.config.config_loss.type, loss_config=self.config.config_loss)
```
tdomhan: would it make sense to create a constant (with better naming ;)) `OPTIMIZER_LOSSES_WITH_METRIC = [C.CROSS_ENTROPY, C.SMOOTHED_CROSS_ENTROPY]` in order to check that `self.config.config_loss.type in OPTIMIZER_LOSSES_WITH_METRIC` to get a better error message if we ever add new losses, which don't have a corresponding metric. Or is the assumption that every loss will have a corresponding metric? If so this should be documented in `class Loss`. Or we could make it explicit by:
```
class Loss(ABC):
   ...
   @abstract_method
   def create_metric(self):
       pass
```

sockeye/arguments.py
```diff
@@ -529,9 +548,17 @@ def add_training_args(params):
...
     train_params.add_argument('--optimizer',
-                              default='adam',
-                              choices=['adam', 'sgd', 'rmsprop'],
+                              default=C.OPTIMIZER_ADAM,
+                              choices=C.OPTIMIZERS,
                               help='SGD update rule. Default: %(default)s.')
+    train_params.add_argument('--optimizer-params',
+                              type=json_dict(),
+                              default=None,
+                              help='Additional optimizer params as JSON dictionary, should only be used for testing.')
```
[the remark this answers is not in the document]
mjdenkowski: It turns out this is actually a problematic format since it uses both quoting and spacing, making it less than ideal for scripts.  I'm going to work out a new format that we can use as the Plan A for passing args to optimizers.

sockeye/training.py
```diff
@@ -281,6 +282,10 @@ def _fit(self,
         """
         metric_train = self._create_eval_metric(metrics)
         metric_val = self._create_eval_metric(metrics)
+        # Track loss for optimizer if needed
+        if isinstance(self.module._curr_module._optimizer, SockeyeOptimizer):
```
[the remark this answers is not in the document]
mjdenkowski: Good to know!

sockeye/training.py
mjdenkowski: Good idea.  That looks cleaner.

mjdenkowski:
@szha That would be great!  The main piece that we would need is a way to get training loss along with the gradients when `update()` is called.  A lot of the code here is just getting that signal to the optimizer.

@tdomhan @fhieber The last commit refactors Loss so that each class creates an EvalMetric version.  Right now all of the tests pass, but I still want to add a better format for specifying optimizer params.

sockeye/arguments.py
```diff
@@ -529,9 +562,17 @@ def add_training_args(params):
...
     train_params.add_argument('--optimizer',
-                              default='adam',
-                              choices=['adam', 'sgd', 'rmsprop'],
+                              default=C.OPTIMIZER_ADAM,
+                              choices=C.OPTIMIZERS,
                               help='SGD update rule. Default: %(default)s.')
+    train_params.add_argument('--optimizer-params',
+                              type=simple_dict(),
+                              default=None,
+                              help='Additional optimizer params as dictionary. Format: key1:value1,key2:value2,...')
```
tdomhan: I like the idea of a simple format instead of json here, as we anyway only need a flat list of keys and values. Just a as a ntoe: `key=value` would seem like the more natural format.

sockeye/loss.py
```diff
@@ -27,40 +29,49 @@ class LossConfig(config.Config):
...
 
-    :param type: Loss name.
+    :param name: Loss name.
     :param vocab_size: Target vocab size.
     :param normalize: Whether to normalize loss value.
     :param smoothed_cross_entropy_alpha: Smoothing value for smoothed-cross-entropy loss.
     """
     def __init__(self,
-                 type: str,
+                 name: str,
```
tdomhan: this is a backwards-incompatible change. You should bump the Sockeye version.

sockeye/arguments.py
mjdenkowski: One of the challenges is that shells like to do their own thing with characters that look like argument indicators.  Even trying to specify a negative number like `key:-1` causes problems.  This probably wants another iteration.

sockeye/loss.py
mjdenkowski: Good catch.  When the rest of the PR is ready to go, I'll add a version increment commit.

mjdenkowski:
A few more updates:
- An option to use the update rules from "Nadam" (Nesterov Adam) in place of regular Adam, based on the new implementation in MXNet: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer.py#L934.
- An option to use the optimized metric instead of the loss, as some tests showed Perplexity to outperform CrossEntropy.
- All options for Eve are now folded into the new `--optimizer-params` training arg.

sockeye/loss.py
```diff
@@ -110,33 +128,58 @@ def get_loss(self, logits: mx.sym.Symbol, labels: mx.sym.Symbol) -> List[mx.sym.
...
+            # Ignore padding
+            ignore = (label == C.PAD_ID).astype(dtype=prob.dtype)
+            prob = prob * (1 - ignore) + ignore
+            # Sum, normalizing if needed
+            loss = -mx.nd.log(prob + 1e-8)
+            if self.loss_config.normalize:
+                loss = loss / mx.nd.sum(1 - ignore)
+                self.num_inst += 1
+            else:
+                self.num_inst += label.size - mx.nd.sum(ignore).asscalar()
```
tdomhan: This metric should reflect what the corresponding loss is doing, no? When we have `normalize=False` then the loss is divided by the batch size. Unless I'm getting this wrong here you add the number of non-pad tokens in the batch to the denominator. This is available through `label.shape[0]` and also actually what the mxnet [CrossEntropy](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/metric.py#L918) metric does.

How about contributing this back to MXNet in the long term? Ignore pad symbols seems like a generic modification. We'd just need to remove the dependency on LossConfig, which should be easy. We don't need to do this now, but could add a `TODO` to not forget.

sockeye/loss.py
```diff
@@ -148,23 +191,58 @@ def get_loss(self, logits: mx.sym.Symbol, labels: mx.sym.Symbol) -> List[mx.sym.
...
+        sces = preds[::2]
+        for label, sce in zip(labels, sces):
+            # SCE is pre-computed, so just sum
+            self.sum_metric += sce.asnumpy().sum()
+            # Only scale if loss is not already normalized
+            if self.loss_config.normalize:
+                self.num_inst += 1
+            else:
+                ignore = (label == C.PAD_ID).astype(dtype=sce.dtype)
+                self.num_inst += label.size - mx.nd.sum(ignore).asscalar()
```
tdomhan: same thing here. If `normalize=False` we should normalize by the batch size. This is btw done through the `rescale_grad` argument to the optimizer in [train.py](https://github.com/awslabs/sockeye/blob/master/sockeye/train.py#L533). We set this argument explicitly, but it is also what MXNet does by default.

sockeye/loss.py
mjdenkowski: By default, metrics add up the loss and number of instances (tokens, with or without ignoring padding based on the metric).  The special case here is for _not_ doing this when we've already normalized, instead adding the loss and incrementing the number of examples by 1 since the loss is already divided by the number of tokens.

Good idea contributing this to MXNet--I'll add a TODO.

sockeye/loss.py
mjdenkowski: Same as above.  Right now we're doing the same thing in a few different ways and a few different places.  It would be great if MXNet's loss and metrics were more synchronized.

mjdenkowski:
This should be ready to go.  The only remaining item is a version bump which we can add prior to merge.

sockeye/loss.py
tdomhan: Yeah I get that it makes sense to divide by one in the case `normalize=True`. My point was that in the other case you should divide by the number of elements in the batch, as is done in the corresponding [CrossEntropy](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/metric.py#L918) metric in mxnet, as this is effectively the loss that is optimized when `normalize=False`. So something like:
```
self.num_inst += label.shape[0]
```
This way the training loss and the metric would be consistent.

sockeye/loss.py
tdomhan: Well in our case we know that with `normalize=False` the loss is something like:
```
loss = 1./batch_size * sum(loss(next_word) for next_word in batch)
```
and I think it would make sense to define the metric accordingly.

sockeye/loss.py
mjdenkowski: `shape[0]` should be equivalent to `size` after calling `ravel()`.  Are you suggesting that we don't subtract the number of padding tokens here so we're consistent with MXNet?

sockeye/optimizers.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Optimizer that has access to additional information from the last batch and the last chekpoint
+    when updating weights.
+
+    :param request_optimized_metric: Whether to request the optimized metric (e.g. perplexity) in
+                                     place of optimizer loss (e.g. cross-entropy).
+    """
+    def __init__(self, request_optimized_metric: bool = False, **kwargs) -> None:
+        self.request_optimized_metric = request_optimized_metric
+        self.batch_state = Optional[BatchState]
```
tdomhan: `Optional` is meant to be used as a type annotation, but shouldn't be instantiated and assigned. If you want to declare the type of the member variable you can use:
```
self.batch_size = None # type: Optional[BatchState]
```
FYI: [PEP 526](https://www.python.org/dev/peps/pep-0526/) will allow for variable annotations, but requires python 3.6 and we currently support >=3.4:
```
self.batch_size: Optional[BatchState] = None
```

Optional means a variable might be `None`, so these variables should just be initialized to `None` here.

sockeye/loss.py
mjdenkowski: I see what you're saying now.  With `normalize=False`, we divide by batch size.  MXNet metrics divide by number of elements, sometimes ignoring padding.  Right now we're consistent with metrics whereas if we're using this as training feedback, we want to be consistent with loss.  Makes sense, though it also sounds like the best way to avoid problems is to always use `normalize=True` : )  I'll update and add some comments to this effect.

sockeye/optimizers.py
```diff
@@ -0,0 +1,244 @@
...
+        assert isinstance(weight, NDArray)
+        assert isinstance(grad, NDArray)
+        lr = self._get_lr(index)
+        wd = self._get_wd(index)
+        self._update_count(index)
+
+        t = self._index_update_count[index]
+
+        # Preprocess grad
+        grad *= self.rescale_grad + wd * weight
```
tdomhan: `wd` is the weight decay, no? If so it shouldn't be multiplied by the gradient, right?
But rather be something like:
```
grad = grad * self.rescale_grad + wd * weight
```
Or apply weight decay later after clipping.

mjdenkowski:
I updated the metrics to be consistent with loss.  Side note: do we have any use case for running with `normalize=False`?

tdomhan:
to me it's not entirely clear whether normalization by batch size or by word is better and I do sometimes get better performance with the former and would therefore still keep `normalize=False` around. I do however think it would be appropriate to rename this parameter, as we always normalize one way or another (batch size vs number of words).

sockeye/optimizers.py
mjdenkowski: Good catch!  I took this nice compact line from MXNet's Nadam implementation which actually had that typo (`grad *= ...` vs `grad = grad * ...`).  For clipping, there's actually inconsistent ordering in MXNet.  For SGD, we have `rescaled_grad[row] = lr * rescale_grad * clip(grad[row], clip_gradient) + wd * weight[row]`.  For Adam, we have `rescaled_grad[row] = clip(grad[row] * rescale_grad + wd * weight[row], clip_gradient)`.  I'm using the order for Adam here.

sockeye/optimizers.py
```diff
@@ -0,0 +1,244 @@
...
+"""
+Extra optimizers not included in MXNet.
+"""
+
+from abc import abstractmethod
+from collections import namedtuple
+import math
+from typing import Optional, Tuple
+
+from mxnet.ndarray import NDArray, sqrt, zeros_like
```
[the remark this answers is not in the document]
mjdenkowski: The original style follows MXNet's `optimizer.py` for easy comparison.  At this point we've made enough changes that I agree it makes more sense to use Sockeye's code style.

sockeye/optimizers.py
mjdenkowski: Thanks!  Felix and I were going back and forth looking for something that would make all versions of Python happy.

PR #146, opened 2017-09-19
Training resumption with non-empty model dir.

tdomhan:
Do not refuse training when no model state is found. So far the training
just died if the model directory already existed but did not contain
a training state. I do not see any good reason to not just start training
from scratch from the existing directory.

Additionally this commit adds a nicer error message for decoding when no
parameter file is found in the model directory.

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -150,7 +150,7 @@ def add_io_args(params):
                              help='Folder where model & training results are written to.')
     data_params.add_argument('--overwrite-output',
                              action='store_true',
-                             help='Overwrite output folder if it exists.')
+                             help='Delete all contents of the model directory if it already exists.')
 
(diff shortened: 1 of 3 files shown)


davvil:
The assumption is that an existing training dir without training state is actually a finished training, and we don't want to overwrite it (the training state dir is deleted when training ends succesfully). The goal was to avoid overwriting finished systems by mistake. You could change the logic to something like "if not exists(traning_state_dir) and exists(params file) then refuse overwriting".

tdomhan:
makes sense. With what you're proposing we get the best of both options: resume if training stopped before the first checkpoint but do not overwrite finished trainings. I'll change the PR accordingly.

tdomhan:
updated the PR, let me know if it looks ok now :)

PR #147, opened 2017-09-19
Refactored main function in train.py

davvil:


sockeye/train.py
```diff
@@ -79,27 +84,29 @@ def _dict_difference(dict1: Dict, dict2: Dict):
...
-                    "Must optimize either BLEU or one of tracked metrics (--metrics)")
+def check_resume(args: argparse.Namespace, output_folder: str) -> Tuple[bool, str]:
+    """
+    Check if we should resume a broken training run.
 
-    # Checking status of output folder, resumption, etc.
+    :param args: Arguments as returned by argparse.
+    :param output_folder: Main output folder for the model.
+    :return: Flag signaling if we are resuming training and the directory with
+        the training status
```
fhieber: nitpick: missing period

sockeye/train.py
```diff
@@ -79,27 +84,29 @@ def _dict_difference(dict1: Dict, dict2: Dict):
...
+    """
     # Create temporary logger to console only
     logger = setup_main_logger(__name__, file_logging=False, console=not args.quiet)
```
fhieber: why create the temporary logger here again? You added it outside of main at the top.
You might need to use global logger in the main then...

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
     else:
         os.makedirs(output_folder)
 
-    logger = setup_main_logger(__name__,
-                               file_logging=True,
-                               console=not args.quiet, path=os.path.join(output_folder, C.LOG_NAME))
+    return resume_training, training_state_dir
+
+
+def log_basic_info(args) -> None:
```
fhieber: this function could be resued by other clis, maybe move this to utils.py?

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
+    """
+    Create the encoder config.
+
+    :param args: Arguments as returned by argparse.
+    :param vocab_source_size: The source vocabulary.
+    :param config_conv: The config for the convolutional encoder (optional).
+    :return: The encoder config.
+    """
+    encoder_num_layers, _ = args.num_layers
+    config_encoder = Config()  # Make mypy happy
```
fhieber: why not `config_encoder = None  # type: Optional[Config]` ?

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
+                                     config_decoder=config_decoder,
+                                     config_loss=config_loss,
+                                     lexical_bias=args.lexical_bias,
+                                     learn_lexical_bias=args.learn_lexical_bias,
+                                     weight_tying=args.weight_tying,
+                                     weight_tying_type=args.weight_tying_type if args.weight_tying else None)
+    return model_config
+
+
+def create_training_model(model_config: model.ModelConfig, args: argparse.Namespace,
```
fhieber: maybe newline after each argument for easy reading

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
+                          context: List[mx.Context], train_iter: data_io.ParallelBucketSentenceIter,
+                          lr_scheduler_instance: lr_scheduler.LearningRateScheduler,
+                          resume_training: bool, training_state_dir: str) -> training.TrainingModel:
+    """
+    Create a training model and load the parameters from disk if needed
```
fhieber: missing period

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
+    elif args.params:
+        logger.info("Training will initialize from parameters loaded from '%s'", args.params)
+        training_model.load_params_from_file(args.params)
+
+    return training_model
+
+
+def define_optimizer(args, lr_scheduler_instance) -> Tuple[str, Dict]:
+    """
+    Defines the optimizer to use and its parameters
```
fhieber: missing period

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
 
-            # write vocabularies
-            vocab.vocab_to_json(vocab_source, os.path.join(output_folder, C.VOCAB_SRC_NAME) + C.JSON_SUFFIX)
-            vocab.vocab_to_json(vocab_target, os.path.join(output_folder, C.VOCAB_TRG_NAME) + C.JSON_SUFFIX)
+def main():
+    params = argparse.ArgumentParser(description='CLI to train sockeye sequence-to-sequence models.')
+    arguments.add_train_cli_args(params)
+    args = params.parse_args()
+
+    seedRNGs(args)
```
fhieber: seedRNGs could probably be moved to utils.py. It might be useful for other CLIs at some point.

sockeye/train.py
```diff
@@ -132,278 +139,460 @@ def main():
...
+
+    if args.use_fused_rnn:
```
fhieber: why not move these checks into a function as well? Something like `check_args_compatibility()`?

sockeye/train.py
davvil: Right! I forgot to remove it here!

sockeye/train.py
davvil: Because I was not so familiar with mypy :-D Will do!

sockeye/utils.py
```diff
@@ -80,6 +82,29 @@ def parse_version(version_string: str) -> Tuple[str, str, str]:
     return release, major, minor
 
 
+def log_basic_info(args) -> None:
```
fhieber: lets use this in translate.py as well

sockeye/train.py
```diff
@@ -132,278 +139,446 @@ def main():
...
+
+    :param args: Arguments as returned by argparse.
+    :param vocab_target_size: The size of the target vocabulary.
+    :return: The config for the decoder.
+    """
+    _, decoder_num_layers = args.num_layers
+
+    decoder_weight_tying = args.weight_tying and C.WEIGHT_TYING_TRG in args.weight_tying_type \
+        and C.WEIGHT_TYING_SOFTMAX in args.weight_tying_type
+    config_decoder = Config()
```
fhieber: could also initialize to None here again

sockeye/utils.py
davvil: Done

sockeye/train.py
```diff
@@ -132,278 +139,446 @@ def main():
...
+        logger.info("Device: CPU")
+        context = [mx.cpu()]
+    else:
+        num_gpus = utils.get_num_gpus()
+        check_condition(num_gpus >= 1,
+                        "No GPUs found, consider running on the CPU with --use-cpu "
+                        "(note: check depends on nvidia-smi and this could also mean that the nvidia-smi "
+                        "binary isn't on the path).")
+        if args.disable_device_locking:
+            context = utils.expand_requested_device_ids(args.device_ids)
```
tdomhan: not a blocker, but if we're already factoring out the different parts into individual functions it would be nice to make the dependencies explicit, by passing in the required arguments instead of the entire namespace, e.g.:
```
def determine_context(use_cpu, disable_device_locking, device_ids, lock_dir, exit_stack: ExitStack) -> List[mx.Context]:
```
This way the individual functions could be used independent of argparse.
Anyway, we don't have to do this now and the current refactoring is also very valuable :)

sockeye/train.py
```diff
@@ -132,278 +139,446 @@ def main():
...
+        context = [mx.gpu(gpu_id) for gpu_id in context]
+    return context
+
+
+def load_or_create_vocabs(args: argparse.Namespace, resume_training: bool, output_folder: str) -> Tuple[Dict, Dict]:
+    """
+    Load the vocabularies from disks if given, create them if not.
+
+    :param args: Arguments as returned by argparse.
+    :param resume_training: Gives if an interrupted training should be resumed.
```
tdomhan: "Gives if" -> "When True the vocabulary will be loaded from an existing output folder."

sockeye/train.py
davvil: I'm ambivalent about this one. On the one hand it's good to remove the dependency of argparse. On the other hand, I find it convenient to pass the whole arguments and let the function pick what it needs. In this way we reduce the number of places that must be aware of the parameters. Think about the unpacking we would need to do for create_model_config.

PR #149, opened 2017-09-21
Added CLI to build vocabulary

fhieber:


sockeye/vocab.py
```diff
@@ -147,3 +150,24 @@ def reverse_vocab(vocab: Mapping) -> Dict:
...
     """
     return {v: k for k, v in vocab.items()}
+
+
+def main():
+    params = argparse.ArgumentParser(description='CLI to build source and target vocab(s).')
+    arguments.add_build_vocab_args(params)
+    args = params.parse_args()
+
+    num_words, _ = args.num_words
```
tdomhan: would it make sense to assert that both values are identical to make it explicit that for this CLI only one value should be provided?

fhieber: good idea, will do that

PR #152, opened 2017-09-21
Word batching update

mjdenkowski:
This update simplifies memory handling for bucket-specific batch sizes by guaranteeing that the default bucket (largest by sequence length) will also have the largest number of total elements of any batch.  If any other batch has more elements than the default, we bump the default batch size up by the number of devices.  We can then just use the default bucket keys and batch size for `provide_data` and MXNet should do the right thing.  This change resolves the re-bucketing warnings and overall GPU memory usage is lower.

The price for this is that the batch size for the largest bucket gets bumped up slightly, but the effect is relatively minor with reasonable batch sizes.  On an example data set with a requested batch size of 4000, other buckets range from 3800-4200 while the largest is about 4300.

### DIFF
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -21,7 +21,7 @@ import math
 import pickle
 import random
 from collections import OrderedDict
-from typing import Any, Dict, Iterator, Iterable, List, Optional, Tuple
+from typing import Any, Dict, Iterator, Iterable, List, NamedTuple, Optional, Tuple
 
 import mxnet as mx
 import numpy as np
@@ -360,6 +360,16 @@ def get_parallel_bucket(buckets: List[Tuple[int, int]],
     return bucket
 
 
+BucketBatchSize = NamedTuple("BucketBatchSize", [
+    ("batch_size", int),
+    ("average_words_per_batch", float)
+])
+"""
+:param batch_size: Number of sentences in each batch.
+:param average_words_per_batch: Approximate number of non-padding tokens in each batch.
+"""
+
+
 # TODO: consider more memory-efficient data reading (load from disk on demand)
 # TODO: consider using HDF5 format for language data
 class ParallelBucketSentenceIter(mx.io.DataIter):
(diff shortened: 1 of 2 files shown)


fhieber:
Have you tried this with sentence-based batches? It seems that this change also affects the default bucket keys in `provide_*` if batch-type == sentence. Does it change memory consumption or behavior in this case?

mjdenkowski:
When batching by sentences, each bucket's batch size is set as `batch_size_seq = self.batch_size`, so the final size check should be a no-op.  I added an`if` statement to make that explicit.

sockeye/data_io.py
```diff
@@ -422,30 +422,29 @@ def __init__(self,
...
         self._convert_to_array()
 
-        # "Staging area" that needs to fit any size batch we're using by total number of elements
+        # "Staging area" that needs to fit any size batch we're using by total number of elements.
+        # When computing per-bucket batch sizes, we guarantee that the default bucket will have the
+        # largest total batch size.
         self.provide_data = [
             mx.io.DataDesc(name=source_data_name,
-                           shape=(self.default_bucket_batch_size, self.default_bucket_key[0]),
+                           shape=(self.bucket_batch_sizes[-1][0], self.default_bucket_key[0]),
```
tdomhan: I'd still advocate for making the elements of `bucket_batch_sizes ` a `namedtuple("bucket_batch_size", ["batch_size", "word_per_batch"])`, to make `[-1][0]` a little clearer by using `[-1].batch_size` instead.

sockeye/data_io.py
```diff
@@ -577,18 +572,17 @@ def _populate_bucket_batch_sizes(self):
                 batch_size_seq = self.batch_size
                 batch_size_word = batch_size_seq * average_seq_len
             self.bucket_batch_sizes[buck_idx] = (batch_size_seq, batch_size_word)
-        # Default bucket batch size: batch size of largest bucket batch by total elements, scaled for default bucket key
-        # Effectively, set default values to reserve enough space to provide a batch from any bucket
-        largest_total_batch_size = 0
+            # Track largest batch size by total elements
+            largest_total_batch_size = max(largest_total_batch_size, batch_size_seq * max(*bucket_shape))
+        # Final step: guarantee that largest bucket by sequence length also has largest total batch size.
```
tdomhan: it might be worth pointing out that this is just a heuristic and does not guarantee full memory reuse across buckets (as this would require the default bucket to consist of both the maximum batch size and the maximum sentence length), which would allocate a lot of memory, but at the same time make sure all ndarrays from the default executor can be shared.

mjdenkowski: Good idea to make that clear in the code!

sockeye/data_io.py
mjdenkowski: Good idea now that we aren't modifying the values elsewhere.

Issue #154, opened 2017-09-23
Recommended paramters for transformer model


emjotde:
Hi,
what are your recommended command line options for the transformer model? I would like to see if I can reproduce the results from tensor2tensor with sockeye. My first attempts with your transformer implementation do not really seem to converge.

My current options:
```
python3 -m sockeye.train -s train.en \                                                                                                                                                                                                 
                         -t train.de \                                                                                                                                                                                                 
                         -vs valid.en \                                                                                                                                                                                                
                         -vt valid.de \                                                                                                                                                                                                
                         --num-embed 512 \                                                                                                                                                                                             
                         --rnn-num-hidden 2048 \                                                                                                                                                                                       
                         --attention-type dot \                                                                                                                                                                                        
                         --max-seq-len 50 \                                                                                                                                                                                            
                         --encoder transformer --decoder transformer \                                                                                                                                                                 
                         --num-layers 6 --layer-normalization \                                                                                                                                                                        
                         --weight-tying-type src_trg_softmax --weight-tying \                                                                                                                                                          
                         -o wmt_model
```
I guess some learning rate magic might be needed?

fhieber:
Hi,

`--rnn-*` and `--attention-*` flags do not apply for the transformer and you are right, some learning rate scheduling should help with this (as written in the original paper). Also, layer normalization is always enabled for the transformer in the latest version (unless you choose a `--transformer-pre/postprocess` string not containing 'n')

You can try the following:
```
python3 -m sockeye.train -s train.en \                                                                                                                                                                                                 
                         -t train.de \                                                                                                                                                                                                 
                         -vs valid.en \                                                                                                                                                                                                
                         -vt valid.de \                                                                                                                                                                                                
                         --num-embed 512 \                                                                                                                                                                                                                                                                                                                                                                      
                         --max-seq-len 50 \                                                                                                                                                                                            
                         --encoder transformer --decoder transformer \                                                                                                                                                                 
                         --num-layers 6 --layer-normalization \          
                         --transformer-model-size 512 --transformer-feed-forward-num-hidden 2048 \
                         --transformer-dropout-prepost 0.1 \                                                                                                                                                         
                         --weight-tying-type src_trg_softmax --weight-tying \
                         --clip-gradient -1 --initial-learning-rate 0.0003 --learning-rate-warmup 50000 --normalize-loss \                                                                                                                                                         
                         -o wmt_model
```

[...]

PR #156, opened 2017-10-05
Always using utf-8 for vocab to avoid problems with LC_ALL and LANG environment variables.

tdomhan:
Since we changed the vocab to utf-8 we have had several issues where people did not have their  `LANG` or `LC_ALL` environment variables set to `C.UTF_8`. This caused `open` to choose `ascii` as an encoding, which in turn lead to problems because we use `ensure_ascii=False`. With this small change we make sure we always use utf-8 for reading and writing the vocab.


tdomhan:
For example Marcin had this problem as reported here: https://github.com/awslabs/sockeye/issues/154.

PR #159, opened 2017-10-10
Decode in chunks instead of truncating.

tdomhan:
Any model trained with learned positional encodings is limited to a maximum input length that guarantees to never exceed the maximum output length seen during training. Additionally for other models it also makes sense to never translate any sentence longer than what was seen during training. Before we truncated sentences that were too long. With this change we will split the long sentence into chunks that are digestible (max length), translate each individually and concatenate the output.


sockeye/inference.py
```diff
@@ -462,18 +470,21 @@ class LengthPenalty:
     def __init__(self, alpha: float = 1.0, beta: float = 0.0) -> None:
         self.alpha = alpha
         self.beta = beta
-        self.denominator = (self.beta + 1) ** self.alpha
+        self.denominator = float((self.beta + 1.) ** self.alpha)
```
mjdenkowski: Is `float()` needed if both operands are floats?

sockeye/inference.py
```diff
@@ -631,11 +644,51 @@ def _make_result(self,
...
+
+    def _concat_translations(self, translations: List[Translation]) -> Translation:
+        """
+        Combine translations through concatenation.
+
+        :param translations: A list of translations (sequence, attention_matrix), score and length.
+        :return: A concatenation if the translations with a score.
+        """
+        # Concatenation of all target ids without BOS and EOS
+        target_ids = list(itertools.chain.from_iterable(translation.target_ids[1:-1] for translation in translations))
```
mjdenkowski: Do we guarantee that 0 and -1 will be BOS and EOS at this point regardless of model settings and think this will hold for all future models?

sockeye/inference.py
```diff
@@ -631,11 +644,51 @@ def _make_result(self,
...
+            else:
+                # Alignment without neither BOS nor EOS
+                attention_matrix_combined[pos_t:pos_t + len_t-2, pos_s:pos_s + len_s] = attention_matrix[1:-1,:]
+                pos_t += len_t-2
+                pos_s += len_s
+
+        # Unnormalize + sum and renormalize:
+        score = sum(translation.score * self.length_penalty(len(translation.target_ids))
+                    for translation in translations)
+        score = score / self.length_penalty(len(target_ids))
```
mjdenkowski: We could be clearer by keeping sufficient statistics for scores and length penalty, effectively our feature set.  Each `Translation` could have `Scores` with clearly defined sum and eval operations.  That can probably wait, but we could add a TODO to refactor this part if we add more score components.

sockeye/inference.py
```diff
@@ -462,18 +470,21 @@ class LengthPenalty:
     def __init__(self, alpha: float = 1.0, beta: float = 0.0) -> None:
         self.alpha = alpha
         self.beta = beta
-        self.denominator = (self.beta + 1) ** self.alpha
+        self.denominator = float((self.beta + 1.) ** self.alpha)
 
-    def __call__(self, lengths: mx.nd.NDArray) -> mx.nd.NDArray:
+    def __call__(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray, float]:
```
mjdenkowski: I hadn't seen the `Union` type before.  That's very useful!

sockeye/inference.py
tdomhan: yeah it's probably not necessary especially as with python3.6 there's a special int division operator `//`. I'll revert this.

sockeye/inference.py
tdomhan: yeah, it's a nice feature :)

sockeye/inference.py
tdomhan: that's a good point. As this is a "private" method we're somewhat save, but I agree that this is something that could easily change in the future. I'll add an assertion to fail early if we every change this behavior. The BOS is guaranteed through the way `_beam_search` is set up, see [see](https://github.com/awslabs/sockeye/blob/master/sockeye/inference.py#L733). However now that I think about it, if we reach the max length and the top hypothesis has not finished yet we might actually drop the last word. I'll update the code accordingly.

sockeye/inference.py
tdomhan: yeah, that would make a lot of sense.

sockeye/inference.py
fhieber: Yes, there is no guarantee that the last word is EOS

sockeye/inference.py
fhieber: However, that would require yet another set of variables we need to run sort/take operations on, which tend to be a bit slow. The inplace renormalization is probably faster.

sockeye/inference.py
```diff
@@ -581,7 +592,16 @@ def translate(self, trans_input: TranslatorInput) -> TranslatorOutput:
                                     attention_matrix=np.asarray([[0]]),
                                     score=-np.inf)
 
-        return self._make_result(trans_input, *self.translate_nd(*self._get_inference_input(trans_input.tokens)))
+        if len(trans_input.tokens) > self.max_input_length:
+            logger.warning("Input (%d) exceeds max bucket size (%d). Splitting into chunks of size %d.",
```
fhieber: bucket size is actually a bit confusing to the user I think. Can we rename to max input length or something?

sockeye/inference.py
```diff
@@ -631,11 +644,51 @@ def _make_result(self,
...
+        """
+        Combine translations through concatenation.
+
+        :param translations: A list of translations (sequence, attention_matrix), score and length.
+        :return: A concatenation if the translations with a score.
+        """
+        # Concatenation of all target ids without BOS and EOS
+        target_ids = list(itertools.chain.from_iterable(translation.target_ids[1:-1] for translation in translations))
+        # Adding back the EOS symbol
+        target_ids = [self.vocab_target[C.BOS_SYMBOL]] + target_ids + [self.vocab_target[C.EOS_SYMBOL]]
```
fhieber: can we use itertools.chain here as well for speed?

sockeye/inference.py
fhieber: Instead of duplicating a lot of EOS/BOS removal logic and handling attention matrices, would it make more sense to move concatenation of translations AFTER the make_result calls? make_result is already handling this stuff and I am afraid that we will add too many places that somehow deal with stripping special symbols from the output.

sockeye/inference.py
tdomhan: yeah, makes sense.

sockeye/inference.py
tdomhan: changed the code anyway to address michael's comment.

sockeye/inference.py
tdomhan: in make_result we already concatenate to strings, which makes the concat more difficult. I wanted to write the concat method in a way to produce an output which is always compatible to make_result, in that the translations still start with BOS for both the target_ids and the attention_matrix.

test/unit/test_inference.py
```diff
@@ -17,6 +17,31 @@
 import sockeye.inference
 
 
+_BOS = 0
```
fhieber: Would it make sense to put this into constants or make it equivalent to whats actually happening? its already implicitly there in VOCAB_SYMBOLS: EOS is index 3 in VOCAB_SYMBOLS so EOS always gets id 3

tdomhan: I left it in the test, as the function should be independent of the concrete values of BOS and EOS defined elsewhere.

fhieber: Ok

PR #164, opened 2017-10-12
Start from best parameters (and optimizer states) after updating learning rates if not improved

fhieber:
This is a small change but potentially quite useful. With this PR we reset the model parameters during training back to the current best params when adjusting the learning rate. That means we will continue from the currently best known point with a lower learning rate. Before we did (by default) 3000 updates with the larger learning rate and potentially moved away from the current best before continuing with a smaller learning rate.

I tested this on a small dataset and final validation perplexity was a bit lower (-0.3) than the baseline.
With the change, the model was trained longer (more checkpoints) and train perplexity was overall higher, suggesting less overfitting to train.

Open question: Should we also reset the optimizer states (e.g. for Adam) in these cases? Might not be as easy to do since we only keep one copy of them (they tend to be fairly large). In any case, that should be another PR.

### DIFF
--- a/sockeye/lr_scheduler.py
+++ b/sockeye/lr_scheduler.py
@@ -29,8 +29,11 @@ class LearningRateScheduler:
         self.log_warmup_every_t = self.warmup // 10
         self.last_warmup_log = -1
 
-    def new_evaluation_result(self, has_improved: bool):
-        pass
+    def new_evaluation_result(self, has_improved: bool) -> bool:
+        """
+        Returns true if learning rate was adjusted.
+        """
+        return False
 
     def __call__(self, num_updates):
         pass
@@ -71,7 +74,7 @@ class LearningRateSchedulerFixedStep(LearningRateScheduler):
         self.latest_t = 0
         self._update_rate(self.current_step)
 
-    def new_evaluation_result(self, has_improved: bool):
+    def new_evaluation_result(self, has_improved: bool) -> bool:
         logger.info("Checkpoint learning rate: %1.2e (%d/%d updates)",
                     self.current_rate,
                     self.latest_t - self.current_step_started_at,
@@ -79,6 +82,7 @@ class LearningRateSchedulerFixedStep(LearningRateScheduler):
         if self.latest_t >= self.next_step_at:
             self.current_step += 1
             self._update_rate(self.current_step)
+        return False
 
     def _update_rate(self, step: int):
         if self.current_step < len(self.schedule):
@@ -190,7 +194,7 @@ class LearningRateSchedulerPlateauReduce(LearningRateScheduler):
                     " the validation score doesn't improve %d times.",
                     reduce_factor, reduce_num_not_improved)
 
-    def new_evaluation_result(self, has_improved: bool):
+    def new_evaluation_result(self, has_improved: bool) -> bool:
         if self.lr is None:
             assert self.base_lr is not None
             self.lr = self.base_lr
@@ -204,6 +208,8 @@ class LearningRateSchedulerPlateauReduce(LearningRateScheduler):
                 logger.info("%d checkpoints since improvement or rate scaling, "
                             "lowering learning rate: %1.2e -> %1.2e", self.num_not_improved, old_lr, self.lr)
                 self.num_not_improved = 0
+                return True
+        return False
 
     def __call__(self, t):
         if self.lr is None:
--- a/sockeye/lr_scheduler.py
+++ b/sockeye/lr_scheduler.py
@@ -31,7 +31,10 @@ class LearningRateScheduler:
 
     def new_evaluation_result(self, has_improved: bool) -> bool:
         """
-        Returns true if learning rate was adjusted.
+        Returns true if the parameters should be reset to the ones with the best validation score.
+
+        :param has_improved: Whether the model improved on held-out validation data.
+        :return: True if parameters should be reset to the ones with best validation score.
         """
         return False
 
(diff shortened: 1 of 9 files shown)


sockeye/lr_scheduler.py
```diff
@@ -29,8 +29,11 @@ def __init__(self, warmup: int = 0) -> None:
         self.log_warmup_every_t = self.warmup // 10
         self.last_warmup_log = -1
 
-    def new_evaluation_result(self, has_improved: bool):
-        pass
+    def new_evaluation_result(self, has_improved: bool) -> bool:
+        """
+        Returns true if learning rate was adjusted.
```
tdomhan: looks good, but the naming is a little confusing as there are different lr scheduler that constantly adjust the learning rate, but will never return True. Why not make it explicit:
```
Returns true when the parameters should be reset to the ones with the best validation score.
```

sockeye/training.py
```diff
@@ -443,17 +447,24 @@ def _log_params(self):
...
         params_base_fname = C.PARAMS_NAME % checkpoint
         self.save_params_to_file(os.path.join(output_folder, params_base_fname))
 
+    def _load_params(self, output_folder: str, checkpoint: int):
+        """
+        Loads parameters from disk, sets self.params and module's parameters.
+        """
+        params_fname = os.path.join(output_folder, C.PARAMS_NAME % checkpoint)
+        self.load_params_from_file(params_fname)  # sets self.params
+        self.module.set_params(self.params, {})
```
tdomhan: ah the second argument are the aux_params, right? I was a little confused by this for a sec. Could you make it a named arg?
```
self.module.set_params(self.params, aux_params={})
```

sockeye/training.py
```diff
@@ -393,8 +393,12 @@ def _fit(self,
                     checkpoint_state = CheckpointState(checkpoint=train_state.checkpoint, metric_val=m_val)
                     optimizer.pre_update_checkpoint(checkpoint_state)
 
+                # learning rate adjustment
                 if self.lr_scheduler is not None:
-                    self.lr_scheduler.new_evaluation_result(has_improved)
+                    lr_adjusted = self.lr_scheduler.new_evaluation_result(has_improved)
+                    if lr_adjusted and not has_improved:
+                        logger.info("Loading parameters from last best checkpoint: %d", best_checkpoint)
```
mjdenkowski: We could leave ourselves a comment here like:

    # TODO: Try also loading or completely resetting optimizer state

sockeye/training.py
```diff
@@ -393,8 +393,12 @@ def _fit(self,
                     checkpoint_state = CheckpointState(checkpoint=train_state.checkpoint, metric_val=m_val)
                     optimizer.pre_update_checkpoint(checkpoint_state)
 
+                # learning rate adjustment
                 if self.lr_scheduler is not None:
-                    self.lr_scheduler.new_evaluation_result(has_improved)
+                    lr_adjusted = self.lr_scheduler.new_evaluation_result(has_improved)
+                    if lr_adjusted and not has_improved:
```
tdomhan: could we make this behavior optional? e.g. in a async sgd setup one might not want to reload the params here. We could make it the new default but have an option to turn it off.

mjdenkowski:
Great update!  My main suggestion would be to make this a CLI option until we run more experiments and also look at resetting the optimizer state for consistency.

For at least one other toolkit, Graham's [XNMT](https://github.com/neulab/xnmt), the default decay operation is rolling back parameters *and* completely resetting optimizer state (effectively starting a new training run).  We can also experiment with rolling back the optimizer state with a little more bookkeeping.

We could add add optimizer reset support in another update, test the various combinations, and have a CLI option for vanilla decay vs smart rollback.

davvil:
Most of the logic for storing and resetting the optimizer state should already be there for training resumption. It stores and restores from disk, but hopefully we can reuse the code.

sockeye/lr_scheduler.py
fhieber: Good idea, done

sockeye/training.py
fhieber: good point, done

fhieber:
Thanks for the comments, I updated the PR so we can experiment with the new settings (including the optimizer state reset).
From a quick test on seqcopy, it looks like setting the optimizer states back to the previous best instead of initial seems to look a bit more stable, but thats very preliminary.

sockeye/training.py
fhieber: done

sockeye/training.py
```diff
@@ -284,12 +292,15 @@ def _fit(self,
...
         """
         # TODO: Push update to MXNet to expose the optimizer (Module should have a get_optimizer method)
-        if self.bucketing:
-            optimizer = self.module._curr_module._optimizer
-        else:
-            optimizer = self.module._optimizer
+        optimizer = self.module._curr_module._optimizer if self.bucketing else self.module._optimizer
+        if lr_decay_opt_states_reset == C.LR_DECAY_OPT_STATES_RESET_INITIAL:
+            self.save_optimizer_states(os.path.join(output_folder, C.OPT_STATES_INITIAL))
+        #elif lr_decay_opt_states_reset == C.LR_DECAY_OPT_STATES_RESET_BEST:
```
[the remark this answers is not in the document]
fhieber: Already removed it completely as its actually not needed :)

sockeye/training.py
```diff
@@ -393,15 +402,31 @@ def _fit(self,
...
+                    best_params_path = os.path.join(output_folder, C.PARAMS_BEST_NAME)
+                    if os.path.lexists(best_params_path):
+                        os.remove(best_params_path)
+                    actual_best_params_fname = C.PARAMS_NAME % best_checkpoint
+                    os.symlink(actual_best_params_fname, best_params_path)
+                    if lr_decay_opt_states_reset == C.LR_DECAY_OPT_STATES_RESET_BEST:
+                        best_opt_states_fname = os.path.join(output_folder, C.OPT_STATES_BEST)
+                        if os.path.exists(best_opt_states_fname):
+                            os.remove(best_opt_states_fname)
+                        self.save_optimizer_states(best_opt_states_fname)
```
davvil: Minor: We will leave the optimizer state file in the output directory.

fhieber: good point, I will add deletion at the end and the merge.

PR #166, opened 2017-10-16
Revert to commented out system tests

fhieber:
I was trying to add a weekly travis cronjob for running system tests, but Travis has a timeout on long-running jobs. We should find another way of regularly running system tests...


tdomhan:
this cron event thing wasn't working or why the revert?

fhieber:
It did work (it triggered one cron-typed build) but timed out, as the travis build machine seems to be too slow to finished the system tests within the timeout of travis builds.

PR #169, opened 2017-10-17
Added PrintValue operator for debugging

davvil:


sockeye/utils.py
```diff
@@ -646,3 +646,46 @@ def get_validation_metric_points(model_path: str, metric: str):
     metrics_path = os.path.join(model_path, C.METRICS_NAME)
     data = read_metrics_file(metrics_path)
     return [(d['%s-val' % metric], cp) for cp, d in enumerate(data, 1)]
+
+
+class PrintValue(mx.operator.CustomOp):
```
fhieber: Can we make it log instead of print if given a logger? Or even cooler, it could optionally write to tensorboard if given a summary_writer in the constructor.

Also, it could optionally also log the gradients from the backward pass.

davvil: Does tensorboard support arbitrary tensors?

fhieber: we could pass a function that summarizes a tensor to a value, e.g. the norm. But maybe thats a separate PR.

PR #171, opened 2017-10-17
Exposing the kvstore.

tdomhan:


sockeye/train.py
```diff
@@ -161,6 +161,9 @@ def determine_context(args: argparse.Namespace, exit_stack: ExitStack) -> List[m
             context = utils.expand_requested_device_ids(args.device_ids)
         else:
             context = exit_stack.enter_context(utils.acquire_gpus(args.device_ids, lock_dir=args.lock_dir))
+        check_condition(args.batch_size % len(context) == 0, "When using multiple devices the batch size must be "
+                                                             "divisible by the number of devices. Choose a batch size "
+                                                             "that is a multiple of %d." % len(context))
```
mjdenkowski: We could limit this check to when `args.batch_type == C.BATCH_TYPE_SENTENCE`, since the case of `C.BATCH_TYPE_WORD` means we'll compute the sentence-based batch sizes later and make sure they fit the number of devices.

tdomhan: good point! I'll make that change.

tdomhan:
ok, I added a bunch of tests for things which actually do not work with distributed training (due to the optimizer running remotely on the server instead of in the local process).

tdomhan:
sounds good, done.

PR #172, opened 2017-10-19
Batch decoding for better speed

artemsok:
* Implemented decoding in batches. Decoding speed improvement up to 3 times (for similar length inputs)
* Reading inputs is now done in chunks (--chunk-size, default 10k). In the future a possible alternative would be to batch inputs (supplied as an iteratable) directly.
* Batch size controlled from CLI

### DIFF
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.8.4'
+__version__ = '1.9.0'
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -754,6 +754,15 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--batch-size',
+                               type=int_greater_or_equal(1),
+                               default=16,
+                               help='Batch size during decoding. Determines how many sentences are translated simultaneously.'
+                                    'Default: %(default)s.')
+    decode_params.add_argument('--chunk-size',
+                               type=int_greater_or_equal(1),
+                               default=10000,
+                               help='Size of the chunks to be read from input at once. Default: %(default)s.')
     decode_params.add_argument('--ensemble-mode',
                                type=str,
                                default='linear',
--- a/sockeye/checkpoint_decoder.py
+++ b/sockeye/checkpoint_decoder.py
@@ -108,19 +108,18 @@ class CheckpointDecoder:
 
     def decode_and_evaluate(self,
                             checkpoint: Optional[int] = None,
-                            output_name: str = os.devnull,
-                            speed_percentile: int = 99) -> Dict[str, float]:
+                            output_name: str = os.devnull) -> Dict[str, float]:
         """
         Decodes data set and evaluates given a checkpoint.
 
         :param checkpoint: Checkpoint to load parameters from.
         :param output_name: Filename to write translations to. Defaults to /dev/null.
-        :param speed_percentile: Percentile to compute for sec/sent. Default: p99.
         :return: Mapping of metric names to scores.
         """
         models, vocab_source, vocab_target = inference.load_models(self.context,
                                                                    self.max_input_len,
                                                                    self.beam_size,
+                                                                   16,  # batch size
                                                                    [self.model],
                                                                    [checkpoint],
                                                                    softmax_temperature=self.softmax_temperature,
@@ -133,20 +132,19 @@ class CheckpointDecoder:
                                           models,
                                           vocab_source,
                                           vocab_target)
-        trans_wall_times = np.zeros((len(self.input_sentences),))
+        trans_wall_time = 0.0
+        translations = []
         with data_io.smart_open(output_name, 'w') as output:
             handler = sockeye.output_handler.StringOutputHandler(output)
-            translations = []
-            for i, input_sentence in enumerate(self.input_sentences):
-                tic = time.time()
-                trans_input = translator.make_input(i, input_sentence)
-                trans_output = translator.translate(trans_input)
+            tic = time.time()
+            trans_inputs = [translator.make_input(i, line) for i, line in enumerate(self.input_sentences)]
+            trans_outputs = translator.translate(trans_inputs)
+            trans_wall_time = time.time() - tic
+            for trans_input, trans_output in zip(trans_inputs, trans_outputs):
                 handler.handle(trans_input, trans_output)
-                trans_wall_time = time.time() - tic
-                trans_wall_times[i] = trans_wall_time
                 translations.append(trans_output.translation)
-        percentile_sec_per_sent = np.percentile(trans_wall_times, speed_percentile)
+        avg_time = trans_wall_time / len(self.input_sentences)
 
         # TODO(fhieber): eventually add more metrics (METEOR etc.)
         return {C.BLEU_VAL: sockeye.bleu.corpus_bleu(translations, self.target_sentences),
-                C.SPEED_PCT % speed_percentile: percentile_sec_per_sent}
+                C.AVG_TIME: avg_time}
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -15,6 +15,7 @@
 Code for inference/translation
 """
 import logging
+import itertools
 import os
 from typing import Callable, Dict, List, NamedTuple, Optional, Tuple, Union, Set
 
@@ -52,6 +53,7 @@ class InferenceModel(model.SockeyeModel):
                  context: mx.context.Context,
                  fused: bool,
                  beam_size: int,
+                 batch_size: int,
                  checkpoint: Optional[int] = None,
                  softmax_temperature: Optional[float] = None,
                  max_output_length_num_stds: int = C.DEFAULT_NUM_STD_MAX_OUTPUT_LENGTH) -> None:
@@ -69,7 +71,7 @@ class InferenceModel(model.SockeyeModel):
 
         self.beam_size = beam_size
         self.softmax_temperature = softmax_temperature
-        self.encoder_batch_size = 1
+        self.batch_size = batch_size
         self.context = context
 
         self._build_model_components(fused)
@@ -205,7 +207,7 @@ class InferenceModel(model.SockeyeModel):
         :return: List of data descriptions.
         """
         return [mx.io.DataDesc(name=C.SOURCE_NAME,
-                               shape=(self.encoder_batch_size, bucket_key),
+                               shape=(self.batch_size, bucket_key),
                                layout=C.BATCH_MAJOR)]
 
     def _get_decoder_data_shapes(self, bucket_key: Tuple[int, int]) -> List[mx.io.DataDesc]:
@@ -219,8 +221,8 @@ class InferenceModel(model.SockeyeModel):
         source_max_length, target_max_length = bucket_key
         return self.decoder_data_shapes_cache.setdefault(
             bucket_key,
-            [mx.io.DataDesc(C.TARGET_NAME, (self.beam_size, target_max_length), layout="NT")] +
-            self.decoder.state_shapes(self.beam_size,
+            [mx.io.DataDesc(name=C.TARGET_NAME, shape=(self.batch_size * self.beam_size, target_max_length), layout="NT")] +
+            self.decoder.state_shapes(self.batch_size * self.beam_size,
                                       self.encoder.get_encoded_seq_len(source_max_length),
                                       self.encoder.get_num_hidden()))
 
@@ -233,7 +235,7 @@ class InferenceModel(model.SockeyeModel):
         Returns encoder representation of the source, source_length, initial hidden state of decoder RNN,
         and initial decoder states tiled to beam size.
 
-        :param source: Integer-coded input tokens.
+        :param source: Integer-coded input tokens. Shape (batch_size, source length).
         :param source_max_length: Bucket key.
         :return: Encoded source, source length, initial decoder hidden state, initial decoder hidden states.
         """
@@ -245,7 +247,7 @@ class InferenceModel(model.SockeyeModel):
         self.encoder_module.forward(data_batch=batch, is_train=False)
         decoder_states = self.encoder_module.get_outputs()
         # replicate encoder/init module results beam size times
-        decoder_states = [mx.nd.broadcast_axis(s, axis=0, size=self.beam_size) for s in decoder_states]
+        decoder_states = [mx.nd.repeat(s, repeats=self.beam_size, axis=0) for s in decoder_states]
         return decoder_states
 
     def run_decoder(self,
@@ -304,6 +306,7 @@ class InferenceModel(model.SockeyeModel):
 def load_models(context: mx.context.Context,
                 max_input_len: Optional[int],
                 beam_size: int,
+                batch_size: int,
                 model_folders: List[str],
                 checkpoints: Optional[List[int]] = None,
                 softmax_temperature: Optional[float] = None,
@@ -332,6 +335,7 @@ def load_models(context: mx.context.Context,
                                context=context,
                                fused=False,
                                beam_size=beam_size,
+                               batch_size=batch_size,
                                softmax_temperature=softmax_temperature,
                                checkpoint=checkpoint)
         models.append(model)
@@ -461,13 +465,13 @@ class ModelState:
         """
         self.states = [mx.nd.take(ds, best_hyp_indices) for ds in self.states]
 
-
 class LengthPenalty:
     """
     Calculates the length penalty as:
     (beta + len(Y))**alpha / (beta + 1)**alpha
 
-    See Wu et al. 2016.
+    See Wu et al. 2016 (note that in the paper beta has a different meaning, 
+    and a fixed value 5 was used for this parameter)
 
     :param alpha: The alpha factor for the length penalty (see above).
     :param beta: The beta factor for the length penalty (see above).
@@ -577,7 +581,8 @@ class Translator:
         self.models = models
         self.interpolation_func = self._get_interpolation_func(ensemble_mode)
         self.beam_size = self.models[0].beam_size
-        # after models are loaded we ensured that they agree on max_input_length and max_output_length
+        self.batch_size = self.models[0].batch_size
+        # after models are loaded we ensured that they agree on max_input_length, max_output_length and batch size
         self.max_input_length = self.models[0].max_input_length
         max_output_length = self.models[0].get_max_output_length(self.max_input_length)
         if bucket_source_width > 0:
@@ -588,7 +593,7 @@ class Translator:
             self.buckets_target = data_io.define_buckets(max_output_length, step=bucket_target_width)
         else:
             self.buckets_target = [max_output_length]
-        self.pad_dist = mx.nd.full((self.beam_size, len(self.vocab_target)), val=np.inf, ctx=self.context)
+        self.pad_dist = mx.nd.full((self.batch_size * self.beam_size, len(self.vocab_target)), val=np.inf, ctx=self.context)
         logger.info("Translator (%d model(s) beam_size=%d ensemble_mode=%s "
                     "buckets_source=%s buckets_target=%s)",
                     len(self.models),
@@ -630,45 +635,86 @@ class Translator:
         tokens = list(data_io.get_tokens(sentence))
         return TranslatorInput(id=sentence_id, sentence=sentence.rstrip(), tokens=tokens)
 
-    def translate(self, trans_input: TranslatorInput) -> TranslatorOutput:
+    def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
         """
-        Translates a TranslatorInput and returns a TranslatorOutput
+        Batch-translates a list of TranslatorInputs, returns a list of TranslatorOutputs.
+        Splits oversized sentences to sentence chunks of size less than max_input_length.
 
-        :param trans_input: TranslatorInput as returned by make_input().
-        :return: translation result.
+        :param trans_input: List of TranslatorInputs as returned by make_input().
+        :return: List of translation results.
         """
-        if not trans_input.tokens:
-            return TranslatorOutput(id=trans_input.id,
-                                    translation="",
-                                    tokens=[""],
-                                    attention_matrix=np.asarray([[0]]),
-                                    score=-np.inf)
+        # keep input ids to be able to concatenate translated sentence chunks back
+        chunk_markers = []
+        input_chunks = []
 
-        if len(trans_input.tokens) > self.max_input_length:
-            logger.debug("Input (%d) exceeds max input length (%d). Splitting into chunks of size %d.",
-                         len(trans_input.tokens), self.buckets_source[-1], self.max_input_length)
-            token_chunks = utils.chunks(trans_input.tokens, self.max_input_length)
-            translations = [self.translate_nd(*self._get_inference_input(tokens))
-                            for tokens in token_chunks]
-            translation = self._concat_translations(translations)
-            return self._make_result(trans_input, translation)
-        else:
-            return self._make_result(trans_input, self.translate_nd(*self._get_inference_input(trans_input.tokens)))
+        # prepare chunks, keeping track of empty or splitted inputs
+        for trans_input in trans_inputs:
+            if not trans_input.tokens:  # empty input?
+                chunk_markers.append(C.EMPTY_TRANSLATION_INPUT_ID)
+            elif len(trans_input.tokens) > self.max_input_length:  # oversized input?
+                logger.debug("Input %d has length (%d) that exceeds max input length (%d). Splitting into chunks of size %d.",
+                             trans_input.id, len(trans_input.tokens), self.buckets_source[-1], self.max_input_length)
+                token_chunks = list(utils.chunks(trans_input.tokens, self.max_input_length))
+                input_chunks.extend(token_chunks)
+                chunk_markers.extend([trans_input.id] * len(token_chunks))
+            else:
+                input_chunks.append(trans_input.tokens)
+                chunk_markers.append(trans_input.id)
 
-    def _get_inference_input(self, tokens: List[str]) -> Tuple[mx.nd.NDArray, int]:
+        # start translation in batches
+        translations = []  # type: List[Translation]
+        for i in range(0, len(input_chunks), self.batch_size):
+            logger.debug("Translating batch %d", i)
+            batch = input_chunks[i:i + self.batch_size]
+            # underfilled batch will be filled to a full batch size with copies of the 1st input
+            rest = self.batch_size - len(batch)
+            if rest > 0:
+                logger.debug("Extending the last batch to the full batch size (%d)", self.batch_size)
+                batch = batch + [batch[0]] * rest
+            # translate
+            translated_batch = self.translate_nd(*self._get_inference_input(batch))
+            # truncate to remove duplicated translations
+            if rest > 0:
+                translated_batch = translated_batch[:-rest]
+            translations.extend(translated_batch)
+
+        # concatenate translations corresponding to spans of chunk markers with the same input id
+        results = []  # type: List[TranslatorOutput]
+        start = 0
+        while start < len(chunk_markers):
+            trans_input = trans_inputs[len(results)]
+            if chunk_markers[start] == C.EMPTY_TRANSLATION_INPUT_ID:
+                results.append(TranslatorOutput(id=trans_input.id,
+                                                translation="",
+                                                tokens=[""],
+                                                attention_matrix=np.asarray([[0]]),
+                                                score=-np.inf))
+                start += 1
+            else:
+                _, translations_to_concat = zip(*itertools.takewhile(                                               # while..
+                                                lambda tup: chunk_markers[start] == chunk_markers[start + tup[0]],  # ..it's the same marker (sent id),
+                                                enumerate(translations[start:])))                                   # take pairs of indices & translations.
+                concatenated = translations_to_concat[0] if len(translations_to_concat) == 1 else \
+                               self._concat_translations(translations_to_concat)
+                results.append(self._make_result(trans_input, concatenated))
+                start += len(translations_to_concat)
+        return results
+
+    def _get_inference_input(self, sequences: List[List[str]]) -> Tuple[mx.nd.NDArray, int]:
         """
-        Returns NDArray of source ids (shape=(1, bucket_key)) and corresponding bucket_key.
+        Returns NDArray of source ids (shape=(batch_size, bucket_key)) and corresponding bucket_key.
 
-        :param tokens: List of input tokens.
+        :param tokens: List of lists of input tokens.
         :return NDArray of source ids and bucket key.
         """
-        bucket_key = data_io.get_bucket(len(tokens), self.buckets_source)
+        bucket_key = data_io.get_bucket(max(len(tokens) for tokens in sequences), self.buckets_source)
 
         utils.check_condition(C.PAD_ID == 0, "pad id should be 0")
-        source = mx.nd.zeros((1, bucket_key))
-        ids = data_io.tokens2ids(tokens, self.vocab_source)
-        for i, wid in enumerate(ids):
-            source[0, i] = wid
+        source = mx.nd.zeros((len(sequences), bucket_key))
+        for j, tokens in enumerate(sequences):
+            ids = data_io.tokens2ids(tokens, self.vocab_source)
+            for i, wid in enumerate(ids):
+                source[j, i] = wid
         return source, bucket_key
 
     def _make_result(self,
@@ -709,26 +755,26 @@ class Translator:
 
     def translate_nd(self,
                      source: mx.nd.NDArray,
-                     source_length: int) -> Translation:
+                     source_length: int) -> List[Translation]:
         """
         Translates source of source_length, given a bucket_key.
 
-        :param source: Source ids. Shape: (1, bucket_key).
+        :param source: Source ids. Shape: (batch_size, bucket_key).
         :param source_length: Bucket key.
 
         :return: Sequence of translated ids, attention matrix, length-normalized negative log probability.
         """
         return self._get_best_from_beam(*self._beam_search(source, source_length))
 
-    def _encode(self, source: mx.nd.NDArray, source_length: int) -> List[ModelState]:
+    def _encode(self, sources: mx.nd.NDArray, source_length: int) -> List[ModelState]:
         """
         Returns a ModelState for each model representing the state of the model after encoding the source.
 
-        :param source: Source ids. Shape: (1, bucket_key).
+        :param sources: Source ids. Shape: (batch_size, bucket_key).
         :param source_length: Bucket key.
         :return: List of ModelStates.
         """
-        return [ModelState(states=m.run_encoder(source, source_length)) for m in self.models]
+        return [ModelState(model.run_encoder(sources, source_length)) for model in self.models]
 
     def _decode_step(self,
                      sequences: mx.nd.NDArray,
@@ -739,7 +785,7 @@ class Translator:
         """
         Returns decoder predictions (combined from all models), attention scores, and updated states.
 
-        :param sequences: Sequences of current hypotheses. Shape: (beam_size, max_output_length).
+        :param sequences: Sequences of current hypotheses. Shape: (batch_size * beam_size, max_output_length).
         :param t: Beam search iteration.
         :param source_length: Length of the input sequence.
         :param max_output_length: Maximum output length.
@@ -787,10 +833,10 @@ class Translator:
                      source: mx.nd.NDArray,
                      source_length: int) -> Tuple[mx.nd.NDArray, mx.nd.NDArray, mx.nd.NDArray, mx.nd.NDArray]:
         """
-        Translates a single sentence using beam search.
+        Translates multiple sentences using beam search.
 
-        :param source: Source ids. Shape: (1, bucket_key).
-        :param source_length: Source length.
+        :param source: Source ids. Shape: (batch_size, bucket_key).
+        :param source_length: Max source length.
         :return List of lists of word ids, list of attentions, array of accumulated length-normalized
                 negative log-probs.
         """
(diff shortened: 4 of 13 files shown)


sockeye/arguments.py
```diff
@@ -659,16 +659,6 @@ def add_training_args(params):
                               default=0,
                               help="Number of warmup steps. If set to x, linearly increases learning rate from 10%% "
                                    "to 100%% of the initial learning rate. Default: %(default)s.")
-    train_params.add_argument('--learning-rate-decay-param-reset',
```
fhieber: please rebase to latest master or put these back in :)

sockeye/arguments.py
```diff
@@ -741,6 +731,14 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--batch-size',
+                               type=int_greater_or_equal(1),
+                               default=1,
+                               help='Size of the batch. Default: %(default)s.')
+    decode_params.add_argument('--portion-size',
```
fhieber: can we rename this to --chunk-size (and similarly in the code below: portion -> chunk)? I think this is a more common term.

sockeye/arguments.py
```diff
@@ -741,6 +731,14 @@ def add_inference_args(params):
...
+                               type=int_greater_or_equal(1),
+                               default=10000,
+                               help='Size of the batch. Default: %(default)s.')
```
fhieber: Please update help string.

sockeye/arguments.py
```diff
@@ -741,6 +731,14 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--batch-size',
+                               type=int_greater_or_equal(1),
+                               default=1,
+                               help='Size of the batch. Default: %(default)s.')
```
fhieber: Could use a more descriptive help message, e.g. Batch size during decoding. Determines how many sentences are translated simultaneously.

sockeye/inference.py
```diff
@@ -70,6 +72,7 @@ def __init__(self,
         self.beam_size = beam_size
         self.softmax_temperature = softmax_temperature
         self.encoder_batch_size = 1
```
fhieber: Why is this still 1? Why not encode in batches as well?

sockeye/inference.py
```diff
@@ -233,7 +236,7 @@ def run_encoder(self,
         Returns encoder representation of the source, source_length, initial hidden state of decoder RNN,
         and initial decoder states tiled to beam size.
 
-        :param source: Integer-coded input tokens.
+        :param source: Integer-coded input tokens. Shape (1, source length)
```
fhieber: missing period

sockeye/inference.py
```diff
@@ -340,6 +345,9 @@ def load_models(context: mx.context.Context,
                           "Source vocabulary ids do not match")
     utils.check_condition(all(set(vocab.items()) == set(target_vocabs[0].items()) for vocab in target_vocabs),
                           "Target vocabulary ids do not match")
+    utils.check_condition(all(m.batch_size == models[0].batch_size for m in models),
```
fhieber: why this check? You load all models a couple of lines above with the same batch_size so how can they differ?

sockeye/inference.py
```diff
@@ -452,31 +460,40 @@ class ModelState:
     A ModelState encapsulates information about the decoder states of an InferenceModel.
     """
 
-    def __init__(self, states: List[mx.nd.NDArray]) -> None:
+    def __init__(self, states: List[mx.nd.NDArray]=None) -> None:
```
fhieber: Why?

sockeye/inference.py
```diff
@@ -452,31 +460,40 @@ class ModelState:
...
         self.states = states
 
-    def sort_state(self, best_hyp_indices: mx.nd.NDArray):
+    def sort_state(self, best_hyp_indices: mx.nd.NDArray) -> None:
```
fhieber: This function does not return None

sockeye/inference.py
```diff
@@ -452,31 +460,40 @@ class ModelState:
...
         """
         Sorts states according to k-best order from last step in beam search.
         """
         self.states = [mx.nd.take(ds, best_hyp_indices) for ds in self.states]
 
+    def append_state(self, states: List[mx.nd.NDArray]) -> None:
```
fhieber: this one neither

sockeye/inference.py
```diff
@@ -452,31 +460,40 @@ class ModelState:
...
+            utils.check_condition(len(states) == len(self.states), "State lists have different lengths")
+            self.states = [mx.nd.concat(old_s, new_s, dim=0) for old_s, new_s in zip(self.states, states)]
+        else:
+            self.states = states
 
 class LengthPenalty:
     """
     Calculates the length penalty as:
-    (beta + len(Y))**alpha / (beta + 1)**alpha
+    (gamma + len(Y))**alpha / (gamma + 1)**alpha
```
fhieber: if you rename this parameter, you should also rename the CLI argument, effectively making this a breaking change so the version should be bumped. Do you have a good reason to change it ?

sockeye/inference.py
```diff
@@ -630,6 +648,63 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
         tokens = list(data_io.get_tokens(sentence))
         return TranslatorInput(id=sentence_id, sentence=sentence.rstrip(), tokens=tokens)
 
+    def batch_translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
+        """
+        Batch-translates a list of TranslatorInput, returns a list of TranslatorOutputs
```
fhieber: missing period

sockeye/inference.py
```diff
@@ -630,6 +648,63 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+        """
+        # TODO handle empty inputs, preferably outside inference.py?
+        # TODO avoid accepting and creating lists, use generators?
+
+        # keep input ids to be able to concatenate translated chunks back
+        chunk_markers = []
+        input_chunks = []
+
+        # prepare chunks, keeping track of splitted inputs
+        for tinput in trans_inputs:
```
fhieber: maybe trans_input instead of tinput?

sockeye/inference.py
```diff
@@ -630,6 +648,63 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+        translations = []  # type: List[Translation]
+        for i in range(0, len(input_chunks), self.batch_size):
+            logger.debug("Translating batch %d", i)
+            batch = input_chunks[i:i + self.batch_size]
+            # underfilled batch will be filled to a full batch size with empty inputs
+            rest = self.batch_size - len(batch)
+            if rest > 0:
+                logger.debug("Replicating %d random sentences from the last batch to size it to the full batch size (%d)",
+                             rest, self.batch_size)
+                random_indices = np.random.randint(len(batch), size=rest)
```
fhieber: why random? couldn't you pick the same all the time?

sockeye/inference.py
```diff
@@ -645,30 +720,31 @@ def translate(self, trans_input: TranslatorInput) -> TranslatorOutput:
...
         """
-        Returns NDArray of source ids (shape=(1, bucket_key)) and corresponding bucket_key.
+        Returns NDArray of source ids (shape=(num_sent, bucket_key)) and corresponding bucket_key.
 
-        :param tokens: List of input tokens.
+        :param tokens: List of lists of input tokens.
         :return NDArray of source ids and bucket key.
         """
-        bucket_key = data_io.get_bucket(len(tokens), self.buckets_source)
+        bucket_key = data_io.get_bucket(max(len(t) for t in tokens), self.buckets_source)
```
fhieber: t is a bit confusing, one may think this is a single token and you want its length, whereas this is actually the length of the token sequence.
Maybe rename the function argument to sequences or something.

sockeye/inference.py
```diff
@@ -709,26 +785,30 @@ def _concat_translations(self, translations: List[Translation]) -> Translation:
 
     def translate_nd(self,
                      source: mx.nd.NDArray,
-                     source_length: int) -> Translation:
+                     source_length: int) -> List[Translation]:
         """
         Translates source of source_length, given a bucket_key.
 
-        :param source: Source ids. Shape: (1, bucket_key).
+        :param source: Source ids. Shape: (batch_size, bucket_key) or (1, bucket_key)
```
fhieber: why or? in the baseline case batch_size is just 1.

sockeye/inference.py
```diff
@@ -709,26 +785,30 @@ def _concat_translations(self, translations: List[Translation]) -> Translation:
...
         """
         return self._get_best_from_beam(*self._beam_search(source, source_length))
 
-    def _encode(self, source: mx.nd.NDArray, source_length: int) -> List[ModelState]:
+    def _encode(self, sources: mx.nd.NDArray, source_length: int) -> List[ModelState]:
         """
         Returns a ModelState for each model representing the state of the model after encoding the source.
 
-        :param source: Source ids. Shape: (1, bucket_key).
+        :param sources: Source ids. Shape: (num_sent, bucket_key).
```
fhieber: lets be consistent and name this `batch_size` and not `num_sent`.

sockeye/inference.py
```diff
@@ -709,26 +785,30 @@ def _concat_translations(self, translations: List[Translation]) -> Translation:
...
         :param source_length: Bucket key.
         :return: List of ModelStates.
         """
-        return [ModelState(states=m.run_encoder(source, source_length)) for m in self.models]
+        states = [ModelState()] * len(self.models)
+        for model, state in zip(self.models, states):
+            for src in sources:
+                state.append_state(model.run_encoder(mx.nd.expand_dims(src, axis=0), source_length))
```
fhieber: I don't like this iterative invocation of the encoder module. 
This really should be batched as well.

sockeye/inference.py
```diff
@@ -739,7 +819,7 @@ def _decode_step(self,
         """
         Returns decoder predictions (combined from all models), attention scores, and updated states.
 
-        :param sequences: Sequences of current hypotheses. Shape: (beam_size, max_output_length).
+        :param sequences: Sequences of current hypotheses. Shape: (num_sent * beam_size, max_output_length).
```
fhieber: consistency: batch_size * beam_size

sockeye/inference.py
```diff
@@ -630,6 +648,63 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
         tokens = list(data_io.get_tokens(sentence))
         return TranslatorInput(id=sentence_id, sentence=sentence.rstrip(), tokens=tokens)
 
+    def batch_translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
```
fhieber: would it make the batch_size=1 case less efficient if you collapse `batch_translate()` and `translate`?

sockeye/inference.py
```diff
@@ -787,10 +867,10 @@ def _beam_search(self,
                      source: mx.nd.NDArray,
                      source_length: int) -> Tuple[mx.nd.NDArray, mx.nd.NDArray, mx.nd.NDArray, mx.nd.NDArray]:
         """
-        Translates a single sentence using beam search.
+        Translates multiple sentences using beam search.
 
-        :param source: Source ids. Shape: (1, bucket_key).
-        :param source_length: Source length.
+        :param source: Source ids. Shape: (num_sent, bucket_key). Num_sent can be from 1 up to self.batch_size
```
fhieber: why is it a range? isnt it always exactly batch_size (which can be 1)?

Also missing period.

sockeye/inference.py
```diff
@@ -802,42 +882,53 @@ def _beam_search(self,
         # Maximum output length
         max_output_length = self.models[0].get_max_output_length(source_length)
 
-        # sequences: (beam_size, output_length), pre-filled with <s> symbols on index 0
-        sequences = mx.nd.array(np.full((self.beam_size, max_output_length), C.PAD_ID), dtype='int32', ctx=self.context)
+        num_sent = source.shape[0]
```
fhieber: consistency: batch_size

sockeye/inference.py
```diff
@@ -802,42 +882,53 @@ def _beam_search(self,
...
-        best_hyp_indices = mx.nd.zeros((self.beam_size,), ctx=self.context)
+        best_hyp_indices = mx.nd.zeros((num_sent * self.beam_size,), ctx=self.context, dtype='int32')
         # best_word_indices: column indices of smallest scores (ascending).
-        best_word_indices = mx.nd.zeros((self.beam_size,), ctx=self.context, dtype='int32')
+        best_word_indices = mx.nd.zeros((num_sent * self.beam_size,), ctx=self.context, dtype='int32')
         # scores_accumulated: chosen smallest scores in scores (ascending).
-        scores_accumulated = mx.nd.zeros((self.beam_size, 1), ctx=self.context)
+        scores_accumulated = mx.nd.zeros((num_sent * self.beam_size, 1), ctx=self.context)
+
+        best_hyp_indices_np = np.empty((num_sent * self.beam_size,))
```
fhieber: cosmetics: could move all of these to their ndarray equivalents

sockeye/arguments.py
tdomhan: why did these disappear?

sockeye/inference.py
```diff
@@ -850,12 +941,21 @@ def _beam_search(self,
...
-            scores_accumulated[:] = np.expand_dims(scores_accumulated_np, axis=1)
+            scores = scores.asnumpy()  # convert to numpy once to minimize cross-device copying
+            for sent in range(num_sent):
+                rows = slice(sent * self.beam_size, (sent + 1) * self.beam_size)
+                if mx.nd.sum(finished[rows]).asscalar() == self.beam_size:  # is this sentence finished?
+                    continue
+                sliced_scores = scores if t == 1 and num_sent == 1 else scores[rows]
+                (best_hyp_indices_np[rows], best_word_indices_np[rows]), \
+                        scores_accumulated_np[rows] = utils.smallest_k(sliced_scores, self.beam_size, t == 1)
+                best_hyp_indices_np[rows] += rows.start
```
fhieber: could you add a comment here that you offset the hyp indices ?

sockeye/inference.py
```diff
@@ -880,25 +980,33 @@ def _beam_search(self,
...
-        # attention_matrix: (target_seq_len, source_seq_len)
-        attention_matrix = np.stack(attention_lists[best].asnumpy()[:length, :], axis=0)
-        score = accumulated_scores[best].asscalar()
-        return Translation(sequence, attention_matrix, score)
+        num_sent = sequences.shape[0] // self.beam_size
+        result = []
+        for sent in range(num_sent):
+            idx = sent * self.beam_size + best
+            length = int(lengths[idx].asscalar())
+            sequence = sequences[idx].asnumpy().tolist()
```
fhieber: why do you not cut to length here anymore?

sockeye/translate.py
```diff
@@ -67,40 +68,60 @@ def main():
...
                                                                                  args.max_output_length_num_stds))
-        read_and_translate(translator, output_handler, args.input)
+
+        logger.info("Using batches of size %d", args.batch_size)
+        read_and_translate(translator, output_handler, args.portion_size, args.input)
 
 
 def read_and_translate(translator: sockeye.inference.Translator, output_handler: sockeye.output_handler.OutputHandler,
-                       source: Optional[str] = None) -> None:
+                       portion_size: int, source: Optional[str] = None) -> None:
```
fhieber: as mentioned above, I would prefer chunk_size

sockeye/translate.py
```diff
@@ -67,40 +68,60 @@ def main():
...
     """
     Reads from either a file or stdin and translates each line, calling the output_handler with the result.
 
     :param output_handler: Handler that will write output to a stream.
     :param translator: Translator that will translate each line of input.
+    :param portion_size: The size of the portion to read at a time from the input
```
fhieber: missing period.

sockeye/translate.py
```diff
@@ -67,40 +68,60 @@ def main():
...
+        lines, portion_time = translate(output_handler, portion, translator)
+        total_lines += lines
+        total_time += portion_time
 
-    if i != 0:
-        logger.info("Processed %d lines. Total time: %.4f sec/sent: %.4f sent/sec: %.4f", i, total_time,
-                    total_time / i, i / total_time)
+    if total_lines != 0:
+        logger.info("Processed %d lines in %d batches. Total time: %.4f, sec/sent: %.4f, sent/sec: %.4f",
+                    total_lines, -(-lines // translator.batch_size), total_time,
```
fhieber: why the double minus?

sockeye/translate.py
```diff
@@ -110,16 +131,14 @@ def translate_lines(output_handler: sockeye.output_handler.OutputHandler, source
...
-        logger.debug(" IN: %s", trans_input)
-        trans_output = translator.translate(trans_input)
-        trans_wall_time = time.time() - tic
-        total_time += trans_wall_time
-        logger.debug("OUT: %s", trans_output)
-        logger.debug("OUT: time=%.2f", trans_wall_time)
-        output_handler.handle(trans_input, trans_output, trans_wall_time)
+    tic = time.time()
+    trans_inputs = []
+    for i, line in enumerate(source_data, 1):  # TODO not using comprehesions to keep the last i; fix this
```
fhieber: why not use a list comprehension? you can just get i from the length of trans_inputs, which is a constant operation.

artemsok: Good point, was an artefact of having inputs previously as generators, not lists

sockeye/translate.py
artemsok: A shorter equivalent of math.ceil(lines // translator,batch_size), will change to math.ceil.

sockeye/translate.py
artemsok: Ok, but will conflate a bit with the other meaning of chunk inside inference.py - when splitting the over-sized sentences

sockeye/inference.py
artemsok: Good catch! Was lost in refactoring.

sockeye/inference.py
artemsok: Ok, will rewrite this

sockeye/inference.py
artemsok: True, will take the 0th.

sockeye/inference.py
artemsok: Will do, but this is equally confusing to call in beta, while the beta in the paper is something else completely.

sockeye/inference.py
artemsok: Yeah, was copied from __init__() which had -> None for some reason

sockeye/inference.py
artemsok: Why the default argument? The constructor is used (with the default argument) to pre-fill an array later.

sockeye/translate.py
fhieber: Why not use generators? (just curious what this entailed)

sockeye/inference.py
fhieber: the reason we annotate the return type of constructors as None is to make mypy happy :)
For other functions its not needed

sockeye/translate.py
fhieber: these could be refactored to sentence_chunk or something to make it explicit.

sockeye/inference.py
fhieber: I see. Maybe this won't be needed when encoding is batched as well.

sockeye/translate.py
artemsok: We need one pass over inputs to see if some sentences need splitting before we send them for translation in batches. At this point it's possible to do it with generators, at the cost of slightly more complicated code, but we could have it a separate PR. 

The real mess will start if we decide to fill batches optimally. Then, some of the sentence chunks might end up in different batches. The code will have to keep track of them for later gluing, not being able to output anything over potentially a very long time, storing intermediate results in memory until some of them can be output. In the worst case this may require storing everything until the last batch completes.

fhieber: Ok, this makes a lot of sense :)
Lets stay with lists then.

sockeye/inference.py
```diff
@@ -630,6 +635,62 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+            else:
+                input_chunks.append(trans_input.tokens)
+                chunk_markers.append(trans_input.id)
+
+        # start translation in batches
+        translations = []  # type: List[Translation]
+        for i in range(0, len(input_chunks), self.batch_size):
+            logger.debug("Translating batch %d", i)
+            batch = input_chunks[i:i + self.batch_size]
+            # underfilled batch will be filled to a full batch size with empty inputs
```
fhieber: empty or just the first element of the batch?

sockeye/translate.py
```diff
@@ -67,60 +69,75 @@ def main():
...
-        logger.debug("OUT: time=%.2f", trans_wall_time)
-        output_handler.handle(trans_input, trans_output, trans_wall_time)
-    return i, total_time
+    tic = time.time()
+    trans_inputs = [translator.make_input(i, line) for i, line in enumerate(source_data, 1)]
+    trans_outputs = translator.batch_translate(trans_inputs)
+    total_time = time.time() - tic
+    for trans_input, trans_output in zip(trans_inputs, trans_outputs):
+        output_handler.handle(trans_input, trans_output)
+    return len(trans_outputs), total_time
```
tdomhan: why return the len(outputs)?

sockeye/translate.py
```diff
@@ -67,60 +69,75 @@ def main():
...
 
     source_data = sys.stdin if source is None else sockeye.data_io.smart_open(source)
 
     logger.info("Translating...")
 
-    i, total_time = translate_lines(output_handler, source_data, translator)
+    total_time, total_lines = 0.0, 0
+    for chunk in grouper(source_data, chunk_size):
+        lines, chunk_time = translate(output_handler, chunk, translator)
+        total_lines += lines
```
tdomhan: couldn't you just do `total_lines += len(chunk)`? And otherwise I'd rename to `num_lines`.

sockeye/inference.py
```diff
@@ -630,6 +635,62 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
         tokens = list(data_io.get_tokens(sentence))
         return TranslatorInput(id=sentence_id, sentence=sentence.rstrip(), tokens=tokens)
 
+    def batch_translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
```
fhieber: since you are not calling `translate()` anymore, why not rename batch_translate to translate?

sockeye/inference.py
```diff
@@ -802,42 +864,52 @@ def _beam_search(self,
         # Maximum output length
         max_output_length = self.models[0].get_max_output_length(source_length)
 
-        # sequences: (beam_size, output_length), pre-filled with <s> symbols on index 0
-        sequences = mx.nd.array(np.full((self.beam_size, max_output_length), C.PAD_ID), dtype='int32', ctx=self.context)
+        batch_size = source.shape[0]
```
fhieber: you can use self.batch_size , no ?

sockeye/inference.py
```diff
@@ -880,25 +961,33 @@ def _beam_search(self,
...
-        :param sequences: Array of word ids. Shape: (beam_size, bucket_key).
+        :param sequences: Array of word ids. Shape: (batch_size * beam_size, bucket_key).
         :param attention_lists: Array of attentions over source words. Shape: (length, bucket_key).
         :param accumulated_scores: Array of length-normalized negative log-probs.
-        :return: Top sequence, top attention matrix, top accumulated score (length-normalized negative log-probs)
-                 and length.
+        :return: Top sequence, top attention matrix, top accumulated score (length-normalized
+                 negative log-probs) and length.
         """
+        #utils.check_condition(sequences.shape[0] == attention_lists.shape[0] \
```
fhieber: can you move that check into the loop below?

test/common.py
```diff
@@ -178,3 +178,69 @@ def run_train_translate(train_params: str,
             sockeye.evaluate.main()
 
         return perplexity, bleu
+
+
+def run_train_translate_twice(train_params: str,
+                        translate_params1: str,
```
fhieber: indentation is a bit off :)

test/integration/test_seq_copy_int.py
```diff
@@ -97,3 +97,27 @@ def test_seq_copy(train_params, translate_params):
...
+        train_source_path = os.path.join(work_dir, "train.src")
+        train_target_path = os.path.join(work_dir, "train.tgt")
+        dev_source_path = os.path.join(work_dir, "dev.src")
+        dev_target_path = os.path.join(work_dir, "dev.tgt")
+        generate_digits_file(train_source_path, train_target_path, _TRAIN_LINE_COUNT, _LINE_MAX_LENGTH)
+        generate_digits_file(dev_source_path, dev_target_path, _DEV_LINE_COUNT, _LINE_MAX_LENGTH)
+        # Test model configuration
+        translate_params_batch = translate_params + " --batch-size 2"
+        run_train_translate_twice(train_params,
+                            translate_params,
```
tdomhan: the indentation is off here.

test/common.py
```diff
@@ -178,3 +178,69 @@ def run_train_translate(train_params: str,
...
+                        dev_source_path: str,
+                        dev_target_path: str,
+                        max_seq_len: int = 10,
+                        work_dir: Optional[str] = None) -> Tuple[float, float]:
+    """
+    Train a model and translate a dev set twice.  
+    Ensure the translation under different translation parameters are the same
+
+    :param train_params: Command line args for model training.
+    :param translate_params: Command line args for translation.
```
tdomhan: the doc string doesn't match the parameters above.

test/common.py
tdomhan: the indentation is not aligned :P

sockeye/inference.py
```diff
@@ -630,6 +635,62 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+
+        # concatenate translations corresponding to spans of chunk markers with the same input id
+        results = []  # type: List[TranslatorOutput]
+        start = 0
+        while start < len(chunk_markers):
+            _, translations_to_concat = zip(*itertools.takewhile(                                               # while..
+                                            lambda tup: chunk_markers[start] == chunk_markers[start + tup[0]],  # ..it's the same marker (sent id),
+                                            enumerate(translations[start:])))                                   # take pairs of indices & translations.
+            results.append(self._make_result(trans_inputs[len(results)],
+                           self._concat_translations(translations_to_concat)))
```
tdomhan: we should avoid the additional overhead of _concat_translations when there is only one translation:
```
results.append(self._make_result(trans_inputs[len(results)],
                          translations_to_concat[0] if len(translations_to_concat) == 1 else self._concat_translations(translations_to_concat)))
```

sockeye/inference.py
artemsok: Good catch, will do

sockeye/inference.py
artemsok: Right, the 1st

sockeye/inference.py
```diff
@@ -630,6 +635,62 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+    def batch_translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
+        """
+        Batch-translates a list of TranslatorInput, returns a list of TranslatorOutputs.
+        Splits oversized sentneces to sentence chunks of size less than max_input_length.
+
+        :param trans_input: List of TranslatorInputs as returned by make_input().
+        :return: List of translation results.
+        """
+        # TODO handle empty inputs, preferably outside inference.py?
+        # TODO avoid accepting and creating lists, use generators?
```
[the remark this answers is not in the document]
artemsok: Ok

sockeye/inference.py
```diff
@@ -630,6 +635,62 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
 
+    def batch_translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
+        """
+        Batch-translates a list of TranslatorInput, returns a list of TranslatorOutputs.
+        Splits oversized sentneces to sentence chunks of size less than max_input_length.
+
+        :param trans_input: List of TranslatorInputs as returned by make_input().
+        :return: List of translation results.
+        """
+        # TODO handle empty inputs, preferably outside inference.py?
```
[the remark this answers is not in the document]
artemsok: Ok, will submit an update for this

sockeye/inference.py
artemsok: Fixed

sockeye/translate.py
artemsok: True. I think we can get rid of accumulating total_lines, since it will be always equal to len(source_data)

fhieber: That  I am not sure about since source_data is a generator, right? Len would the consume it to calculate length which is bad :)

tdomhan: it's actually a list (see line 99)

fhieber: True, nevermind my comment then.

sockeye/translate.py
```diff
@@ -67,60 +69,75 @@ def main():
...
+def translate(output_handler: sockeye.output_handler.OutputHandler, source_data: Iterable[str],
+                    translator: sockeye.inference.Translator) -> float:
     """
-    Translates each line from source_data, calling output handler for each result.
+    Translates each line from source_data, calling output handler after translating a batch.
 
     :param output_handler: A handler that will be called once with the output of each translation.
     :param source_data: A enumerable list of source sentences that will be translated.
     :param translator: The translator that will be used for each line of input.
     :return: The number of lines translated, and the total time taken.
```
fhieber: Outdated docstring for return.

sockeye/inference.py
```diff
@@ -630,45 +635,86 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
         tokens = list(data_io.get_tokens(sentence))
         return TranslatorInput(id=sentence_id, sentence=sentence.rstrip(), tokens=tokens)
 
-    def translate(self, trans_input: TranslatorInput) -> TranslatorOutput:
+    def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
         """
-        Translates a TranslatorInput and returns a TranslatorOutput
+        Batch-translates a list of TranslatorInput, returns a list of TranslatorOutputs.
+        Splits oversized sentneces to sentence chunks of size less than max_input_length.
```
fhieber: typo: sentences.

artemsok: Thank you! Updating..

sockeye/inference.py
```diff
@@ -630,45 +635,86 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+            translated_batch = self.translate_nd(*self._get_inference_input(batch))
+            # truncate to remove duplicated translations
+            if rest > 0:
+                translated_batch = translated_batch[:-rest]
+            translations.extend(translated_batch)
+
+        # concatenate translations corresponding to spans of chunk markers with the same input id
+        results = []  # type: List[TranslatorOutput]
+        start = 0
+        while start < len(chunk_markers):
```
tdomhan: Just a suggestion but I think this could be simplified with `groupby`:
```
for translation_id, translations_and_markers in groupby(zip(translations, chunk_markers), key=[1]):
    translations_to_concat = [translation for translation, marker in translations_and_markers]
    if translation_id == C.EMPTY_TRANSLATION_INPUT_ID:
        for translation in translations_to_concat:
            results.append(TranslatorOutput(id=translation.id,
                                                     translation="",
                                                     tokens=[""],
                                                     attention_matrix=np.asarray([[0]]),
                                                     score=-np.inf))
    else:
        concatenated = translations_to_concat[0] if len(translations_to_concat) == 1 else \
                               self._concat_translations(translations_to_concat)
        results.append(self._make_result(trans_input, concatenated))
    
```

sockeye/inference.py
```diff
@@ -802,42 +848,51 @@ def _beam_search(self,
...
+        best_hyp_indices_np = np.empty((self.batch_size * self.beam_size,), dtype='int32')
+        best_word_indices_np = np.empty((self.batch_size * self.beam_size,), dtype='int32')
+        scores_accumulated_np = np.empty((self.batch_size * self.beam_size,))
 
         # reset all padding distribution cells to np.inf
         self.pad_dist[:] = np.inf
 
-        # (0) encode source sentence
+        # (0) encode source sentence, returns a list
+        # [encoder representation of the source, source_length, initial hidden state of decoder RNN, initial decoder states] tiled to beam size
```
tdomhan: This comment is specifically for the RNN, but depending on the model there might be different hidden states. In general the list will contain whatever hidden states the decoder provides.

sockeye/inference.py
```diff
@@ -630,45 +635,86 @@ def make_input(sentence_id: int, sentence: str) -> TranslatorInput:
...
+        :return: List of translation results.
         """
-        if not trans_input.tokens:
-            return TranslatorOutput(id=trans_input.id,
-                                    translation="",
-                                    tokens=[""],
-                                    attention_matrix=np.asarray([[0]]),
-                                    score=-np.inf)
+        # keep input ids to be able to concatenate translated sentence chunks back
+        chunk_markers = []
```
tdomhan: you know my opinion about sorting here. Let's leave it out for now, but reconsider later.

sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.8.4'
+__version__ = '1.8.5'
```
tdomhan: Let's bump the major version and go to '1.9.0' as the Translator interface changes.

sockeye/inference.py
```diff
@@ -802,42 +848,51 @@ def _beam_search(self,
...
+        # sequences: (batch_size * beam_size, output_length), pre-filled with <s> symbols on index 0
+        sequences = mx.nd.full((self.batch_size * self.beam_size, max_output_length), val=C.PAD_ID, ctx=self.context, dtype='int32')
         sequences[:, 0] = self.start_id
 
-        lengths = mx.nd.ones((self.beam_size, 1), ctx=self.context)
-        finished = mx.nd.zeros((self.beam_size,), dtype='int32', ctx=self.context)
+        lengths = mx.nd.ones((self.batch_size * self.beam_size, 1), ctx=self.context)
+        finished = mx.nd.zeros((self.batch_size * self.beam_size,), ctx=self.context, dtype='int32')
 
         # attentions: (beam_size, output_length, encoded_source_length)
```
tdomhan: the comment about the dimensionality is not up to date.

sockeye/inference.py
```diff
@@ -880,25 +944,31 @@ def _beam_search(self,
...
                             attention_lists: mx.nd.NDArray,
                             accumulated_scores: mx.nd.NDArray,
-                            lengths: mx.nd.NDArray) -> Translation:
+                            lengths: mx.nd.NDArray) -> List[Translation]:
         """
         Return the best (aka top) entry from the n-best list.
 
-        :param sequences: Array of word ids. Shape: (beam_size, bucket_key).
+        :param sequences: Array of word ids. Shape: (batch_size * beam_size, bucket_key).
         :param attention_lists: Array of attentions over source words. Shape: (length, bucket_key).
```
tdomhan: can you update the shape comment? Should be `(batch_size * self.beam_size, max_output_length, encoded_source_length)` no?

artemsok: fixed

sockeye/inference.py
artemsok: Removed the comment, copied it indeed from the RNN decoder for my notes

sockeye/inference.py
artemsok: fixed

sockeye/inference.py
artemsok: Unfortunately doesn't work since chunks and translations have different sizes and are not aligned (there are no entries in translation for empty inputs, since we are not transalting them). But I like the groupby approach, - -will have to learn itertools better, maybe it can be fixed.

tdomhan: ah yeah that's right, makes sense.

sockeye/checkpoint_decoder.py
```diff
@@ -140,7 +141,7 @@ def decode_and_evaluate(self,
             for i, input_sentence in enumerate(self.input_sentences):
                 tic = time.time()
                 trans_input = translator.make_input(i, input_sentence)
-                trans_output = translator.translate(trans_input)
+                trans_output = translator.translate([trans_input])[0]
```
tdomhan: mhh, now that the batch size is 16, we should also feed in sentence in chunks here, no? Why not just feed in the entire input_sentences?

fhieber: Agreed.
On a related note, we could also cache the results of make_input but that could be done in another PR.

artemsok: This will break the percentile calculation, no? I think this was why I left batch size 1 initially for checkpoints. As a workaround, we could report some surrogate percentiles, approximated from batch translation times and batch_size.

sockeye/checkpoint_decoder.py
```diff
@@ -133,20 +132,19 @@ def decode_and_evaluate(self,
...
+        translations = []
         with data_io.smart_open(output_name, 'w') as output:
             handler = sockeye.output_handler.StringOutputHandler(output)
-            translations = []
-            for i, input_sentence in enumerate(self.input_sentences):
-                tic = time.time()
-                trans_input = translator.make_input(i, input_sentence)
-                trans_output = translator.translate(trans_input)
+            tic = time.time()
+            trans_inputs = [translator.make_input(i, line) for i, line in enumerate(source_data)]
```
tdomhan: I think there's a copy/paste error here. You should use `self.input_sentences ` no? That's why mypy is unhappy.

artemsok: Yes, should be fixed now

tdomhan: have you run a quick seqcopy model with bleu monitoring to double check (as this is something not covered by the unit/integration tests)?

PR #173, opened 2017-10-23
Update to MXNet 0.12

fhieber:
WIP.
Checks will likely fail until MXNet 0.12.0 is released.
Created a new major version to update some flags. Please let me know if you want to integrate any backwards incompatible changes into this.


sockeye/loss.py
```diff
@@ -168,87 +163,3 @@ def update(self, labels, preds):
...
-
-        cross_entropy = mx.sym.MakeLoss(cross_entropy, name=C.SMOOTHED_CROSS_ENTROPY)
-        probs = mx.sym.BlockGrad(probs, name=C.SOFTMAX_NAME)
-        return [cross_entropy, probs]
-
-    def create_metric(self) -> "SmoothedCrossEntropyMetric":
-        return SmoothedCrossEntropyMetric(self.loss_config)
-
-
-class SmoothedCrossEntropyMetric(EvalMetric):
```
mjdenkowski: It's great that we can merge smoothed and non-smoothed loss!  We should make sure that `CrossEntropyMetric` can also apply smoothing to match `CrossEntropyLoss`.

fhieber: I thought about that, but its also actually not too bad to have the same loss values for smoothed and non-smoothed gradients for easy comparison. I will look into it.

mjdenkowski: The main place we're using the loss-replicating metrics is for optimizers like Eve that want the exact loss from training.  The idea is that `Loss` classes can return corresponding `Metric` versions that use the same config and return the same results.  For checkpoints, we're still using the vanilla MXNet metrics.

fhieber: Ah you are right, I will put that back in then!

fhieber: I added back the smoothed cross-entropy for the Loss metric. However, this is super slow as we are not caching the label distributions as before in the symbol graph. This probably removes all the benefit we have from the fusioning of the smoothed CE grad in MXNet 0.12.
I will think about a different way for this.

mjdenkowski: Do we have a way to return the loss?  The custom metrics were originally only used for the Eve optimizer, which just needs the loss.

fhieber: The problem is the smoothed loss in general. Explicitly creating the soft labels / label distribution is super expensive. As it is right now, we reduce speed by 60%.

tdomhan:
I remembered what other breaking change I was thinking about adding: Adding another digit or two to the parameter file names to reduce the chance of overflowing at some point. What do you think?

fhieber:
fine with me :) will add that

philschulz:
Seem like mxnet 0.12 has been released (at least it's on pypi). Assuming that I have mxnet 0.12 installed, is it safe to rebase my code on this branch?

fhieber:
I'd say yes you can. I will rebase this PR on latest master.

philschulz:
Great, thanks a lot! I'll rebase tomorrow (it's 9pm in Australia right now).

README.md
```diff
@@ -50,19 +50,13 @@ remaining instructions to work you will need to use `python3` instead of `python
...
+Depending on your version of CUDA, you can do this by running the following:
 ```bash
-> wget https://raw.githubusercontent.com/awslabs/sockeye/master/requirements.gpu-cu75.txt
-> pip install sockeye --no-deps -r requirements.gpu-cu75.txt
-> rm requirements.gpu-cu75.txt
+> wget https://raw.githubusercontent.com/awslabs/sockeye/master/requirements.gpu-cu${CUDA_VERSION}.txt
+> pip install sockeye --no-deps -r requirements.gpu-cu${CUDA_VERSION}.txt
+> rm requirements.gpu-cu${CUDA_VERSION}.txt
 ```
+where `${CUDA_VERSION}` can be `75`, `80`, or `90`.
```
tdomhan: maybe mention the fact that 7.5 = 75, 8.0=80 and 9.0=90

sockeye/layers.py
```diff
@@ -108,6 +112,64 @@ def __call__(self) -> mx.sym.Symbol:
...
+                 num_hidden: int,
+                 num_embed: int,
+                 vocab_size: int,
+                 weight_tying: bool,
+                 embed_weight: Optional[mx.sym.Symbol],
+                 weight_normalization: bool,
+                 prefix: str = '') -> None:
+        self.vocab_size = vocab_size
+
+        self.w = mx.sym.Variable("%scls_weight" % prefix, shape=(vocab_size, num_hidden))
```
tdomhan: how about moving these into the else branch of `weight_tying` to only create it if weight tying is turned off?

fhieber: good call, will change!

README.md
fhieber: done!

sockeye/loss.py
fhieber: I will leave as is right now as this is correct. If we see that eve + label smoothing is too slow for us we may take another look at this.

PR #176, opened 2017-10-27
Fixing an error on inputs with lots of empty lines

artemsok:


sockeye/inference.py
```diff
@@ -661,43 +661,40 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
-            # translate
             translated_batch = self.translate_nd(*self._get_inference_input(batch))
             # truncate to remove duplicated translations
             if rest > 0:
                 translated_batch = translated_batch[:-rest]
             translations.extend(translated_batch)
 
-        # concatenate translations corresponding to spans of chunk markers with the same input id
+        # concatenate translations with the same input id into a dict of complete translations
+        concat_translations = dict()
```
fhieber: you might need a type annotation here to make mypy happy

sockeye/utils.py
```diff
@@ -710,3 +711,16 @@ def create_operator(self, ctx, shapes, dtypes):
         return PrintValue(self.print_name,
                           print_grad=self.print_grad,
                           use_logger=self.use_logger)
+
+
+def grouper(iterable: Iterable, size: int) -> Iterable:
+    """
+    Collect data into fixed-length chunks or blocks
+    without either dicarding underfilled chunks or padding them
```
fhieber: missing period.
missing param doc string.

Also, why the 'either'? Why not: "without discarding underfilled chunks or padding them".
Maybe explicitly say that the last chunk might be smaller than size.

fhieber: typo: dicarding

sockeye/inference.py
artemsok: looks like it's happy with it already:)

PR #177, opened 2017-10-29
imported sacrebleu

mjpost:
Added reference BLEU implementation.


sacrebleu/setup.py
```diff
@@ -0,0 +1,74 @@
...
+    # Versions should comply with PEP440.  For a discussion on single-sourcing
+    # the version across setup.py and the project code, see
+    # https://packaging.python.org/en/latest/single_source_version.html
+    version = VERSION,
+
+    description = 'Hassle-free computation of shareable, comparable, and reproducible BLEU scores',
+    long_description = 'SacréBLEU is a standard BLEU implementation that downloads and manages WMT datasets, produces scores on detokenized outputs, and reports a string encapsulating BLEU parameters, facilitating the production of shareable, comparable BLEU scores.',
+
+    # The project's main homepage.
+    url = 'https://github.com/mjpost/sacreBLEU',
```
fhieber: You should update this URL and the other information in this file.

sacrebleu/test.sh
```diff
@@ -0,0 +1,42 @@
...
+        sys=$(basename $sgm .sgm | perl -pe 's/newstest2017\.//')
+        txt=$(dirname $sgm | perl -pe 's/sgm/txt/')/$(basename $sgm .sgm)
+        src=wmt17-submitted-data/sgm/sources/newstest2017-$source$target-src.$source.sgm
+        ref=wmt17-submitted-data/sgm/references/newstest2017-$source$target-ref.$target.sgm
+
+        bleu1=$($MOSES/scripts/generic/mteval-v13a.pl -c -s $src -r $ref -t $sgm 2> /dev/null | grep "BLEU score" | cut -d' ' -f9)
+        bleu1=$(echo "print($bleu1 * 100)" | python)
+        bleu2=$(cat $txt | sacrebleu -t wmt17 -l $source-$target | cut -d' ' -f3)
+
+        echo "$source-$target $sys mteval: $bleu1 sacreBLEU: $bleu2"
```
fhieber: would it be hard to make this test and actual test, i.e. fail if results are not as expected?

mjdenkowski:
Great addition to the toolkit!  What do we think about creating a `contrib` directory for `sacrebleu` and other things we've discussed pulling in like a tokenizer or BPE?  Following the Moses model, `contrib` would be software distributed with Sockeye that--while not tightly integrated--makes for an all-in-one MT toolkit.  Over time, we can move some of the cool features we want, like automatically downloading and pre-processing data, into Sockeye modules (full integration, documentation, unit tests, etc.).  Until then, we can import independent tools into `contrib`.

sacrebleu/setup.py
mjpost: Ah, missed that one, will do.

sacrebleu/test.sh
mjpost: Not hard, will do that.

mjpost:
@tdomhan, thanks for the tip, I'll add this.

@mjdenkowski I think putting it under `contrib/` makes a lot of sense. I'll fix this in an amended commit, unless there are objections from others.

contrib/sacrebleu/sacrebleu.py
```diff
@@ -0,0 +1,820 @@
...
+        process_to_text(rawfile, outfile)
+        found.append(outfile)
+
+    return found
+
+
+BLEU = namedtuple('BLEU', 'score, ngram1, ngram2, ngram3, ngram4, bp, sys_len, ref_len')
+
+
+def compute_bleu(instream, refstreams, smooth=0., force=False, lc=False, tokenize=False) -> BLEU:
```
artemsok: Without having any strong feelings about any of the BLEU flavours, are we going to sync this BLEU implementation with the one in sockeye? 

The latter has some minimal speed optimizations (for the WMT'17 bandit learning shared task):

```python
from random import choice
from string import ascii_uppercase
from collections import defaultdict
from itertools import  tee, islice
from sockeye.blue import zipngram

a=' '.join(choice(ascii_uppercase) for i in range(100))

%timeit extract_ngrams(a, 4)
10000 loops, best of 3: 129 µs per loop

%timeit for i in range(4): zipngram(a, i)
10000 loops, best of 3: 56.8 µs per loop
```

fhieber: Yes, if possible we should rely on sacrebleus BLEU implementation for the metric computation in sockeye.evaluate and the checkpoint_decoder.

mjpost: I tested this version and it's slower than the current one.

    def extract_ngrams_fast(line, max=4):
        tokens = line.split()
        ngrams = Counter()
        for n in range(1, max + 1):
            ngrams += Counter(zip(*(islice(it, pos, None) for pos, it in enumerate(tee(tokens, n + 1)))))
        return ngrams

    a = ' '.join(choice(ascii_uppercase) for i in range(100))
    print(timeit.timeit(lambda: extract_ngrams(a), number=10000))
    print(timeit.timeit(lambda: extract_ngrams_fast(a), number=10000))

This prints:

    2.814126971992664
    3.3099639220017707

But I will rejigger things so that Sockeye makes use of sacreBLEU and uses it as a requirement.

mjpost:
Pulled in master.

mjpost:
please kill me

contrib/sacrebleu/sacrebleu.py
```diff
@@ -0,0 +1,894 @@
...
+            elif smooth == 'floor':
+                precisions[n] = 100. * smooth_floor / total[n]
+        else:
+            precisions[n] = 100. * correct[n] / total[n]
+
+    brevity_penalty = 1.0
+    if sys_len < ref_len:
+        brevity_penalty = math.exp(1 - ref_len / sys_len) if sys_len > 0 else 0.0
+
+    bleu = 1. * brevity_penalty * math.exp(sum(map(my_log, precisions)) / NGRAM_ORDER)
```
artemsok: If none of the hypothesis had any 4-grams, do we still want to divide by 4? (cf. effective_order in bleu.py)

If all hypothesis are completely empty (not just missing some high-order n-grams), looks like the BLEU will not be 0 - is it expected?

BTW, what's the point of multiplying by 1?

contrib/sacrebleu/setup.py
```diff
@@ -0,0 +1,103 @@
...
+        # that you indicate whether you support Python 2, Python 3 or both.
+        'Programming Language :: Python :: 3 :: Only',
+    ],
+
+    # What does your project relate to?
+    keywords = ['machine translation, evaluation, NLP, natural language processing, computational linguistics'],
+
+    # You can just specify the packages manually here if your project is
+    # simple. Or you can use find_packages().
+#    packages = ['sacreBLEU'],
```
tdomhan: can you remove this?

contrib/sacrebleu/setup.py
```diff
@@ -0,0 +1,103 @@
...
+A setuptools based setup module.
+
+See:
+- https://packaging.python.org/en/latest/distributing.html
+- https://github.com/pypa/sampleproject
+"""
+
+# Always prefer setuptools over distutils
+from setuptools import setup
+import sacrebleu
```
tdomhan: you should not include scarebleu here. This is actually why Travis is failing. In python 3.4 `typing` is a library that needs to be installed whereas in python 3.5+ it comes shipped with python. Now importing scarebleu in setup.py will try to import the `typing` module, which will only be installed once setup.py from sacrebleu has finished running. Hope that makes sense.

You will probably need to do a similar version regex dance as we do in the Sockeye setup.py:
https://github.com/awslabs/sockeye/blob/master/setup.py#L26

contrib/sacrebleu/sacrebleu.py
mjpost: Yes, this is correct. If there are 0 four-grams and there was no smoothing, it won't matter. This also passes all sockeye tests and is exactly the same as metal-v13a.pl in Moses.

1-multiplication was a holdover, I'll remove it.

contrib/sacrebleu/setup.py
mjpost: Thank you!

contrib/sacrebleu/setup.py
mjpost: Yes.

mjpost:
Okay! I added effective bleu smoothing per @artemsok's request, and also added a test case for it to the unit tests. I think this is ready for a merge.

PR #178, opened 2017-10-30
Reduce memory footprint of data i/o: added a SentenceReader that stre…

fhieber:
…ams integer sequences from disk. Removed label arrays from ParallelBucketSentenceIter. Labels are created on the fly from target sequence.

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -173,6 +173,10 @@ def add_io_args(params):
     data_params.add_argument('--target', '-t',
                              required=True,
                              help='Target side of parallel training data.')
+    data_params.add_argument('--limit',
+                             default=None,
+                             type=int,
+                             help="Maximum number of training sequences to read. Default: %(default)s.")
 
     data_params.add_argument('--validation-source', '-vs',
                              required=True,
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -100,24 +100,6 @@ def get_bucket(seq_len: int, buckets: List[int]) -> Optional[int]:
     return buckets[bucket_idx]
 
 
-def read_parallel_corpus(data_source: str,
-                         data_target: str,
-                         vocab_source: Dict[str, int],
-                         vocab_target: Dict[str, int]) -> Tuple[Iterator[List[int]], Iterator[List[int]]]:
-    """
-    Loads source and target data, making sure they have the same length.
-
-    :param data_source: Path to source training data.
-    :param data_target: Path to target training data.
-    :param vocab_source: Source vocabulary.
-    :param vocab_target: Target vocabulary.
-    :return: Tuple of (source sentences, target sentences).
-    """
-    source_sentences = SentenceReader(data_source, vocab_source, add_bos=False)
-    target_sentences = SentenceReader(data_target, vocab_target, add_bos=True)
-    return source_sentences, target_sentences
-
-
 def length_statistics(source_sentences: Iterable[List[Any]], target_sentences: Iterable[List[int]]) -> Tuple[float, float]:
     """
     Returns mean and standard deviation of target-to-source length ratios of parallel corpus.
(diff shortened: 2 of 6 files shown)


fhieber:
I realized that this is only mxnet12 compatible due to multi-indexing support.
I will only merge this once we have mxnet-12 update in place. But reviewing can start before :)

sockeye/arguments.py
```diff
@@ -173,6 +173,10 @@ def add_io_args(params):
     data_params.add_argument('--target', '-t',
                              required=True,
                              help='Target side of parallel training data.')
+    data_params.add_argument('--limit',
+                             default=None,
+                             type=int,
+                             help="Maximum number of training sequences to read. Default: %(default)s.")
```
mjdenkowski: What use case are we thinking of for the `--limit` option?  It definitely makes it more efficient to train on a subset of the data, though sizing down data usually includes some kind of selection technique that would pick something other than the first N sentences and create new files.

fhieber: Nothing specific really. We had the limit argument to reading alreay and exposing it could be useful for quick tests.

mjdenkowski: Testing is a good use case--makes sense.

sockeye/data_io.py
```diff
@@ -167,18 +148,16 @@ def get_training_data_iters(source: str, target: str,
...
     logger.info("Creating train data iterator")
-    train_source_sentences, train_target_sentences = read_parallel_corpus(source,
-                                                                          target,
-                                                                          vocab_source,
-                                                                          vocab_target)
-
-    max_observed_source_len = max((len(s) for s in train_source_sentences if len(s) <= max_seq_len_source), default=0)
-    max_observed_target_len = max((len(t) for t in train_target_sentences if len(t) <= max_seq_len_target), default=0)
+    train_source_sentences = SentenceReader(source, vocab_source, add_bos=False, limit=sequence_limit)
+    train_target_sentences = SentenceReader(target, vocab_target, add_bos=True, limit=sequence_limit)
```
mjdenkowski: Without an explicit call to `read_parallel_corpus`, we could add some comments about when/where we're streaming or storing data.  For instance, we could add to the docstring for `ParallelBucketSentenceIter` that it streams the corpus from disk, stores retained sentences as a NumPy array, etc.  I had to scan through the code to remind myself exactly what the process looks like.

fhieber: I can add a couple of comments, but be aware that this PR does not give us a streaming data iterator. It just streams the int sequences from disk WHILE assigning to buckets. The resulting ndarrays in the buckets are still all in memory.

fhieber:
@tdomhan Rebased onto latest master (mxnet 0.12) and checks now pass here

sockeye/data_io.py
```diff
@@ -312,30 +292,48 @@ def tokens2ids(tokens: Iterable[str], vocab: Dict[str, int]) -> List[int]:
...
+        assert C.UNK_SYMBOL in vocab
+        assert vocab[C.PAD_SYMBOL] == C.PAD_ID
+        assert C.BOS_SYMBOL in vocab
+        assert C.EOS_SYMBOL in vocab
+        self._iter = None  # type: Optional[Iterator]
+        self._iterated_once = False
+        self.count = 0
+
+    def __iter__(self):
+        self._iter = read_content(self.path, self.limit)
```
tdomhan: I'm not a huge fan of this as the following will not do what you'd expect:
```
iter1 = iter(sentence_reader)
el1 = iter1.next()
iter2 = iter(sentence_reader) # resets the iter
el2 = iter1.next() # gives the first element a second time.
```
how about raising `if self._iter is not None` to make it explicit that this kind of usage is forbidden?

sockeye/data_io.py
```diff
@@ -100,27 +100,7 @@ def get_bucket(seq_len: int, buckets: List[int]) -> Optional[int]:
...
-    """
-    source_sentences = read_sentences(data_source, vocab_source, add_bos=False)
-    target_sentences = read_sentences(data_target, vocab_target, add_bos=True)
-    check_condition(len(source_sentences) == len(target_sentences),
-                    "Number of source sentences does not match number of target sentences")
-    return source_sentences, target_sentences
-
-
-def length_statistics(source_sentences: List[List[Any]], target_sentences: List[List[int]]) -> Tuple[float, float]:
+def length_statistics(source_sentences: Iterable[List[Any]], target_sentences: Iterable[List[int]]) -> Tuple[float, float]:
```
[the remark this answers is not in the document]
tdomhan: ok, it's not part of the PR :P but would make sense to change it now that we touch this method ;)

tdomhan:
great change that should reduce the memory footprint considerably, as python lists are quite memory hungry. This should at least reduce the problem before we we move to some sharding/streaming solution.

sockeye/data_io.py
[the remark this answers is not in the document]
fhieber: Good question, I will change.

sockeye/data_io.py
fhieber: oh good catch, I will add a check!

PR #180, opened 2017-11-01
Lexicon-based vocabulary selection for faster decoding

mjdenkowski:
This update implements the target vocabulary selection technique discussed in [Devlin (2017)](https://arxiv.org/abs/1705.01991) and others to speed up decoding.  For each input sentence, a lexicon is used to slice the target vocabulary to include only the top k translations of each source word.  This dramatically reduces the size of the matrices used in the expensive logit/softmax computation at each decoder time step.  This can be applied to any model with an output layer that scores the target vocabulary (RNN, Transformer, CNN).

MXNet doesn't natively support all of the operations we need to implement this, so design feedback is welcome.

The main additions in this update include:
- A new CLI for building a top-k lexicon from a fast_align format lex table.  The file is stored as JSON using the model's vocabulary (fast load, debuggable).  The main method lives in sockeye.lexicon since existing lexicon functionality doesn't need a separate CLI.
- Decoders can now return logits (default behavior) or inputs to logit computation (skipping expensive logits in graph).
- Decoders also have an NDArray implementation of computing logits (mx.nd version of the symbolic output layer).
- During inference, a custom target vocabulary can be created for each input by slicing via the results of the top-k lexicon and source sentence.
- The final logit computation and softmax can be carried out using the mx.nd versions on the sliced vocabulary.  As much as possible is done in the same context.
- Unit tests for building top-k lexicons and integration tests using vocab selection with all major architectures.

Implementation notes:
- Ideally, we would use something like [TensorFlow's set operations](https://www.tensorflow.org/api_docs/python/tf/sets) for creating custom vocabs.  Unfortunately MXNet doesn't have set ops, so we use [NumPy's arraysetops](https://docs.scipy.org/doc/numpy-1.13.0/reference/routines.set.html).
- Ideally, we would use [MXNet's sparse ndarray/symbol API](https://mxnet.incubator.apache.org/versions/master/api/python/ndarray/sparse.html) to slice matrices, but right now it's not production ready.  Operations we're interested in are not yet supported and default back to the dense versions (no speedup, likely slowdown).


mjdenkowski:
@fhieber That's a good idea.  I was thinking about something like OutputLayer when I added the common `decode_step_logits_nd` method.  Now that it's merged in, we can consolidate implementations.

sockeye/lexicon.py
```diff
@@ -158,3 +183,108 @@ def _init_default(self, sym_name, arr):
...
+            # Unk token will always be part of target vocab, so no need to track it here
+            if src_id == src_unk_id or trg_id == trg_unk_id:
+                continue
+            _lex[src_id][trg_id] = prob
+        # Sort and copy top-k trg_ids to lex array row src_id
+        for src_id, trg_entries in _lex.items():
+            top_k = list(sorted(trg_entries.items(), key=operator.itemgetter(1), reverse=True))[:k]
+            self.lex[src_id, :len(top_k)] = list(sorted(trg_id for (trg_id, _) in top_k))
+            # Free memory after copy
+            trg_entries.clear()
```
fhieber: would it make sense to clear _lex after this loop as well?

test/unit/test_lexicon.py
```diff
@@ -0,0 +1,69 @@
...
+    vocab = dict((y, x) for (x, y) in enumerate(C.VOCAB_SYMBOLS + vocab_list))
+    k = 2
+    lex = sockeye.lexicon.TopKLexicon(vocab, vocab)
+
+    # Create from known lexicon
+    with TemporaryDirectory(prefix="test_topk_lexicon.") as work_dir:
+        input_lex_path = os.path.join(work_dir, "input.lex")
+        with open(input_lex_path, "w") as out:
+            for line in lexicon:
+                print(line, file=out)
```
fhieber: why print here? would it make sense to assert on comparing `line` with an expected line?

sockeye/arguments.py
```diff
@@ -798,6 +825,11 @@ def add_inference_args(params):
                                help='Number of target-to-source length ratio standard deviations from training to add '
                                     'to calculate maximum output length for beam search for each sentence. '
                                     'Default: %(default)s.')
+    decode_params.add_argument('--restrict-lexicon',
+                               type=str,
+                               default=None,
+                               help="Specify top-k lexicon to restrict output vocabulary based on source.  See lexicon "
+                                    "module.")
```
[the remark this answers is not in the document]
mjdenkowski: Done

sockeye/decoder.py
```diff
@@ -58,6 +58,9 @@ class Decoder(ABC):
     For the inference module to be able to keep track of decoder's states
     a decoder provides methods to return initial states (init_states), state variables and their shapes.
     """
+    def __init__(self):
```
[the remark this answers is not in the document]
mjdenkowski: Done

sockeye/decoder.py
```diff
@@ -58,6 +58,9 @@ class Decoder(ABC):
...
+        # Tracked to find params for logit computation
+        self.output_layer = None  # type: layers.OutputLayer
```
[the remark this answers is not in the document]
mjdenkowski: Looks like mypy is alright with this.

sockeye/lexicon.py
mjdenkowski: We should be okay since _lex goes out of scope right after this.

test/unit/test_lexicon.py
mjdenkowski: Looked back and saw this wasn't documented well enough.  We write a fast_align format lex table to the temp dir, then call sockeye.lexicon to build the top-k lexicon from the fast_align lex table.  I added some comments.

fhieber: You are right sorry. I am still getting used to print writing to file descriptors :)

sockeye/inference.py
```diff
@@ -805,8 +848,16 @@ def _decode_step(self,
                 sequences = mx.nd.slice_axis(sequences, axis=1, begin=0, end=target_max_length)
 
         model_probs, model_attention_probs, model_states = [], [], []
-        for model, state in zip(self.models, states):
-            probs, attention_probs, state = model.run_decoder(sequences, bucket_key, state)
+        for model, out_w, out_b, state in itertools.zip_longest(
```
fhieber: good catch with the izip longest. I assume this is because not all models may have caching enabled and could return None here. 
Could you add a comment for this?

sockeye/inference.py
```diff
@@ -881,6 +932,21 @@ def _beam_search(self,
...
+        models_output_layer_b = list()
+        pad_dist = self.pad_dist
+        vocab_slice_ids = None  # type: np.ndarray
+        if self.restrict_lexicon:
+            # TODO: See note in method about migrating to pure MXNet when operations are supported.
+            vocab_slice_ids = self.restrict_lexicon.get_trg_ids(source.astype("int32").asnumpy())
+            pad_dist = mx.nd.full((self.batch_size * self.beam_size, vocab_slice_ids.shape[0]),
+                                  val=np.inf, ctx=self.context)
+            for m in self.models:
+                models_output_layer_w.append(mx.nd.array(m.output_layer_w[vocab_slice_ids], ctx=self.context))
```
fhieber: The reason you are saving <model>.output_layer_{w,b} as numpy arrays is because of these two lines because you have a list of indices in vocab_slice_ids, correct?
Instead of re-creating the mx.nd.array here every single time you can also use MXNet's `take` operation for this, if m.output_layer_{w,b} is NDArray and `lexicon.get_trg_ids` would return NDArray.
Even in the new 'fluent' mode:
```
...apend(m.output_layer_w.take(vocab_slice_ids))
```
Copying the vocab_slice_ids to the context is probably cheaper than copying the weight and bias parameters back and forth.

sockeye/inference.py
```diff
@@ -125,6 +136,10 @@ def initialize(self, max_input_length: int, get_max_output_length_function: Call
         self.encoder_module.init_params(arg_params=self.params, allow_missing=False)
         self.decoder_module.init_params(arg_params=self.params, allow_missing=False)
 
+        if self.cache_output_layer_w_b:
+            self.output_layer_w = self.params[self.decoder.output_layer.w.name].asnumpy()
```
fhieber: see my comment below in inference.py about mx.nd.take. This might save you the asnumpy here.

sockeye/decoder.py
```diff
@@ -91,18 +94,23 @@ def decode_step(self,
                     target: mx.sym.Symbol,
                     target_max_length: int,
                     source_encoded_max_length: int,
-                    *states: mx.sym.Symbol) \
-            -> Tuple[mx.sym.Symbol, mx.sym.Symbol, List[mx.sym.Symbol]]:
+                    *states: mx.sym.Symbol,
```
fhieber: What about changing the interface here to always return 
pre-output-layer symbol AND the logits?
The caller of this method can then decide whether it wants to use logit_inputs or logits or both. If one is not needed it gets out of scope and within the decoders returning both is no additional overhead. This would simplify the signature of this method a bit. 
It also would make implementations of decoders a bit less hard to read as you can get rid of all the `if then else None` lines.

sockeye/inference.py
```diff
@@ -84,6 +90,11 @@ def __init__(self,
         self.decoder_module = None  # type: Optional[mx.mod.BucketingModule]
         self.decoder_default_bucket_key = None  # type: Optional[Tuple[int, int]]
         self.decoder_data_shapes_cache = None  # type: Optional[Dict]
+        self.decoder_return_logit_inputs = decoder_return_logit_inputs
+
+        self.cache_output_layer_w_b = cache_output_layer_w_b
+        self.output_layer_w = None  # type: np.ndarray
```
fhieber: see my comments below, could be mx NDArray

sockeye/inference.py
```diff
@@ -921,11 +989,14 @@ def _beam_search(self,
                     scores_accumulated_np[rows] = utils.smallest_k(sliced_scores, self.beam_size, t == 1)
                 # offsetting since the returned smallest_k() indices were slice-relative
                 best_hyp_indices_np[rows] += rows.start
+            # Map from restricted to full vocab ids if needed
+            if self.restrict_lexicon:
+                best_word_indices_np[:] = vocab_slice_ids[best_word_indices_np]
```
fhieber: if `vocab_slice_ids` becomes NDArray this probably needs to move below line 998

test/common.py
```diff
@@ -174,6 +192,31 @@ def run_train_translate(train_params: str,
...
+            # Translate corpus with restrict-lexicon
+            out_path = os.path.join(work_dir, "out-restrict.txt")
+            params = "{} {} {} {}".format(sockeye.translate.__file__,
+                                          _TRANSLATE_PARAMS_COMMON.format(model=model_path,
+                                                                          input=dev_source_path,
+                                                                          output=out_path),
+                                          translate_params,
+                                          _TRANSLATE_PARAMS_RESTRICT.format(json=json_path))
+            with patch.object(sys, "argv", params.split()):
+                sockeye.translate.main()
```
fhieber: should we also compute BLEU on the output of translate with lexicon restriction and return it from this function? We could then explicitly test in the integration tests for not loosing too much BLEU when using it.

sockeye/decoder.py
mjdenkowski: That's a fair point.  Right now, none of our decoders require extra computation to return both.

sockeye/inference.py
mjdenkowski: Added.

test/common.py
mjdenkowski: That would be a good system test.  It looks like we're getting a BLEU score of about 0.001 on the fast integration tests so unfortunately there's not much room to test for degradation.

fhieber: Yes, true. I actually meant the system tests that run for longer and produce reasonable ppls and BLEU scores. I always confuse the two...

sockeye/inference.py
mjdenkowski: Good catch--that's the op I was looking for.  It would be more natural if NDArray supported slicing by arrays like NumPy, but I see why they would want to use `take` for consistency between nd and sym APIs.

fhieber: I agree, numpy-like indexing and item assignment would be very useful to have. I think its on their roadmap for MXNet 0.13(?).
This just caused me some trouble as well here: #187 :)

sockeye/inference.py
```diff
@@ -936,15 +936,16 @@ def _beam_search(self,
...
         models_output_layer_b = list()
         pad_dist = self.pad_dist
-        vocab_slice_ids = None  # type: np.ndarray
+        vocab_slice_ids = None  # type: mx.nd.NDArray
         if self.restrict_lexicon:
-            # TODO: See note in method about migrating to pure MXNet when operations are supported.
-            vocab_slice_ids = self.restrict_lexicon.get_trg_ids(source.astype("int32").asnumpy())
+            # TODO: See note in method about migrating to pure MXNet when set operations are supported.
+            #       We currently convert source to NumPy and target ids back to NDArray.
+            vocab_slice_ids = mx.nd.array(self.restrict_lexicon.get_trg_ids(source.astype("int32").asnumpy()))
```
fhieber: you probably need to set ctx=self.context here to not crash on the GPU

sockeye/inference.py
```diff
@@ -137,8 +137,8 @@ def initialize(self, max_input_length: int, get_max_output_length_function: Call
         self.decoder_module.init_params(arg_params=self.params, allow_missing=False)
 
         if self.cache_output_layer_w_b:
-            self.output_layer_w = self.params[self.decoder.output_layer.w.name].asnumpy()
-            self.output_layer_b = self.params[self.decoder.output_layer.b.name].asnumpy()
+            self.output_layer_w = mx.nd.array(self.params[self.decoder.output_layer.w.name], ctx=self.context)
```
fhieber: this always copies even if source and target context are the same.
You could use ndarray.as_in_context() instead:
https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.NDArray.as_in_context

sockeye/decoder.py
```diff
@@ -1218,9 +1225,12 @@ def decode_step(self,
                                                            axis=2, begin=0, end=1),
                                          shape=(0, -1))
 
+        # logit inputs aka target_hidden
+        logit_inputs = mx.sym.reshape(target_hidden, shape=(-2,), name=C.LOGIT_INPUTS_NAME)
```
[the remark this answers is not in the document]
mjdenkowski: If we try to return a symbol (like target_hidden) without assigning a new name, MXNet throws error about multiple symbols having the same name.  This reshape is basically a no-op that attaches a name to the symbol.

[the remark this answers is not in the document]
fhieber: Makes sense!
I just saw that there is mxnet.symbol.identity which basically is the explicit no-op that allows you to assign a new name to a symbol; might be a bit clearer than reshape.

sockeye/decoder.py
```diff
@@ -92,17 +95,17 @@ def decode_step(self,
                     target_max_length: int,
                     source_encoded_max_length: int,
                     *states: mx.sym.Symbol) \
-            -> Tuple[mx.sym.Symbol, mx.sym.Symbol, List[mx.sym.Symbol]]:
+            -> Tuple[mx.sym.Symbol, mx.sym.Symbol, mx.sym.Symbol, List[mx.sym.Symbol]]:
         """
         Decodes a single time step given the previous word ids in target and previous decoder states.
-        Returns logits, attention probabilities, and next decoder states.
+        Returns logit inputs (optional), logits (optional), attention probabilities, and next decoder states.
```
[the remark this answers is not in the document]
tdomhan: mhh yeah, what does the optional mean?

sockeye/inference.py
```diff
@@ -24,6 +24,7 @@
 
 from . import constants as C
 from . import data_io
+from .lexicon import TopKLexicon
```
tdomhan: could you do `from . import lexicon` and `lexicon.TopKLexicon` instead to be consistent with the rest of the code?

sockeye/inference.py
```diff
@@ -46,6 +47,9 @@ class InferenceModel(model.SockeyeModel):
     :param checkpoint: Checkpoint to load. If None, finds best parameters in model_folder.
     :param softmax_temperature: Optional parameter to control steepness of softmax distribution.
     :param max_output_length_num_stds: Number of standard deviations as safety margin for maximum output length.
+    :param decoder_return_logit_inputs: Decoder returns inputs to logit computation instead of softmax over target
+                                        vocabulary.  Used when logits/softmax are handled separately.
+    :param cache_output_layer_w_b: Cache weights and biases for logit computation as NumPy arrays.
```
tdomhan: weights and biases below are mxnet ndarrays and not numpy arrays as mentioned in the comment, right?

sockeye/decoder.py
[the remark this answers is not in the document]
mjdenkowski: Nice find--that's what I was looking for.

sockeye/inference.py
mjdenkowski: Fixed.

sockeye/inference.py
mjdenkowski: Fixed.

PR #189, opened 2017-11-08
Remove RNN parameter packing and FusedRNN support

fhieber:
Although this is approved, this PR will include #196 once ready.


mjdenkowski:
+1 for removing unused/non-working code.

fhieber:
Thanks, let's hold off on merging this for now, since I have another change in the pipeline that would bump the major version

PR #191, opened 2017-11-09
Fix weight normalization with vocab selection decoding. Fixed a corne…

fhieber:
…r case when reduced vocab for a sentence is smaller than the beam size. Tightened and fixed system tests. They pass now.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -137,7 +137,14 @@ class InferenceModel(model.SockeyeModel):
         self.decoder_module.init_params(arg_params=self.params, allow_missing=False)
 
         if self.cache_output_layer_w_b:
-            self.output_layer_w = self.params[self.decoder.output_layer.w.name].as_in_context(self.context)
+            if self.decoder.output_layer.weight_normalization:
+                # precompute normalized output layer weight imperatively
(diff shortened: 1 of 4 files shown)


sockeye/inference.py
```diff
@@ -942,6 +949,18 @@ def _beam_search(self,
...
+            if vocab_slice_ids.shape[0] < self.beam_size + 1:
+                # This fixes an edge case for toy models, where the number of vocab ids from the lexicon is
+                # smaller than the beam size.
+                logger.warning("Padding vocab_slice_ids (%d) with EOS to have at least %d+1 elements to expand",
+                               vocab_slice_ids.shape[0], self.beam_size)
+                n = self.beam_size - vocab_slice_ids.shape[0] + 1
+                vocab_slice_ids = mx.nd.concat(vocab_slice_ids,
+                                               mx.nd.full((n,), val=self.vocab_target[C.EOS_SYMBOL]),
+                                               dim=0)
+                assert vocab_slice_ids.shape[0] == self.beam_size + 1
```
tdomhan: I guess this assert was for development and can be removed, right?

fhieber: yes, will remove

PR #193, opened 2017-11-09
Changing to non-mkl mxnet version.

tdomhan:


fhieber:
For reference: https://github.com/apache/incubator-mxnet/issues/8532

PR #194, opened 2017-11-09
Preventing accidental use of setuptools install.

tdomhan:


fhieber:
Related note: should we do a `pip install .` in the travis build?

tdomhan:
we don't install sockeye in travis, no?

tdomhan:
let's actually hold off from merging. I want to run some more tests with this to make sure we don't block anything.

tdomhan:
`bdist_wheel` actually runs `install`.... so maybe it's not a good idea to forcefully forbid using `setup.py install`...

fhieber:
Makes sense.
Regarding travis: no we dont install but we could as an additional test maybe.

tdomhan:
will discard for now...

PR #196, opened 2017-11-09
[PR against #189] Refactor embedding and output layers

fhieber:
NOTE: this pr is against the #189 request. Once approved, I would merge on the #189 branch to have only one major version bump and update the changelog there.

This promotes source & target embedding layers, as well as the output layer to core model components in a SockeyeModel instance (see `model.py::SockeyeModel._build_model_components()`).

As a result, the various weight tying settings are easier to get right and less convoluted.
In addition, decoders are conceptually simpler, as they are returning their decoder output without computing logits. This is now the task of `model.output_layer.` Various Config classes now have significantly less parameters. Less code is duplicated in decoders and Decoder classes are now much closer to Encoder classes.

There are 3 backwards incompatible changes:
1. output weight renamed to target_output_{weight,bias} since the "cls" string was a bit strange anyway.
2. Transformer models do not 'overwrite' --num-embed anymore as they did silently before. As a result there is an error thrown if `--num-embed != --transformer-model-size` which is more transparent.
3. Removed the lexical biasing functionality (Arthur ETAL'16). This never really worked out and was just cluttering the code since there a dependency between RNN attention  probabilities and the output layer.

Other comments
- All system tests still pass, but I also tightened the bounds a little bit to detect regressions better.
- The up-scaling of embedding vectors before the fixed positional embeddings are added in transformer models is now done in the fixed positional embedding class, which is conceptually cleaner. As this has a negative effect on CNN models, I added a flags to this class to control scaling.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/model.py
```diff
@@ -56,33 +56,42 @@ def __init__(self,
...
 
 
 class SockeyeModel:
     """
     SockeyeModel shares components needed for both training and inference.
+    The main components of a Sockeye model are
+    1) Source embedding
+    2) Target embedding
+    3) Encoder
+    4) Deoder
```
davvil: Type: De**c**oder

sockeye/model.py
```diff
@@ -154,26 +166,66 @@ def save_version(folder: str):
...
 
+    def _get_embed_weights(self) -> Tuple[mx.sym.Symbol, mx.sym.Symbol, mx.sym.Symbol]:
+        """
+        Returns embedding parameters for source and target.
+
+        :return: Tuple of source and target parameter symbols.
+        """
+        assert isinstance(self.config.config_embed_source, encoder.EmbeddingConfig)
+        assert isinstance(self.config.config_embed_target, encoder.EmbeddingConfig)
+        w_embed_source = mx.sym.Variable(C.SOURCE_EMBEDDING_PREFIX + "weight",
```
davvil: Put that into an if-clause for shared weights. I know it would not really make a difference, but it makes the code clearer.

sockeye/model.py
```diff
@@ -154,26 +166,66 @@ def save_version(folder: str):
         with open(fname, "w") as out:
             out.write(__version__)
 
+    def _get_embed_weights(self) -> Tuple[mx.sym.Symbol, mx.sym.Symbol, mx.sym.Symbol]:
+        """
+        Returns embedding parameters for source and target.
+
+        :return: Tuple of source and target parameter symbols.
+        """
+        assert isinstance(self.config.config_embed_source, encoder.EmbeddingConfig)
```
davvil: Ugly python lacking proper type checking :-D However, are we expecting that this is not the case?

sockeye/train.py
```diff
@@ -92,6 +91,16 @@ def check_arg_compatibility(args: argparse.Namespace):
     check_condition(args.optimized_metric == C.BLEU or args.optimized_metric in args.metrics,
                     "Must optimize either BLEU or one of tracked metrics (--metrics)")
 
+    if args.encoder == C.TRANSFORMER_TYPE:
+        check_condition(args.transformer_model_size == args.num_embed[0],
```
davvil: If that is the case, could we remove the args for transformer model sizes and just take the corrsponding embedding sizes?

sockeye/decoder.py
```diff
@@ -58,54 +56,49 @@ class Decoder(ABC):
...
-        Returns logit inputs, logits, attention probabilities, and next decoder states.
+        Decodes a single time step given the embedded target sequence and previous decoder states.
+        Returns decoder representation for the next prediction, attention probabilities, and next decoder states.
         Implementations can maintain an arbitrary number of states.
 
-        :param target: Previous target word ids. Shape: (batch_size, target_max_length).
-        :param target_max_length: Size of time dimension in prev_word_ids.
+        :param target_embed: Embedded target sequence. Shape: (batch_size, target_embed_max_length, target_num_embed).
+        :param target_embed_lengths: Lengths of embedded target sequences. Shape: (batch_size,).
+        :param target_embed_max_length: Size of embedded target sequence dimension.
```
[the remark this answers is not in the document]
fhieber: ok

sockeye/decoder.py
```diff
@@ -58,54 +56,49 @@ class Decoder(ABC):
...
+        :param target_embed_prev: Previous target word embedding. Shape: (batch_size, target_num_embed).
```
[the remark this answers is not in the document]
fhieber: no, its actually just one word :)

sockeye/inference.py
```diff
@@ -195,28 +205,47 @@ def sym_gen(bucket_key: Tuple[int, int]):
...
                                                 source_encoded_seq_len,
                                                 *states)
-            if self.softmax_temperature is not None:
-                logits /= self.softmax_temperature
 
-            # Return logit inputs or softmax over target vocab
             if self.decoder_return_logit_inputs:
-                outputs = logit_inputs
+                # skip output layer in graph
+                outputs = mx.sym.identity(target_decoded, name=C.LOGIT_INPUTS_NAME)
```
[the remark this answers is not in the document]
fhieber: yes exactly

sockeye/inference.py
```diff
@@ -289,16 +318,16 @@ def run_decoder(self,
         """
         Runs forward pass of the single-step decoder.
 
-        :return: Probability distribution over next word, attention scores, updated model state.
+        :return: logit_inputs or probability distribution over next word, attention scores, updated model state.
```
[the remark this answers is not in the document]
fhieber: Changed to
":return: Decoder stack output (logit inputs or probability distribution), attention scores, updated model state."

sockeye/model.py
fhieber: fixed

sockeye/model.py
fhieber: we are not expecting this, but otherwise the IDE gives a type warning :/

sockeye/model.py
fhieber: not sure I know what you mean.
Its much easier with overwriting the references below.

sockeye/train.py
fhieber: not sure, args is fixed and cannot be changed and I don't like these implicit overrides. This fails early and makes it clear what to change.
In the future we can think about an optional projection layer from num_embed to model_size if needed.

sockeye/decoder.py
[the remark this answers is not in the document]
davvil: There are batch_size words ;-) But it's ok to leave it as singular as well.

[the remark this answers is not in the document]
fhieber: true but out convention before was also to say "length" to (batch, 1) and "lengths" to (batch, time).

sockeye/model.py
davvil: I am not a big fan of (relatively complex) var initializations and then overwriting them. I would prefer something like

```python
if not self.config.weight_tying:
  w_embed_source = ...
  w_embed_target = ...
  w_out_target = ...
else:
  [...]
```
Makes it clearer, for me at least.

davvil: Hmm, looking at the code in more detail I see that they are overwritten conditionally! Forget about this, then.

sockeye/train.py
davvil: I was referring to removing the argument altogether, not to overwriting args.

fhieber: I see. I would be in favor of keeping it as it refers to the parameter introduced by Vaswani etal, 17. Overloading --num-embed for something else seems quite complicated. Especially since the other architectures have their own 'size' args, for example `--rnn-num-hidden`.

sockeye/encoder.py
```diff
@@ -392,11 +354,19 @@ def encode(self,
         :param seq_len: sequence length.
         :return: (batch_size, source_seq_len, num_embed)
         """
-        embedding = mx.sym.broadcast_add(data,
-                                         mx.sym.BlockGrad(mx.symbol.Custom(length=seq_len,
-                                                                           depth=self.num_embed,
-                                                                           name="%spositional_encodings" % self.prefix,
-                                                                           op_type='positional_encodings')))
+        # add positional embeddings to data
+        if self.scale_input:
```
tdomhan: would it make sense to call the parameters `scale_up_init` and `scale_down_positions` to make it a bit clearer why one is `**0.5` while the other is `**-0.5`?

sockeye/model.py
tdomhan: even if one used `self.config_embed_source = ... # type: encoder.EmbeddingConfig` in the `ModelConfig` class?

sockeye/model.py
```diff
@@ -154,26 +166,65 @@ def save_version(folder: str):
...
+        assert isinstance(self.config.config_embed_target, encoder.EmbeddingConfig)
+        self.embedding_source = encoder.Embedding(self.config.config_embed_source,
+                                                  prefix=C.SOURCE_EMBEDDING_PREFIX,
+                                                  embed_weight=embed_weight_source)
+        self.embedding_target = encoder.Embedding(self.config.config_embed_target,
+                                                  prefix=C.TARGET_EMBEDDING_PREFIX,
+                                                  embed_weight=embed_weight_target)
+
+        if self.config.weight_tying and C.WEIGHT_TYING_SOFTMAX in self.config.weight_tying_type:
```
tdomhan: wouldn't it make more sense to have this check in `_get_embed_weights `?

sockeye/layers.py
```diff
@@ -123,46 +123,34 @@ class OutputLayer:
...
-        if weight_tying:
-            utils.check_condition(num_hidden == num_embed,
-                                  "Weight tying requires target embedding size and decoder hidden size " +
-                                  "to be equal: %d vs. %d" % (num_embed, num_hidden))
-
-            logger.info("Tying the target embeddings and prediction matrix.")
-            assert embed_weight is not None, "Must provide embed_weight if weight_tying == True"
-            self.w = embed_weight
+        if weight is None:
+            self.w = mx.sym.Variable("target_output_weight", shape=(vocab_size, hidden_size))
```
tdomhan: why are you no longer adding the prefix? Who is to say that you wouldn't maybe want multiple output layers in the future.

sockeye/decoder.py
```diff
@@ -1006,19 +891,11 @@ def __init__(self,
...
-                                           self.config.vocab_size,
-                                           prefix=C.TARGET_EMBEDDING_PREFIX,
-                                           embed_weight=embed_weight,
-                                           dropout=config.embed_dropout)
         self.pos_embedding = encoder.get_positional_embedding(config.positional_embedding_type,
-                                                              config.num_embed,
+                                                              num_embed=config.num_embed,
                                                               max_seq_len=config.max_seq_len_target,
+                                                              fixed_scale_input=False,
+                                                              fixed_scale_positions=False,
```
tdomhan: it probably makes sense to also scale this for the CNN models?

sockeye/encoder.py
```diff
@@ -523,9 +499,17 @@ def get_num_hidden(self) -> int:
         return self.num_embed
 
 
-def get_positional_embedding(positional_embedding_type, num_embed, max_seq_len, prefix) -> PositionalEncoder:
+def get_positional_embedding(positional_embedding_type: str,
+                             num_embed: int,
+                             max_seq_len: int,
+                             fixed_scale_input: bool = False,
```
tdomhan: oh the "fixed" is for "fixed embedding"? Initially I read this is a fixed as in fixing something that is broken. It might make sense to add a docstring to explicitly state that this only affects the fixed embeddings, or maybe some naming like `scale_up_fixed_pos_embed`, `scale_down_fixed_embed_input`.

CHANGELOG.md
```diff
@@ -1,20 +1,26 @@
...
-All notable changes to this project will be documented in this file.
+All notable changes to the project are documented in this file.
 
-We use version numbers with three digits such as 1.0.0.
+Version numbers are of the form `1.0.0`.
 Any version bump in the last digit is backwards-compatible, in that a model trained with the previous version can still
 be used for translation with the new version.
-Any bump in the second digit indicates potential backwards incompatibilities, e.g. due to changing the architecture or
-simply modifying weight names.
+Any bump in the second digit indicates a major version change introduces potential backwards incompatibilities,
```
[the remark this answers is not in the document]
fhieber: ok will change

sockeye/encoder.py
fhieber: makes sense, will change

sockeye/layers.py
fhieber: Will add a prefix argument to the OutputLayer constructor

sockeye/model.py
fhieber: yes, will move it there

sockeye/decoder.py
fhieber: probably, but let's have this in a separate change.

sockeye/encoder.py
fhieber: renamed the parameters to be more explicit.

sockeye/inference.py
```diff
@@ -195,28 +205,47 @@ def sym_gen(bucket_key: Tuple[int, int]):
...
-            source_max_len, target_max_len = bucket_key
-            source_encoded_seq_len = self.encoder.get_encoded_seq_len(source_max_len)
+            source_seq_len, target_seq_len = bucket_key
+            source_embed_seq_len = self.embedding_source.get_encoded_seq_len(source_seq_len)
+            source_encoded_seq_len = self.encoder.get_encoded_seq_len(source_embed_seq_len)
 
             self.decoder.reset()
-            prev_word_ids = mx.sym.Variable(C.TARGET_NAME)
+            target = mx.sym.Variable(C.TARGET_NAME)
+            target_length = utils.compute_lengths(target)
```
tdomhan: minor: I guess it would make sense to use the plural `target_lengths` here as well.

fhieber: will fix

PR #197, opened 2017-11-10
 Fixed some typos and made minor modifications in data_io.py

xingniu:
Some fast and minor changes.


sockeye/data_io.py
```diff
@@ -262,7 +262,7 @@ def smart_open(filename: str, mode: str = "rt", ftype: str = "auto", errors:str
         return open(filename, mode=mode, encoding='utf-8', errors=errors)
 
 
-def read_content(path: str, limit: bool = None) -> Iterator[List[str]]:
+def read_content(path: str, limit: int = None) -> Iterator[List[str]]:
```
fhieber: Good catch, but could you change to "Optional[int]" since its set to None by default?

sockeye/data_io.py
```diff
@@ -481,14 +481,14 @@ def __init__(self,
         # operations that produce shapes larger than the default bucket size.  In these cases, MXNet
         # will silently allocate additional memory.
         self.provide_data = [
-            mx.io.DataDesc(name=source_data_name,
+            mx.io.DataDesc(name=self.source_data_name,
```
fhieber: fine with me, but any reason why this is important?

xingniu: It's not important. I just made it consistent with the use of all other parameters in __init__ (i.e. use self.*).

sockeye/data_io.py
xingniu: Done

PR #222, opened 2017-11-27
Sorting the sentences before batch decoding.

tdomhan:
- Sorting sentences before splitting them in batches.
 - Better default for the chunk size, which now depends on whether
 batching is enabled or not.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.14.1]
+### Changed
+ - Sorting sentences during decoding before splitting them into batches.
+ - Default chunk size: The default chunk size when batching is enabled is now batch_size * 500 during decoding to avoid
+  users accidentally forgetting to increase the chunk size.
+  
 ## [1.14.0]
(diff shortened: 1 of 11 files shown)


mjpost:
Good call on the chunk size — I had fixed that myself on an internal branch. Have you run any tests to see if there is a speed improvement here?

artemsok:
Looks good to me. One thing we might think of, after some benchmarking, is adapting the effective batch_size to the length distribution (treating the supplied one as an upper limit) to avoid fillers and for more uniform lengths inside a batch.

fhieber:
@tdomhan is this ready to merge after rebasing on latest master?

tdomhan:
Let's not merge yet. On some tests I observed different translations than the current master. I need to investigate what's going on.

tdomhan:
ok, it's not really a bug. What happens (and also happened before the PR) is that the maximum output length is dependent on the bucket a sentence falls in, which in turn depends on the other sentences it gets grouped with. This leads to cases where sentences can potentially lead to a different translation with bucketing. This is only really the case when the maximum length was hit with a single sentence. 

I'd say we merge this PR as the behavior stays the same (just now sentences will be grouped differently). We can then separately discuss if we want to change this behavior.

fhieber:
Thanks for the clarification, I will merge this then.

PR #226, opened 2017-11-28
Fast decoding for transformer models. Caches keys and values of self-attention before softmax.

fhieber:
This adds fast decoding for transformers by caching intermediate, previous outputs.
For this to work, inference is now using buckets of size 1 on the target side by default.
As a side effect, we can also again feed in only the previous word instead of the complete target sequences to decoder modules.
I also changed the implemention so that we create bias tensors for masking attentions only once, and not for every layer using SequenceMask.

The cache tensors are also fed into and out of the decoder module as a single, concatenated tensor which turned out to be more efficient.

Also changed finished hyps logic in beam search a little bit so that finished hypotheses can never become 'unfinished'.

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
would appreciate a review, @tdomhan @mjdenkowski

sockeye/decoder.py
```diff
@@ -249,62 +236,98 @@ def _decode(self,
...
-        :param target_embed_lengths: Lengths of embedded target sequences. Shape: (batch_size,).
-        :param target_embed_max_length: Size of embedded target sequence dimension.
+        :param step: Global step of inference procedure.
         :param target_embed_prev: Previous target word embedding. Shape: (batch_size, target_num_embed).
         :param source_encoded_max_length: Length of encoded source time dimension.
         :param states: Arbitrary list of decoder states.
         :return: logit inputs, attention probabilities, next decoder states.
         """
-        source_encoded, source_encoded_lengths = states
+        source_encoded, source_encoded_lengths = states[:2]  # pylint: disable=unbalanced-tuple-unpacking
```
mjdenkowski: Perhaps a comment reminding us what `states` should contain since we're confidently taking the first two elements?

sockeye/decoder.py
```diff
@@ -249,62 +236,98 @@ def _decode(self,
...
+
+        # auto-regressive bias for last position in sequence
+        # (1, target_max_length, target_max_length)
+        target_bias = transformer.get_autoregressive_bias(step, name="%sbias" % self.prefix)
+        target_bias = mx.sym.slice_axis(target_bias, axis=1, begin=-1, end=step)
 
-        # (batch_size, target_max_length, model_size)
-        target = self._decode(source_encoded, source_encoded_lengths, source_encoded_max_length,
-                              target_embed, target_embed_max_length)
+        layer_caches = self._get_layer_caches_from_states(list(states))
```
mjdenkowski: Perhaps a quick note here reminding us what we're retrieving from the cache (even though it's documented in `_get_layer_caches_from_states()`?

sockeye/decoder.py
```diff
@@ -249,62 +236,98 @@ def _decode(self,
...
+        cache = []  # type: List[mx.sym.Symbol]
+        for layer, layer_cache in zip(self.layers, layer_caches):
+            target = layer(target=target,
+                           target_bias=target_bias,
+                           source=source_encoded,
+                           source_bias=source_bias,
+                           cache=layer_cache)
+            cache += [layer_cache['k'], layer_cache['v']]
+        cache = mx.sym.concat(*cache, dim=1, name='new_cache')
```
mjdenkowski: A note reminding us what we're adding to the cache (conceptually) would be helpful too.

mjdenkowski:
Very nice improvement!  I looked though and made a couple of notes about documentation.  Has this been tested on a WMT system for speed/BLEU?

sockeye/decoder.py
fhieber: done

sockeye/decoder.py
fhieber: done

sockeye/decoder.py
fhieber: done

fhieber:
Yes, BLEU remains the same as nothing changes conceptually.
Speed-wise there are some pretty big benefits on CPUs. For GPUs this is less visible, because the constant recomputation in current master can be handled quite well by a P100.

sockeye/inference.py
```diff
@@ -1085,7 +1063,8 @@ def _beam_search(self,
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
             # (6) determine which hypotheses in the beam are now finished
-            finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+            finished += ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+            finished = mx.nd.clip(finished, a_min=0, a_max=1)
```
mjpost: Is the clipping actually necessary? I think any positive value will evaluate to true.

fhieber: It is because we are using finished above to update lengths. So it should only contain values of 0 and 1

CHANGELOG.md
```diff
@@ -10,12 +10,16 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.15.0]
```
tdomhan: why the version bump? old models should still be compatible, just faster, no?

sockeye/layers.py
```diff
@@ -423,19 +445,17 @@ def __init__(self,
...
         Returns a symbol of shape (batch, max_length, output_depth).
 
-        :param queries: Symbol of shape (batch, queries_max_length, input_depth). TODO?
-        :param queries_max_length: Size of queries time dimension.
+        :param queries: Symbol of shape (batch, query_max_length, input_depth).
         :param memory: Symbol of shape (batch, memory_max_length, input_depth).
-        :param memory_lengths: Symbol of shape (batch, 1).
-        :param memory_max_length: Size of memory time dimension.
-        :return: Symbol of shape (batch, queries_max_length, output_depth).
+        :param bias: 3d bias. TODO Optional
```
tdomhan: what is this "TODO Optional"?

CHANGELOG.md
fhieber: I did it because the signature of `--bucket-width` for translate.py has changed (it now only accepts a single number). 
Old models are still compatible though.
What do you think?

sockeye/layers.py
fhieber: A TODO I apparently overlooked :) will fix

CHANGELOG.md
tdomhan: I personally would like to avoid forcing people to having to retrain as much as possible and would like the major version at 1.14, but I see the point regarding the CLI change.

sockeye/decoder.py
```diff
@@ -81,20 +79,17 @@ def decode_sequence(self,
...
-        Decodes a single time step given the embedded target sequence and previous decoder states.
+        Decodes a single time step given the current step, the previous embedded target word,
+        and previous decoder states.
         Returns decoder representation for the next prediction, attention probabilities, and next decoder states.
         Implementations can maintain an arbitrary number of states.
 
-        :param target_embed: Embedded target sequence. Shape: (batch_size, target_embed_max_length, target_num_embed).
-        :param target_embed_lengths: Lengths of embedded target sequences. Shape: (batch_size,).
-        :param target_embed_max_length: Size of embedded target sequence dimension.
+        :param step: Global step of inference procedure.
```
tdomhan: could you add a note whether it's 0 or 1 based?

sockeye/decoder.py
```diff
@@ -1021,7 +1060,8 @@ def decode_step(self,
         new_layer_states = []
 
         # (batch_size, num_embed)
-        target_embed_prev = self.pos_embedding.encode_positions(indices, target_embed_prev)
+        symbolic_step = mx.sym.arange(start=step - 1, stop=step, step=1, name='symbolic_step')
```
tdomhan: minor:
so symbolic step is actually step minus one, no? Would it make sense to call the result `indices` as before?
```
indices = mx.sym.arange(start=step - 1, stop=step, step=1, name='indices')
```

sockeye/inference.py
tdomhan: we still guarantee that models can only produce the PAD symbol once finished. So why has it become necessary to do `+=` and `mx.nd.clip`?

sockeye/decoder.py
fhieber: I can :)

sockeye/decoder.py
fhieber: yes indeed, will change

fhieber:
I removed the clip it is not necessary as you said.

PR #227, opened 2017-11-29
Regularly run system tests.

tdomhan:
- Run one system test per model for each commit.
 - Run all system tests as daily cron jobs.
 - Factored out logging and optionally disable logging in run_train_translate.


tdomhan:
This just adds ~2min to the travis runs (9min now vs 7min before).

tdomhan:
sounds good. I'm also against bumping twice.

tdomhan:
ahh, actually tests before ran in 2minutes. I still think this is acceptable though.

PR #229, opened 2017-11-30
Swish-1 and GELU activation for transformers

mjdenkowski:
This adds support for two activation types to transformers:
- Swish-1 ([Ramachandran et al. 2017: Searching for Activation Functions](https://arxiv.org/pdf/1710.05941.pdf))
- GELU ([Hendrycks and Gimpel 2016: Bridging Nonlinearities and Stochastic Regularizers with Gaussian Error Linear Units](https://arxiv.org/pdf/1606.08415.pdf))

The first paper reports that Swish-1 beats ReLU on transformer models.  The second reports that GELU beats Swish-1, but not explicitly tested on transformers.

Edit: If we want to avoid a bunch of sequential major version bumps, this can wait until #226 is ready to go and we can have a single bump for all transformer updates.

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
+1 on syncing with #226 and #222 to not end up with Sockeye 1.inf :)

sockeye/layers.py
```diff
@@ -23,6 +24,32 @@
...
+      al. (https://arxiv.org/pdf/1710.05941.pdf), Elfwing et al.
+      (https://arxiv.org/pdf/1702.03118.pdf)
+    - Gaussian Error Linear Unit (GELU): Hendrycks and Gimpel
+      (https://arxiv.org/pdf/1606.08415.pdf)
+
+    :param data: input Symbol of any shape.
+    :param act_type: Type of activation.
+    :return: output Symbol with same shape as input.
+    """
+    # TODO: Contribute these to MXNet?  For now it appears that registered activation types must be implemented in C++.
```
fhieber: Have you observed increased memory or speed decreases with the GELU?
You could think about implementing these as custom operators to avoid creating multiple nodes in the computation graph if these turn out to work better than the Relu.
At least for swish, the backward implementation is easy, but the savings will be negligible there.

mjdenkowski: Compared to ReLU, decoding with Swish-1 takes 100.5% time and GELU takes 103.5% time.  Swish-1 uses the same amount of memory and GELU uses 110% memory.  Swish-1 is basically free while GELU is more expensive, but so far shows the best perplexity.  If they work consistently better than ReLU, we should definitely look at performance optimization.  I think this base implementation is fine for experiments.

fhieber: sounds good, we can revisit this if and when it becomes a bottleneck.

mjdenkowski:
This should be up to date and ready to merge.

PR #230, opened 2017-11-30
reorganized beam search for pruning

mjpost:
- each sentence now has an active beam size, which generalizes the former special case for t=1 and allows pruning
- length normalization is applied only to completed hypotheses
- `--beam-prune` specifies a width that will prune all hypotheses outside the beam set by the best completed hyp's value

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -842,6 +842,11 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--beam-prune', '-p',
+                               type=float,
+                               default=0,
+                               help='Pruning threshold for beam search. All hypotheses with unnormalized scores less than '
+                               'this amount below the best (unnormalized) hypothesis are discarded (0 = off).')
     decode_params.add_argument('--batch-size',
                                type=int_greater_or_equal(1),
                                default=1,
--- a/sockeye/checkpoint_decoder.py
+++ b/sockeye/checkpoint_decoder.py
@@ -129,6 +129,7 @@ class CheckpointDecoder:
                                           self.bucket_width_source,
                                           self.bucket_width_target,
                                           inference.LengthPenalty(self.length_penalty_alpha, self.length_penalty_beta),
+                                          0.0,
                                           models,
                                           vocab_source,
                                           vocab_target)
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -625,6 +625,7 @@ class Translator:
     :param context: MXNet context to bind modules to.
     :param ensemble_mode: Ensemble mode: linear or log_linear combination.
     :param length_penalty: Length penalty instance.
+    :param beam_prune: Beam pruning difference threshold.
     :param models: List of models.
     :param vocab_source: Source vocabulary.
     :param vocab_target: Target vocabulary.
@@ -637,12 +638,14 @@ class Translator:
                  bucket_source_width: int,
                  bucket_target_width: int,
                  length_penalty: LengthPenalty,
+                 beam_prune: float,
                  models: List[InferenceModel],
                  vocab_source: Dict[str, int],
                  vocab_target: Dict[str, int],
                  restrict_lexicon: Optional[lexicon.TopKLexicon] = None) -> None:
         self.context = context
         self.length_penalty = length_penalty
+        self.beam_prune = beam_prune
         self.vocab_source = vocab_source
         self.vocab_target = vocab_target
         self.vocab_target_inv = vocab.reverse_vocab(self.vocab_target)
@@ -995,8 +998,9 @@ class Translator:
         # (0) encode source sentence, returns a list
         model_states = self._encode(source, source_length)
 
+        # Records the size of the active beam for each sentence (which permits filtering)
+        active_beam_size = [1] * self.batch_size
         for t in range(1, max_output_length):
-
             # (1) obtain next predictions and advance models' state
             # scores: (batch_size * beam_size, target_vocab_size)
             # attention_scores: (batch_size * beam_size, bucket_key)
@@ -1009,29 +1013,25 @@ class Translator:
                                                                        models_output_layer_w,
                                                                        models_output_layer_b)
 
-            # (2) compute length-normalized accumulated scores in place
-            if t == 1 and self.batch_size == 1:  # only one hypothesis at t==1
-                scores = scores[:1] / self.length_penalty(lengths[:1])
-            else:
-                # renormalize scores by length ...
-                scores = (scores + scores_accumulated * self.length_penalty(lengths - 1)) / self.length_penalty(lengths)
-                # ... but not for finished hyps.
-                # their predicted distribution is set to their accumulated scores at C.PAD_ID.
+            # (2) Replace finished rows with inf in all columns except column 0, which has the
+            # accumulated model score
+            if t > 1:
+                scores = scores + scores_accumulated
                 pad_dist[:, C.PAD_ID] = scores_accumulated
-                # this is equivalent to doing this in numpy:
-                #   pad_dist[finished, :] = np.inf
-                #   pad_dist[finished, C.PAD_ID] = scores_accumulated[finished]
                 scores = mx.nd.where(finished, pad_dist, scores)
 
-            # (3) get beam_size winning hypotheses for each sentence block separately
-            # TODO(fhieber): once mx.nd.topk is sped-up no numpy conversion necessary anymore.
+            # (3) Get beam_size winning hypotheses for each sentence block separately. Only look as
+            # far as the active beam size for each sentence. TODO(fhieber): once mx.nd.topk is
+            # sped-up no numpy conversion necessary anymore.
             scores = scores.asnumpy()  # convert to numpy once to minimize cross-device copying
-            for sent in range(self.batch_size):
-                rows = slice(sent * self.beam_size, (sent + 1) * self.beam_size)
-                sliced_scores = scores if t == 1 and self.batch_size == 1 else scores[rows]
+            for sentno in range(self.batch_size):
+                rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+                active_rows = slice(sentno * self.beam_size, sentno * self.beam_size + active_beam_size[sentno])
+                sliced_scores = scores[active_rows]
                 # TODO we could save some tiny amount of time here by not running smallest_k for a finished sent
                 (best_hyp_indices_np[rows], best_word_indices_np[rows]), \
-                    scores_accumulated_np[rows] = utils.smallest_k(sliced_scores, self.beam_size, t == 1)
+                    scores_accumulated_np[rows] = utils.smallest_k(sliced_scores, self.beam_size)
+
                 # offsetting since the returned smallest_k() indices were slice-relative
                 best_hyp_indices_np[rows] += rows.start
 
@@ -1043,27 +1043,66 @@ class Translator:
             if self.restrict_lexicon:
                 best_word_indices[:] = vocab_slice_ids.take(best_word_indices)
 
-            # (4) get hypotheses and their properties for beam_size winning hypotheses (ascending)
+            # (4) Normalize the scores of newly finished hypotheses
+            newly_finished = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]).expand_dims(axis=1)
+            scores_accumulated = mx.nd.where(newly_finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
+
+                # Set the active beam size to full. It will get reduced if items are filtered.
+                active_beam_size[sentno] = self.beam_size
+
+                rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+                if self.beam_prune > 0.0:
+                    ones_array = mx.nd.ones((self.beam_size,), ctx=self.context, dtype='int32')
+                    zeros_array = mx.nd.zeros((self.beam_size,), ctx=self.context, dtype='int32')
+                    inf_array = mx.nd.full((self.beam_size,1), val=np.inf, ctx=self.context, dtype='float32')
+                    if mx.nd.sum(finished[rows]) > 0:
+                        best_score = mx.nd.min(mx.nd.where(finished[rows].expand_dims(axis=1), scores_accumulated[rows], inf_array))
+
+                        over_thresh = mx.nd.cast(scores_accumulated[rows,0] - best_score > self.beam_prune, dtype='int32')
+                        to_remove = over_thresh
+                        scores_accumulated[rows] = mx.nd.where(to_remove, inf_array, scores_accumulated[rows])
+                        best_word_indices[rows] = mx.nd.where(to_remove, zeros_array, best_word_indices[rows])
+
+                        num_removing = int(mx.nd.sum(to_remove).asscalar())
+                        active_beam_size[sentno] = self.beam_size - num_removing
+
+                        # mark removed ones as finished so they won't block early exiting
+                        finished[rows] = finished[rows] + to_remove
+
+                # Resort the hypotheses
+                sorted_indices = mx.nd.argsort(scores_accumulated[rows,0])
+                best_hyp_indices[rows] = mx.nd.take(best_hyp_indices[rows], sorted_indices)
+                best_word_indices[rows] = mx.nd.take(best_word_indices[rows], sorted_indices)
+                scores_accumulated[rows] = mx.nd.take(scores_accumulated[rows,0], sorted_indices.expand_dims(axis=1))
+
+            # (6) get hypotheses and their properties for beam_size winning hypotheses (ascending)
             sequences = mx.nd.take(sequences, best_hyp_indices)
             lengths = mx.nd.take(lengths, best_hyp_indices)
             finished = mx.nd.take(finished, best_hyp_indices)
             attention_scores = mx.nd.take(attention_scores, best_hyp_indices)
             attentions = mx.nd.take(attentions, best_hyp_indices)
 
-            # (5) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
+            # (7) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
             # pylint: disable=unsupported-assignment-operation
             sequences[:, t] = best_word_indices
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
-            # (6) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+
+            self.print_beam(sequences, scores_accumulated, finished, active_beam_size, t)
+
             if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
                 break
 
-            # (7) update models' state with winning hypotheses (ascending)
+            # (8) update models' state with winning hypotheses (ascending)
             for ms in model_states:
                 ms.sort_state(best_hyp_indices)
+
+        logger.info("Finished after %d / %d steps.", t, max_output_length)
 
         return sequences, attentions, scores_accumulated, lengths
 
@@ -1096,3 +1135,16 @@ class Translator:
             score = accumulated_scores[idx].asscalar()
             result.append(Translation(sequence, attention_matrix, score))
         return result
+
+    def print_beam(self, sequences, scores, finished, active, timestep):
+        print('BEAM AT TIME STEP', timestep)
+        for sentno in range(self.batch_size):
+            idx = sentno * self.beam_size
+            for i in range(self.beam_size):
+                word_ids = [int(x.asscalar()) for x in sequences[idx + i]]
+                hypothesis = ' '.join([self.vocab_target_inv[x] for x in word_ids if x != 0])
+                score = scores[idx + i].asscalar()
+                if i < active[sentno]:
+                    print(i, finished[idx + i].asscalar(), score, hypothesis)
+                else:
+                    print(i, finished[idx + i].asscalar(), score, '----------')
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -214,8 +214,7 @@ class Accuracy(mx.metric.EvalMetric):
             self.num_inst += n
 
 
-def smallest_k(matrix: np.ndarray, k: int,
-               only_first_row: bool = False) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
+def smallest_k(matrix: np.ndarray, k: int) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
     """
     Find the smallest elements in a numpy matrix.
 
@@ -224,10 +223,7 @@ def smallest_k(matrix: np.ndarray, k: int,
     :param only_first_row: If true the search is constrained to the first row of the matrix.
     :return: The row indices, column indices and values of the k smallest items in matrix.
     """
-    if only_first_row:
-        flatten = matrix[:1, :].flatten()
-    else:
-        flatten = matrix.flatten()
+    flatten = matrix.flatten()
 
     # args are the indices in flatten of the k smallest elements
     args = np.argpartition(flatten, k)[:k]
--- a/sockeye/translate.py
+++ b/sockeye/translate.py
@@ -84,6 +84,7 @@ def main():
                                                   bucket_target_width,
                                                   sockeye.inference.LengthPenalty(args.length_penalty_alpha,
                                                                                   args.length_penalty_beta),
+                                                  args.beam_prune,
                                                   models,
                                                   vocab_source,
                                                   vocab_target,
--- a/test/unit/test_arguments.py
+++ b/test/unit/test_arguments.py
@@ -184,6 +184,7 @@ def test_training_arg(test_params, expected_params):
                       checkpoints=None,
                       models=['model'],
                       beam_size=5,
+                      beam_prune=0,
                       batch_size=1,
                       chunk_size=1,
                       ensemble_mode='linear',


sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
...
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
+
+                # Set the active beam size to full. It will get reduced if items are filtered.
+                active_beam_size[sentno] = self.beam_size
+
+                rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+                if self.beam_prune > 0.0:
+                    ones_array = mx.nd.ones((self.beam_size,), ctx=self.context, dtype='int32')
```
fhieber: these are re-created every single time. Why not move them outside of the whole beam search loop?

sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
...
+                    zeros_array = mx.nd.zeros((self.beam_size,), ctx=self.context, dtype='int32')
+                    inf_array = mx.nd.full((self.beam_size,1), val=np.inf, ctx=self.context, dtype='float32')
+                    if mx.nd.sum(finished[rows]) > 0:
+                        best_score = mx.nd.min(mx.nd.where(finished[rows].expand_dims(axis=1), scores_accumulated[rows], inf_array))
+
+                        over_thresh = mx.nd.cast(scores_accumulated[rows,0] - best_score > self.beam_prune, dtype='int32')
+                        to_remove = over_thresh
+                        scores_accumulated[rows] = mx.nd.where(to_remove, inf_array, scores_accumulated[rows])
```
fhieber: can you check whether this can be done easier with latest MXNet's advanced indexing?

sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
...
+                rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+                if self.beam_prune > 0.0:
+                    ones_array = mx.nd.ones((self.beam_size,), ctx=self.context, dtype='int32')
+                    zeros_array = mx.nd.zeros((self.beam_size,), ctx=self.context, dtype='int32')
+                    inf_array = mx.nd.full((self.beam_size,1), val=np.inf, ctx=self.context, dtype='float32')
+                    if mx.nd.sum(finished[rows]) > 0:
+                        best_score = mx.nd.min(mx.nd.where(finished[rows].expand_dims(axis=1), scores_accumulated[rows], inf_array))
+
+                        over_thresh = mx.nd.cast(scores_accumulated[rows,0] - best_score > self.beam_prune, dtype='int32')
+                        to_remove = over_thresh
```
fhieber: why not direectly assign to 'to_remove' ?

sockeye/inference.py
```diff
@@ -1009,29 +1013,25 @@ def _beam_search(self,
                                                                        models_output_layer_w,
                                                                        models_output_layer_b)
 
-            # (2) compute length-normalized accumulated scores in place
-            if t == 1 and self.batch_size == 1:  # only one hypothesis at t==1
-                scores = scores[:1] / self.length_penalty(lengths[:1])
-            else:
-                # renormalize scores by length ...
-                scores = (scores + scores_accumulated * self.length_penalty(lengths - 1)) / self.length_penalty(lengths)
```
fhieber: why does this change not affect the search behavior? If you don't normalize hypotheses by length, the results of smallest_k will be different, or am I missing something?

sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
             if self.restrict_lexicon:
                 best_word_indices[:] = vocab_slice_ids.take(best_word_indices)
 
-            # (4) get hypotheses and their properties for beam_size winning hypotheses (ascending)
+            # (4) Normalize the scores of newly finished hypotheses
+            newly_finished = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]).expand_dims(axis=1)
+            scores_accumulated = mx.nd.where(newly_finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
```
fhieber: how much runtime does this new feature add? How is speed affected if you don't prune (the baseline setting)?

sockeye/inference.py
```diff
@@ -1009,29 +1013,25 @@ def _beam_search(self,
...
-                # renormalize scores by length ...
-                scores = (scores + scores_accumulated * self.length_penalty(lengths - 1)) / self.length_penalty(lengths)
-                # ... but not for finished hyps.
-                # their predicted distribution is set to their accumulated scores at C.PAD_ID.
+            # (2) Replace finished rows with inf in all columns except column 0, which has the
+            # accumulated model score
+            if t > 1:
+                scores = scores + scores_accumulated
                 pad_dist[:, C.PAD_ID] = scores_accumulated
-                # this is equivalent to doing this in numpy:
```
fhieber: why remove this comment? I find it useful :)

sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
...
+            # (7) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
             # pylint: disable=unsupported-assignment-operation
             sequences[:, t] = best_word_indices
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
-            # (6) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+
+            self.print_beam(sequences, scores_accumulated, finished, active_beam_size, t)
```
fhieber: this probably should go away.
I guess you also don't need active_beam_size then anymore

sockeye/inference.py
```diff
@@ -1043,28 +1043,67 @@ def _beam_search(self,
...
+
             if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
                 break
 
-            # (7) update models' state with winning hypotheses (ascending)
+            # (8) update models' state with winning hypotheses (ascending)
             for ms in model_states:
                 ms.sort_state(best_hyp_indices)
 
+        logger.info("Finished after %d / %d steps.", t, max_output_length)
```
fhieber: should either be debug or removed

sockeye/utils.py
```diff
@@ -214,8 +214,7 @@ def update(self, labels, preds):
             self.num_inst += n
 
 
-def smallest_k(matrix: np.ndarray, k: int,
-               only_first_row: bool = False) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
+def smallest_k(matrix: np.ndarray, k: int) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
```
fhieber: can you also adapt smallest_k_mx to make them  equivalent?

sockeye/checkpoint_decoder.py
```diff
@@ -129,6 +129,7 @@ def decode_and_evaluate(self,
                                           self.bucket_width_source,
                                           self.bucket_width_target,
                                           inference.LengthPenalty(self.length_penalty_alpha, self.length_penalty_beta),
+                                          0.0,
```
fhieber: maybe we should use keywords in this constructor call to clarify what all of these mean.

sockeye/inference.py
mjpost: It is the same until you have mixed finished / unfinished hypotheses. At that point, the finished ones have a much lower score and jump to the top. If you use pruning, they make quick work of the beam, throwing out most hypotheses.

sockeye/inference.py
mjpost: No prob.

sockeye/inference.py
mjpost: On GPUs it's the same or a little faster with no pruning; on CPUs it's maybe a 10–20% speedup. There's more work in some places (e.g., here), but you also save a lot in other places. Being able to prune the hypothesis list puts in place a more general framework where we could do more advanced filtering.

sockeye/arguments.py
```diff
@@ -842,6 +842,11 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--beam-prune', '-p',
+                               type=float,
+                               default=0,
+                               help='Pruning threshold for beam search. All hypotheses with unnormalized scores less than '
+                               'this amount below the best (unnormalized) hypothesis are discarded (0 = off).')
```
fhieber: Please add a "Default: %(default)s." string to the end of the help message.

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
             if self.restrict_lexicon:
                 best_word_indices[:] = vocab_slice_ids.take(best_word_indices)
 
-            # (4) get hypotheses and their properties for beam_size winning hypotheses (ascending)
+            # (4) Normalize the scores of newly finished hypotheses
+            newly_finished = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]).expand_dims(axis=1)
```
fhieber: shouldn't an hypothesis also be finished if it produces a pad symbol? at least this is the existing logic below.

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
+                        best_word_indices[rows] = mx.nd.where(to_remove, zeros_array, best_word_indices[rows])
+
+                        num_removing = int(mx.nd.sum(to_remove).asscalar())
+                        active_beam_size[sentno] = self.beam_size - num_removing
+
+                        # mark removed ones as finished so they won't block early exiting
+                        finished[rows] = finished[rows] + to_remove
+
+                # Resort the hypotheses
+                sorted_indices = mx.nd.argsort(scores_accumulated[rows,0])
```
fhieber: nitpick: space between: "rows, 0"

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
+                best_hyp_indices[rows] = mx.nd.take(best_hyp_indices[rows], sorted_indices)
+                best_word_indices[rows] = mx.nd.take(best_word_indices[rows], sorted_indices)
+                scores_accumulated[rows] = mx.nd.take(scores_accumulated[rows,0], sorted_indices.expand_dims(axis=1))
```
fhieber: nitpick: space between: "rows, 0"

sockeye/inference.py
fhieber: I don't fully understand this statement. Why are you sorting based on the first column? If no pruning is applied (the block starting in line 1061 is not entered), do we still need this and why?

sockeye/inference.py
fhieber: it would be useful to comment these lines, otherwise its hard to understand the logic behind these.

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
-            # (5) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
+            # (7) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
             # pylint: disable=unsupported-assignment-operation
             sequences[:, t] = best_word_indices
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
-            # (6) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+
```
fhieber: remove commented code

sockeye/inference.py
```diff
@@ -1096,3 +1133,16 @@ def _get_best_from_beam(self,
             score = accumulated_scores[idx].asscalar()
             result.append(Translation(sequence, attention_matrix, score))
         return result
+
+    def print_beam(self, sequences, scores, finished, active, timestep):
```
fhieber: I think this function should either be removed or changed to use the logger. We never want to print something, imo.

sockeye/inference.py
fhieber: why do you need expand_dims here?

sockeye/inference.py
fhieber: scores_accumulated is `(batch*beam,)`; why do you need the second dimension here?

sockeye/inference.py
fhieber: same question about expand_dims here.
Would it help to initialize scores_accumulated as `(batch * beam, 1)` to avoid some expand_dims?

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
+
+                # Set the active beam size to full. It will get reduced if items are filtered.
+                active_beam_size[sentno] = self.beam_size
+
+                rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+                if self.beam_prune > 0.0:
+                    if mx.nd.sum(finished[rows]) > 0:
+                        best_score = mx.nd.min(mx.nd.where(finished[rows].expand_dims(axis=1), scores_accumulated[rows], inf_array))
```
fhieber: why not use newly_finished? Wouldn't you skip the hyps that were just finished here?

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
+
+                        to_remove = mx.nd.cast(scores_accumulated[rows,0] - best_score > self.beam_prune, dtype='int32')
+                        scores_accumulated[rows] = mx.nd.where(to_remove, inf_array, scores_accumulated[rows])
+                        best_word_indices[rows] = mx.nd.where(to_remove, zeros_array, best_word_indices[rows])
```
fhieber: why do you need this? You are using `sorted_indices` below anyway ?

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
+
+                        num_removing = int(mx.nd.sum(to_remove).asscalar())
+                        active_beam_size[sentno] = self.beam_size - num_removing
+
+                        # mark removed ones as finished so they won't block early exiting
+                        finished[rows] = finished[rows] + to_remove
```
fhieber: so the values in this array can now be > 1? This will cause problems as we compute lengths based on finished.

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
 
-            # (5) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
+            # (7) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
             # pylint: disable=unsupported-assignment-operation
             sequences[:, t] = best_word_indices
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
-            # (6) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
```
fhieber: Does this logic still hold with your change? (you created a new finished array intermediately that may differ).

sockeye/inference.py
```diff
@@ -1043,25 +1047,58 @@ def _beam_search(self,
...
-            # (4) get hypotheses and their properties for beam_size winning hypotheses (ascending)
+            # (4) Normalize the scores of newly finished hypotheses
+            newly_finished = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]).expand_dims(axis=1)
+            scores_accumulated = mx.nd.where(newly_finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
+
+                # Set the active beam size to full. It will get reduced if items are filtered.
+                active_beam_size[sentno] = self.beam_size
```
tdomhan: how does this work exactly? When hypothesis are removed because they're finished the active_beam_size will be reduced for the next step, but then be reset to beam_size for the step after that. Is that right?

sockeye/arguments.py
```diff
@@ -842,6 +842,11 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--beam-prune', '-p',
+                               type=float,
+                               default=0,
+                               help='Pruning threshold for beam search. All hypotheses with unnormalized scores less than '
```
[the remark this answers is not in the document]
mjpost: This is an absolute score difference between each hypothesis and the highest-score completed hypothesis. So yes, in log space.

sockeye/inference.py
mjpost: fixed

sockeye/utils.py
mjpost: done

sockeye/arguments.py
mjpost: done

sockeye/inference.py
mjpost: These are ones that have just been completed on this timestep.

sockeye/inference.py
mjpost: done

mjpost: Above the pruning, scores_accumulated applies pruning only to some (newly finished) hypotheses, so we have to re-sort.

sockeye/inference.py
mjpost: done

mjpost: It either has to be reshaped here, or when compared to newly_finished (if I remember correctly).

sockeye/inference.py
mjpost: Any newly finished hypotheses may or may not be the best among all of them. And old hypotheses have grown by a token, and may have therefore fallen out of the beam, regardless of whether any of them finished this step.

sockeye/inference.py
mjpost: I think it was necessary to prevent the next step of the decoder from choosing valid words and giving them good probabilities. This way their scores will be very high, and topk() will not select anything from such rows.

sockeye/inference.py
mjpost: I suppose this is a possibility, though pruning usually just removes unfinished hypotheses (usually, with pruning, once one hyp. completes, things come to an end within a step or two). But I suppose I should cap it at 1.

sockeye/inference.py
mjpost: active_beam_size is initialized to 1 to handle the t=1 time step (this allows removing the first-step logic from before). Here, we set it to beam_size. If pruning occurs, it is set again to something smaller. This is the same as having an if...else block on pruning.

mjpost:
Here is  a performance graph for two GPUs. The restructuring seems to be quite a bit faster even without pruning. With a small amount of pruning, a slight gain is seen. BLEU scores are the same between master and prune=0, and vary a little bit when prune=1.

[beam_search.pdf](https://github.com/awslabs/sockeye/files/1815624/beam_search.pdf)

sockeye/inference.py
fhieber: can you elaborate? Not sure I understand your reply.

mjpost: This gets only the items for which EOS was generated on the current timestep (not ones that were already finished on previous time steps), so that normalization can be applied. The pad symbol can only be produced after EOS is generated, right?

fhieber: no, why would that be the case? A model can generate any token in the target vocabulary, including PAD.

tdomhan:
thanks for the extensive speed testing! So for everything except for greedy it's faster.

mjpost:
@tdomhan except it looks like I changed two variables (the prune= plots (blue and purple)) are using the mxnet topk() call, which accounts for most of the change. I'll update soon.

mjpost:
The latest push moves post-pruning beam sorting to outside the timestep loop. This restores runtime parity to the current master branch, so that the pruning functionality does not impose a runtime cost. I'll post plots tomorrow when the experiments finish.

mjpost:
Here is a new plot where I have not mixed up any variables. This is over a single timing run, where timing runs often vary by as much as +- 0.1 second/sentence. Conclusions:

- the reorganization doesn't impose any costs when doing no pruning
- pruning with a threshold of 1 can provide a slight decrease in runtime at no cost in BLEU score
- For GPU inference, we should move to MXNet's topk() (also relevant to issue #270). However, for CP inference, MXNet's topk() is much slower. I'll add a plot soon.

beam_search

sockeye/arguments.py
```diff
@@ -992,6 +992,15 @@ def add_inference_args(params):
                                type=int_greater_or_equal(1),
                                default=5,
                                help='Size of the beam. Default: %(default)s.')
+    decode_params.add_argument('--beam-prune', '-p',
+                               type=float,
+                               default=0,
+                               help='Pruning threshold for beam search. All hypotheses with scores not within '
+                               'this amount of the best finished hypothesis are discarded (0 = off). Default: %(default)s')
+    decode_params.add_argument('--beam-search-stop',
+                               choices='all first'.split(),
```
tdomhan: why not give a list?

mjpost:
Merged in the latest master and modified the code to work with @Deseaus's batched top-k.

mjpost:
Here is an up-to-date high-level description of the code:

- It adds two new command-line options enabling pruning and altering the stopping criteria (this latter point isn't as important, and we could remove it)
- It introduces a general ability to inactivate items on the beam. This also allows getting rid of the special treatment for t==1, since this can be handled as a special case of invalidation. In general some of the code at the top of the loop over t is simplified. The pruning code adds some complexity but it is isolated into a single loop that is only used when pruning is enabled.
- Only completed hypotheses are normalized. Pruning is applied in log-prob space against the best-completed hypothesis. By default there is no pruning.

sockeye/inference.py
```diff
@@ -1271,43 +1284,66 @@ def _beam_search(self,
...
+            # (4) Normalize the scores of newly finished hypotheses
+            all_finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+            newly_finished = all_finished - finished
+            scores_accumulated = mx.nd.where(newly_finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
+
+                # All rows are now active (after special treatment of start state at t=1)
+                inactive[:] = 0
```
fhieber: why is this done in the loop? If beam prune is off this can just be done once.

sockeye/inference.py
```diff
@@ -1326,13 +1362,32 @@ def _beam_search(self,
 
             # (7) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
-            if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
-                break
+
+#            self._print_beam(sequences, scores_accumulated, finished, inactive, t)
```
fhieber: please remove

sockeye/inference.py
```diff
@@ -1369,3 +1424,30 @@ def _get_best_from_beam(self,
             else:
                 result.append(Translation(sequence, attention_matrix, score))
         return result
+
+    def _print_beam(self,
```
fhieber: this function should be rewritten to use log.debug statements or be completely removed.

sockeye/inference.py
```diff
@@ -1326,13 +1362,32 @@ def _beam_search(self,
 
             # (7) determine which hypotheses in the beam are now finished
             finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
-            if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
-                break
+
+#            self._print_beam(sequences, scores_accumulated, finished, inactive, t)
+
+            if self.beam_search_stop == 'first' and self.batch_size == 1:
```
fhieber: 'first' and 'all' should be made global constants  in constants.py

sockeye/inference.py
```diff
@@ -1326,13 +1362,32 @@ def _beam_search(self,
...
+                # TODO: doesn't work for batching
```
fhieber: what doesn't work for batching? 'first' top and batch_size > 1? But then this line is unreachable based on 1368, right?

sockeye/inference.py
```diff
@@ -1326,13 +1362,32 @@ def _beam_search(self,
...
+            else:
+                if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
+                    break
 
             # (8) update models' state with winning hypotheses (ascending)
             for ms in model_states:
                 ms.sort_state(best_hyp_indices)
 
+        # (9) Sort the hypotheses within each sentence (normalization for finished hyps breaks the sort)
+        for sentno in range(self.batch_size):
```
fhieber: this throws us back into the for loop logic before the batched topk change. Can we avoid this?

sockeye/inference.py
```diff
@@ -1326,13 +1362,32 @@ def _beam_search(self,
...
+            rows = slice(sentno * self.beam_size, (sentno + 1) * self.beam_size)
+            best_hyp_indices[rows] = rows.start + mx.nd.argsort(scores_accumulated[rows, 0])
+
+        # Make corresponding updates after having sorted
+        sequences = mx.nd.take(sequences, best_hyp_indices)
```
fhieber: these takes are not cheap. Having to do them twice in every search iteration is costly. Is there any way to avoid this?

mjpost: They're only done once per iteration. The final one here takes place after the loop has exited. I can change it to use a top-level operation instead of iterating over the batch (good idea).

sockeye/inference.py
mjpost: Yes, this can be fixed.

sockeye/inference.py
mjpost: For batching, we need to check if there is a sentence completed in every sentence in the batch. The whole concept of early stopping is a little funny with batch processing, so I didn't bother doing this check. We could remove early stopping if you'd like (though I don't think it hurts).

sockeye/inference.py
mjpost: Yes, a mistake.

sockeye/inference.py
```diff
@@ -1271,43 +1284,66 @@ def _beam_search(self,
...
                 best_word_indices[:] = vocab_slice_ids.take(best_word_indices)
 
-            # (4) get hypotheses and their properties for beam_size winning hypotheses (ascending)
+            # (4) Normalize the scores of newly finished hypotheses
+            all_finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+            newly_finished = all_finished - finished
+            scores_accumulated = mx.nd.where(newly_finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
+
+            # (5) Prune out low-probability hypotheses
+            for sentno in range(self.batch_size):
```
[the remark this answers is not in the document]
mjpost: Done. With pruning off, there is now no loop and no overhead.

sockeye/inference.py
mjpost: On the subject of takes, though — if they are indeed expensive, we might be able to get some gain by packing all of these (beam * batch, 1) tracking tensors (scores_accumulated, finished, lengths) into a single array (perhaps indexed with informative constants or as slices), in which case we'd only need a single `take()`. This would apply in the old code as well.

fhieber: That could be an idea, but let's do that in a separate commit.
Also, thanks for clarifying that this is outside of the main loop. I read the indentation wrong. In this case this should be fine I think.

sockeye/arguments.py
```diff
@@ -999,11 +999,11 @@ def add_inference_args(params):
...
                                help='Pruning threshold for beam search. All hypotheses with scores not within '
-                               'this amount of the best finished hypothesis are discarded (0 = off). Default: %(default)s')
+                               'this amount of the best finished hypothesis are discarded (0 = off). Default: %(default)s.')
     decode_params.add_argument('--beam-search-stop',
-                               choices=['all', 'first'],
-                               default='all',
-                               help='Stopping criteria. Quit when (all) hypotheses are finished (default) or when a finished hypothesis is in (first) position')
+                               choices=[C.BEAM_SEARCH_STOP_ALL, C.BEAM_SEARCH_STOP_FIRST],
+                               default=C.BEAM_SEARCH_STOP_ALL,
+                               help='Stopping criteria. Quit when (all) hypotheses are finished (default) or when a finished hypothesis is in (first) position.')
```
fhieber: [nit] add "Default: %(default)s."

sockeye/inference.py
```diff
@@ -1176,6 +1182,45 @@ def _combine_predictions(self,
             neg_logprobs = self.interpolation_func(probs)
         return neg_logprobs, attention_prob_score
 
+    def _prune(self,
```
fhieber: thanks a lot for the refactor, this is much cleaner and easier to understand!

sockeye/inference.py
```diff
@@ -914,6 +921,12 @@ def __init__(self,
             self.buckets_source = [self.max_input_length]
         self.pad_dist = mx.nd.full((self.batch_size * self.beam_size, len(self.vocab_target)), val=np.inf,
                                    ctx=self.context)
+        # These are constants used for manipulation
```
fhieber: manipulation of 'beam/scores'?

sockeye/inference.py
```diff
@@ -1259,8 +1311,10 @@ def _beam_search(self,
         # (0) encode source sentence, returns a list
         model_states = self._encode(source, source_length)
 
+        # Records items in the beam that are inactive. At the beginning (t==1), there is only one valid or active
+        # item on the beam for each sentence
+        inactive = mx.nd.array(([0] + [1] * (self.beam_size - 1)) * self.batch_size, ctx=self.context, dtype='int32')
```
fhieber: I wonder if 
```
inactive = mx.nd.ones((self.batch_size * self.beam_size), dtype='int32', ctx=self.context)
inactive[::beam_size] = 0
```
wouldn't be easier. It would also reuse the common pattern of shape batch * beam, instead of beam * batch.
And you would avoid the python lists...

sockeye/inference.py
```diff
@@ -1271,43 +1325,50 @@ def _beam_search(self,
...
-                #   pad_dist[finished, :] = np.inf
-                #   pad_dist[finished, C.PAD_ID] = scores_accumulated[finished]
-                scores = mx.nd.where(finished, pad_dist, scores)
-
-            # (3) get beam_size winning hypotheses for each sentence block separately
-            best_hyp_indices[:], best_word_indices[:], scores_accumulated[:, 0] = topk(scores, t=t)
+            # (2) Special treatment for finished and inactive rows. Inactive rows are inf everywhere;
+            # finished rows are inf everywhere except column zero, which holds the accumulated model score
+            scores += scores_accumulated
+            pad_dist[:, C.PAD_ID] = mx.nd.where(inactive, self.inf_array_long[:, 0], scores_accumulated[:, 0])
```
fhieber: this is the only place where we use `self.inf_array_long` and you need to get the 0th column here. Would it make sense to have `self.inf_array_long.shape` to be `(self.batch_size * self.beam_size,)` instead of `(self.batch_size * self.beam_size, 1)`?

mjpost:
Okay, I pulled in a lot of your comments and simplified things a bit more, even. But next I want to add some test cases.

mjpost:
Yep, thanks to you, too! I'll work on tests next. I've already uncovered a bug :)

mjpost:
With this last push, I have verified that BLEU scores, model scores, and runtimes are the same between this branch and master for (arch: p2 p3) x (beam: 1 5 10 20 30) x (batch: 1 10) x (prune: 0).

sockeye/inference.py
```diff
@@ -1369,3 +1445,25 @@ def _get_best_from_beam(self,
...
+                    finished: mx.nd.NDArray,
+                    inactive: mx.nd.NDArray,
+                    timestep: int) -> None:
+        """
+        Prints the beam for debugging purposes.
+        :param sequences: The beam histories (shape: batch_size * beam_size, max_output_len).
+        :param accumulated_scores: The accumulated scores for each item in the beam (shape: batch_size * beam_size, target_vocab_size).
+        :param finished: Indicates which items are finished (shape: batch_size * beam_size).
+        :param inactive: Indicates any inactive items (shape: batch_size * beam_size).
+        :param timestep: The current timestep:
```
fhieber: [nit] ':' -> '.'

sockeye/inference.py
```diff
@@ -1369,3 +1445,25 @@ def _get_best_from_beam(self,
...
         return result
+
+    def _print_beam(self,
+                    sequences: mx.nd.NDArray,
+                    accumulated_scores: mx.nd.NDArray,
+                    finished: mx.nd.NDArray,
+                    inactive: mx.nd.NDArray,
+                    timestep: int) -> None:
+        """
+        Prints the beam for debugging purposes.
```
fhieber: [nit] missing newline.

sockeye/inference.py
```diff
@@ -1271,43 +1321,50 @@ def _beam_search(self,
...
-                # this is equivalent to doing this in numpy:
-                #   pad_dist[finished, :] = np.inf
-                #   pad_dist[finished, C.PAD_ID] = scores_accumulated[finished]
-                scores = mx.nd.where(finished, pad_dist, scores)
-
-            # (3) get beam_size winning hypotheses for each sentence block separately
-            best_hyp_indices[:], best_word_indices[:], scores_accumulated[:, 0] = topk(scores, t=t)
+            # (2) Special treatment for finished and inactive rows. Inactive rows are inf everywhere;
+            # finished rows are inf everywhere except column zero, which holds the accumulated model score
+            scores += scores_accumulated
```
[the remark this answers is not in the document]
mjpost: I think you're forgetting — `scores` is just distributions over the next word; we have to add in the accumulated probability to each distribution in each row so that `topk()` selects the best complete hypotheses.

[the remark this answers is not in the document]
fhieber: Ah right, sorry, got confused.

mjpost:
I added an integration test and it completes fine, but I haven't added a system test yet.

I think the kinds of unit tests that would be best here would be to add a k-best output option to the decoder, and then to manually inspect the beam for differences that result from applying different beam thresholds. During this last rebase, I would often find bugs that didn't show up until I decoded entire test sets and noticed a drop in corpus-level BLEU, so the existing tests are too coarse to be useful. What do you think of this? One option would be to use the existing JSON object returned by `--output-type beam_store`, but it might be nicer to have a k-best interface (this is also something that I've seen requested in a number of different places, and also in private conversations).

As a second, related proposal, I would suggest merging this code in now, but also creating an issue registering the need for tests based on inspecting the beam (to make sure it happens), and then adding that test in a separate commit. The reason I want to commit now is because I am worried about the beam search code changing in the meantime and having to rebase again (or having someone else start a project on the beam, and them having to rebase). I can commit to writing the test by the end of next week.

fhieber:
I agree that adding the possibility of returning k-best would be a nice feature, but its independent of this change, so we should do that independently.
As for tests, the integration test is fine for functional testing but seems a little bit redundant (there is a lot of code running then that is already checked by other integration tests, the only difference is that you invoke _prune_beam(). Would it be hard to write a single unit test just for this function and make the integration test using beam_prune > 0 a system test? (system tests are run nightly on master not with every commit)

I agree that we should merge this soon, and from my side the code looks good. I'd like to have at least 1 other reviewer approve this though. @tdomhan  ?

fhieber:
Also, this should get a minor version bump and a feature description in the changelog before merging.

mjpost:
Good idea on the unit test (much simpler than looking at beam output).

test/unit/test_prune.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fhieber: this should be 2018 for new files. But why adding a new file? `test_inference.py` would also work.

mjpost:
Okay, I think it's ready to merge. Do you want to do it? Should we wait for @tdomhan to take a look?

fhieber:
Lets wait for another review by @tdomhan

sockeye/inference.py
```diff
@@ -1276,43 +1326,50 @@ def _beam_search(self,
...
+            # (7) update best hypotheses, their attention lists and lengths (only for non-finished hyps)
             # pylint: disable=unsupported-assignment-operation
             sequences[:, t] = best_word_indices
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
             # (6) optionally save beam history
             if self.store_beam:
-                unnormalized_scores = scores_accumulated * self.length_penalty(lengths - 1)
+                unnormalized_scores = mx.nd.where(finished, scores_accumulated / self.length_penalty(lengths), scores_accumulated)
```
tdomhan: why normalize for `unnormalized_scores`? Why not just take `scores_accumulated` as is? `normalized_scores` down below should on the other hand now be normalized, right?

sockeye/inference.py
```diff
@@ -896,6 +900,8 @@ def __init__(self,
                  strip_unknown_words: bool = False) -> None:
         self.context = context
         self.length_penalty = length_penalty
+        self.beam_prune = beam_prune
```
tdomhan: can you add a `check_conditition` to make sure that `batch_size` is 1?

test/unit/test_inference.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
tdomhan: did we loose copyright for 2017? :P this should probably read: 
```
# Copyright 2017, 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

mjpost: I assumed this should just reflect the latest year. Is it supposed to be a range?

sockeye/inference.py
mjpost: It doesn't need to be size 1 --- this works with batching.

sockeye/inference.py
mjpost: Typo fixed (should be multiplying here, good catch). Normalization only applies to finished items now, so we only unnormalize there.

sockeye/inference.py
tdomhan: sorry, I meant for the `beam_search_stop`. That works only for batch_size 1, correct?

test/unit/test_inference.py
tdomhan: I think generally one should add each year in which changes were made to a file 
> To update the list of year numbers, add each year in which you have made nontrivial changes to the package
https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

sockeye/inference.py
mjpost: ah, of course, fixed.

sockeye/inference.py
tdomhan: oh ok, so you're unnormalizing the finished hypothesis. But then `beam_histories[sent]["normalized_scores"]` a little further down should be normalized, no?

mjpost: It is normalized, but only for finished hypotheses. And this has already been computed in `scores_accumulated` (line 1350).

tdomhan: yes, so I'm wondering whether you should normalize these before adding them to `beam_histories[sent]["normalized_scores"]` to stay consistent with the behavior before this change. Something like:
```
normalized_scores = mx.nd.where(finished, scores_accumulated, scores_accumulated / self.length_penalty(lengths))
```
As this is only done when storing the beam the computational overhead should be fine.

mjpost: I have done this, but I don't know what the desired behavior is. Perhaps @Deseaus can comment?

tdomhan: I think the visualization tools just report both the normalized and the unnormalized scores.

PR #233, opened 2017-12-02
[WIP] Fix stopping criterion bug in beam search

Deseaus:
Small fix for a large bug in beam search. In the context of generating [beam visualisation similar to OpenNMT's](http://opennmt.net/OpenNMT/translation/beam_search/#visualizing-the-beam-search) (PR coming at a later stage) I noticed that the beam search was running along many more time steps after all the hypothesis were completed, often hitting the max length. Here is an example:

beam_broken

In the above example, the beam went on for 4 extra steps that are essentially wasted computation. 

The problem was relying on `best_word_indices` to determine which hypothesis are finished. Since we run the whole beam forward, the hypothesis were generating beyond the `EOS_TOKEN` (ID 3), and this was throwing off the `finished` array.

For example, at this time step `t`, the second hypothesis has finished and all the values seem reasonable:

screen shot 2017-12-02 at 02 58 43

If we move the beam one step forward, `finished` only marks the fourth and fifth hypotheses as finished. The second one went on a loop generating token 30:

screen shot 2017-12-02 at 02 59 33

The fix is not to rely on `best_word_indices` to determine which hypothesis are complete, but get it directly from `sequences` through a `max` over axis 1.

I tested my change on a small model trained on IWSLT data and got a 1.5x speed up (GPU decoding on a p2.xlarge):

* **Without fix**: `[INFO:__main__] Processed 5431 lines in 5431 batches. Total time: 1773.8679, sec/sent: 0.3266, sent/sec: 3.0617`
* **With fix**: `[INFO:__main__] Processed 5431 lines in 5431 batches. Total time: 1181.2254, sec/sent: 0.2175, sent/sec: 4.5978`

The quality also improved (detokenised, IWSLT DE->EN test set):

* **Without fix**: 26.82 BLEU
* **With fix**: 32.76 BLEU

A quick diff on the tokenised data shows that a lot of the end repetitions are gone (fix on the right):

screen shot 2017-12-02 at 05 26 11

A look at a before and after on one of the more wasteful beam searches shows the improvement:

**Without fix**:

screen shot 2017-12-02 at 04 42 42

**With fix**:

screen shot 2017-12-02 at 04 42 54

## Still to do

* The above was tested on sockeye 1.12.0. I would have to confirm the changes on the latest master.
* PR forthcoming on adding the beam visualization and dumping to a json format, to better diagnose these kinds of problems in the future.
* It seems the problem is not fully fixed, as can be seen in one of the beam searches with the fix below, it seems that sockeye is still trying to generate hypotheses past an EOS. I haven't had time to do a deeper dive into this. Probably the beam search code should be rewritten to be cleaner and allow more flexibility for dropping/filtering hypotheses. For now the current change already brings improvement.

screen shot 2017-12-02 at 13 41 41

until you can check this box.
- [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.15]
+### Fixed
+- Fixed a bug during beam search stopping. Instead of stopping when all the `beam_size` hypothesis were finished, the model was continuing to decode due to losing track of when they had all finished. Now the models translate faster and quality also improved.
+
 ## [1.14.1]
 ### Changed
  - Sorting sentences during decoding before splitting them into batches.
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1085,7 +1085,7 @@ class Translator:
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
             # (6) determine which hypotheses in the beam are now finished
-            finished = ((best_word_indices == C.PAD_ID) + (best_word_indices == self.vocab_target[C.EOS_SYMBOL]))
+            finished = mx.nd.max(sequences == self.vocab_target[C.EOS_SYMBOL], axis=1)
             if mx.nd.sum(finished).asscalar() == self.batch_size * self.beam_size:  # all finished
                 break
 


tdomhan:
I'm not sure I fully understand this, yet. Once a hypothesis is finished we make sure that the hypothesis can only generate PAD symbols from that point onward:
```
...
                scores = mx.nd.where(finished, pad_dist, scores)
...
in https://github.com/awslabs/sockeye/blob/master/sockeye/inference.py#L1052
```
What you're saying is that this wasn't working correctly. I'd like to get a better understanding for why it wasn't correctly. Do you have an idea why?

Btw, I'm looking forward to the beam search visualizations :)

Deseaus:
As an update, I trained an RNN and a Transformer on sockeye 14.1 and the problem seems to be gone there. After producing an EOS token the rest of tokens are PAD:

screen shot 2017-12-04 at 14 06 56

The v12 model I had was a transformer, could be that there was a bug in the earlier transformer implementations. I'll have a dig and see. Either way, seems this is now fixed.

fhieber:
These visualizations look really cool, excited to see this coming :)

Agreeing with Tobi here. Theoretically it should not happen that a finished hypothesis is extended with a non-pad symbol, and in that case the updating of the finished hypothesis is correct.
As part of the fast-decoding transformers PR (#226) I did a similar change (https://github.com/awslabs/sockeye/pull/226/files#diff-955620b8cedc37b65c90b372a094fb17R1067), but it also should not be necessary.

tdomhan:
As discussed, I'll close the PR for now.

PR #235, opened 2017-12-05
Minotr rename in inference.

tdomhan:

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -900,8 +900,8 @@ class Translator:
             model_probs.append(probs)
             model_attention_probs.append(attention_probs)
             model_states.append(state)
-        probs, attention_probs = self._combine_predictions(model_probs, model_attention_probs)
-        return probs, attention_probs, model_states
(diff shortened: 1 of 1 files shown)


sockeye/inference.py
```diff
@@ -900,8 +900,8 @@ def _decode_step(self,
             model_probs.append(probs)
             model_attention_probs.append(attention_probs)
             model_states.append(state)
-        probs, attention_probs = self._combine_predictions(model_probs, model_attention_probs)
-        return probs, attention_probs, model_states
+        neg_logprobs, attention_probs = self._combine_predictions(model_probs, model_attention_probs)
```
fhieber: would it make sense to rename the probs variable a couple of lines above as well?

tdomhan: the conversion happens only in combine_predictions so these are actually probabilities and not logprobs. That's actually the reason I wanted to rename the probs after combine_predictions to make that clear.

fhieber: uh yes you are right!

PR #236, opened 2017-12-05
Changed bounds and fixed sorting sys test.

tdomhan:


mjdenkowski:
Procedural question: do we want to include fixes like this in the changelog and increment the minor version?

tdomhan:
I'd leave it out as it's not directly user facing, but I'd be fine either way.

PR #247, opened 2017-12-14
Warn when waiting for checkpoint decoder.

tdomhan:
- Warn when waiting for the checkpoint decoder.
 - Removed functionality to have multiple concurrent decoder, as this
 can in the worst case lead to the training being interrupted, as each
 decoder potentially delays has_improved, which in combination with slow
 decoders leads to the early stopping criterion kicking in. With this
 change the delay is at most one.

until you can check this box.
- [X] Passed code style checking (`./pre-commit.sh` or manual run of pylint & mypy)
- [X] Updated CHANGELOG.md


tdomhan:
yeah, this only is an issue when using the BLEU score. The main point was to add the logging when we wait. As we anyway don't use the other feature and there's a potential issue with when we optimize for BLEU, I remove it in the same commit.

fhieber:
Sounds good!

fhieber:
@tdomhan I will merge after rebase

fhieber:
The travis fail seems to be unrelated: fail of pylint due to matplotlib import error. I will merge this.

PR #264, opened 2018-01-09
init_embedding.py now supports all vocabulary-sized weights and raw Sockeye parameters

xingniu:
init_embedding.py now supports all vocabulary-sized weights (e.g. source/target_embed_weight, target_output_weight and target_output_bias) and raw Sockeye parameters (e.g. params.XXX).

until you can check this box.
- [X] Passed code style checking (`./pre-commit.sh` or manual run of pylint & mypy)
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/init_embedding.py
+++ b/sockeye/init_embedding.py
@@ -15,10 +15,11 @@
 
 """
 Initializing Sockeye embedding weights with pretrained word representations.
+It also supports updating vocabulary-sized weights for a new vocabulary.
 
 Quick usage:
 
-    python3 -m contrib.utils.init_embedding        \
+    python3 -m sockeye.init_embedding        \
             -e embed-in-src.npy embed-in-tgt.npy    \
             -i vocab-in-src.json vocab-in-tgt.json   \
             -o vocab-out-src.json vocab-out-tgt.json  \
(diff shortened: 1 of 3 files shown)


tdomhan:
Thanks for the PR! Could you confirm this contribution is under the terms of the Apache 2.0 license?

sockeye/init_embedding.py
```diff
@@ -28,8 +29,11 @@
 Optional arguments:
 
     --embeddings, -e
```
tdomhan: Now that this change has been generalized to any kind of weight it might make sense to rename the argument (and maybe also the module) to `--weights` or `--weight-files`. What do you think?

sockeye/init_embedding.py
```diff
@@ -104,21 +111,26 @@ def main():
...
-    for embed_file, vocab_in_file, vocab_out_file, name in zip(args.embeddings, args.vocabularies_in, \
+    for weight_file, vocab_in_file, vocab_out_file, name in zip(args.embeddings, args.vocabularies_in, \
                                                                args.vocabularies_out, args.names):
-        logger.info('Loading input embedding weight: %s', embed_file)
-        embed = np.load(embed_file)
+        logger.info('Loading input weight: %s', weight_file)
+        if weight_file.endswith(".npy"):
+            weight = np.load(weight_file)
+        elif weight_file.endswith(".npz"):
+            weight = np.load(weight_file)[name]
```
tdomhan: for both the npz and the mx.nd.load potentially multiple weights come from the same file. Would it make sense to either: load the file only once and then pull out the different weights or restructure the CLI slightly, say:
```
--embeddings emb.npy params.best --names emb weight1:weight2
```
so that for each file one basically accepts a list of names (separated by `:` or some other delimiter).

xingniu:
This contribution is under the terms of the Apache 2.0 license.

sockeye/init_embedding.py
xingniu: I cached loaded files instead of changing the CLI. If weight-file->weight-name supports 1-N mappings, then vocab-file->weight-name is also supposed to support 1-N mappings, which makes conditions too complicated. Let's only use 1-1 mappings since it's clear.

tdomhan: sounds good.

PR #266, opened 2018-01-11
[WIP] initial implementation of scheduled sampling; do not merge

acclift:
Feedback requested: basic implementation of scheduled sampling performed in sym_gen rather than in the decoder. This version runs but converges much more slowly than the vanilla version without sampling. There is an issue in tracking the 'updates' state of the training module to where we compute the sampling threshold.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -46,6 +46,35 @@ def global_norm(ndarrays: List[mx.nd.NDArray]) -> float:
     norms = [mx.nd.square(mx.nd.norm(arr)) for arr in ndarrays if arr is not None]
     return sqrt(sum(norm.asscalar() for norm in norms))
 
+def check_scheduled_sampling_params(scheduled_sampling_type: str, scheduled_sampling_params: List[float]) -> None:
+    if scheduled_sampling_type == 'inv-sigmoid-decay':
+        assert len(scheduled_sampling_params) == 2, \
+            ('The inverse sigmoid decaying option for scheduled sampling requires 2 parameters,'
+             ' but given {}.').format(len(scheduled_sampling_params))
+
+        k, lower_bound = scheduled_sampling_params
+        assert k >= 1, 'Offset should be greater than or equal to 1.'
+        assert lower_bound >= 0 and lower_bound < 1, 'Lower bound should be 0 or greater and less than 1.'
+
+    elif scheduled_sampling_type == 'exponential-decay':
+        assert len(scheduled_sampling_params) == 1, \
+            ('The exponential decaying option for scheduled sampling requires 1 parameter,'
+             ' but given {}.').format(len(scheduled_sampling_params))
+
+        k = scheduled_sampling_params[0]
+        assert k < 1, 'Offset for the exponential decay should be less than 1.'
+
+    elif scheduled_sampling_type == 'linear-decay':
+        assert len(scheduled_sampling_params) == 3, \
+            ('The linear decaying option for scheduled sampling requires 3 parameters,'
+             ' but given {}').format(len(scheduled_sampling_params))
+
+        k, epsilon, slope = scheduled_sampling_params
+        assert k <= 1, 'Offset for the linear decay should be less than 1.'
+        assert epsilon >= 0, 'Epsilon for the linear decay should be greater than or equal to 0.'
+
+
+
 
 class _TrainingState:
     """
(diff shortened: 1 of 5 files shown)


sockeye/training.py
```diff
@@ -85,15 +114,40 @@ def __init__(self,
...
+        elif self.config.scheduled_sampling_type == 'exponential-decay':
+            k = self.config.scheduled_sampling_params[0]
+            sampling_scheduler = lambda: pow(k, i)
+        elif self.config.scheduled_sampling_type == 'linear-decay':
+            k, epsilon, slope = self.config.scheduled_sampling_params
+            sampling_scheduler = lambda: mx.sym.clip(k - slope * i, epsilon, k)
+        else:
+            sampling_scheduler = lambda: 1
+
+        return sampling_scheduler
```
tdomhan: why do you return a function instead of returning the Symbol directly?

sockeye/training.py
```diff
@@ -85,15 +114,40 @@ def __init__(self,
...
+            k, lower_bound = self.config.scheduled_sampling_params
+            sampling_scheduler = lambda: k / (k + mx.sym.exp(i / k) - 1) * (1 - lower_bound) + lower_bound
+        elif self.config.scheduled_sampling_type == 'exponential-decay':
+            k = self.config.scheduled_sampling_params[0]
+            sampling_scheduler = lambda: pow(k, i)
+        elif self.config.scheduled_sampling_type == 'linear-decay':
+            k, epsilon, slope = self.config.scheduled_sampling_params
+            sampling_scheduler = lambda: mx.sym.clip(k - slope * i, epsilon, k)
+        else:
+            sampling_scheduler = lambda: 1
```
tdomhan: this is probably not going to work, as this doesn't return a symbol. you could do `mx.sym.ones(...)` instead

sockeye/training.py
```diff
@@ -188,6 +296,14 @@ def create_eval_metric_composite(metric_names: List[AnyStr]) -> mx.metric.Compos
         metrics = [TrainingModel.create_eval_metric(metric_name) for metric_name in metric_names]
         return mx.metric.create(metrics)
 
+    def _set_states(self, state_names, values):
```
fhieber: doesn't this function exist in the Module implementation of MXNet already (https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/bucketing_module.py#L269) ?

sockeye/training.py
```diff
@@ -135,17 +185,73 @@ def sym_gen(seq_lens):
...
+                target_sliced = mx.sym.split(data=target, num_outputs=target_seq_len, axis=1, squeeze_axis=True)
+
+                for seq_idx in range(target_embed_seq_len):
+                    if seq_idx == 0:
+                        target_prev = target_sliced[seq_idx]
+
+                    else:
+                        rand_val = mx.sym.random_uniform(low=0, high=1.0, shape=1)[0]
+                        target_prev = mx.sym.where(rand_val < threshold,
+                                                     mx.sym.cast(target_sliced[seq_idx], dtype='int32'), sampled_word)
```
fhieber: its a bit unclear how sampled_word is initialized due to the special case if seq_idx==0.

sockeye/training.py
```diff
@@ -135,17 +185,73 @@ def sym_gen(seq_lens):
...
 
-            # decoder
-            # target_decoded: (batch-size, target_len, decoder_depth)
-            target_decoded = self.decoder.decode_sequence(source_encoded, source_encoded_length, source_encoded_seq_len,
+            # target embedding
+            (target_embed,
+             target_embed_length,
+             target_embed_seq_len) = self.embedding_target.encode(target, target_length, target_seq_len)
+
+            if self.config.scheduled_sampling_type is not None:
```
fhieber: it would be nice to move this into its own function to avoid the large if clause

sockeye/training.py
```diff
@@ -135,17 +185,73 @@ def sym_gen(seq_lens):
...
+                # slice target words
+                # target: target_seq_len * (batch_size)
+                target_sliced = mx.sym.split(data=target, num_outputs=target_seq_len, axis=1, squeeze_axis=True)
+
+                for seq_idx in range(target_embed_seq_len):
+                    if seq_idx == 0:
+                        target_prev = target_sliced[seq_idx]
+
+                    else:
+                        rand_val = mx.sym.random_uniform(low=0, high=1.0, shape=1)[0]
```
fhieber: why `[0]`? There should be a way to make sure rand_val and threshold are inferred to the same shape.

sockeye/training.py
```diff
@@ -599,6 +716,12 @@ def _evaluate(self, training_state, val_iter, val_metric):
         val_iter.reset()
         val_metric.reset()
 
+        state_names = ['training_state.updates']
```
fhieber: This could use a comment why we do this. It could also use the _set_state() function you define above.
Also why are we resetting the update count back to 0 here?

sockeye/training.py
```diff
@@ -188,6 +296,14 @@ def create_eval_metric_composite(metric_names: List[AnyStr]) -> mx.metric.Compos
         metrics = [TrainingModel.create_eval_metric(metric_name) for metric_name in metric_names]
         return mx.metric.create(metrics)
 
+    def _set_states(self, state_names, values):
+        states = self.module.get_states(merge_multi_context=False)
+        for state_name, state, value in zip(state_names, states, values):
+            # ACCLIFT: this is something that existed in the previous initial implementation.
```
fhieber: This seems related to the scheduling where you need the # of updates symbolically in the graph.

sockeye/training.py
fhieber: It could also use some shape information in comments

sockeye/training.py
```diff
@@ -135,17 +185,73 @@ def sym_gen(seq_lens):
...
+                    # target_decoded: (batch-size, target_len, decoder_depth)
+                    (target_decoded,
+                     attention_probs,
+                     states) = self.decoder.decode_step(seq_idx, target_embed_prev, source_encoded_seq_len, *states)
+
+                    # logit: (batch_size, target_vocab_size)
+                    logit = self.output_layer(target_decoded)
+                    logits_all.append(logit)
+
+                    prob = mx.sym.softmax(logit)
```
fhieber: would be good to document shapes of these here.

sockeye/training.py
artemsok: It's a bit confusing at first to see sampled_word used before it's initialized (after realizing that it's a loop it becomes clear, but could be something to rewrite for readability).

If I understood it right, target_prev is either the current, **seq_idx**'th, target word or the sampled word from the (**seq_idx-1**)'th distribution, since sampling happened in the previous iteration. Is it intended, shouldn't we chose between a target word and a sample from the distribution from the same position?

sockeye/training.py
```diff
@@ -85,15 +114,40 @@ def __init__(self,
...
+        """
+        Computes the threshold over which we train against model samples rather than gold targets
+        """
+        i = mx.sym.Variable('updates')
+
+        check_scheduled_sampling_params(self.config.scheduled_sampling_type, self.config.scheduled_sampling_params)
+
+        if self.config.scheduled_sampling_type == 'inv-sigmoid-decay':
+            k, lower_bound = self.config.scheduled_sampling_params
+            sampling_scheduler = lambda: k / (k + mx.sym.exp(i / k) - 1) * (1 - lower_bound) + lower_bound
```
artemsok: If lower_bound is just to lowerbound, wouldn't mx.sym.max() be enough?
Why doesn't exponential decay need it?

sockeye/training.py
```diff
@@ -85,15 +114,40 @@ def __init__(self,
...
+        elif self.config.scheduled_sampling_type == 'exponential-decay':
+            k = self.config.scheduled_sampling_params[0]
+            sampling_scheduler = lambda: pow(k, i)
+        elif self.config.scheduled_sampling_type == 'linear-decay':
+            k, epsilon, slope = self.config.scheduled_sampling_params
+            sampling_scheduler = lambda: mx.sym.clip(k - slope * i, epsilon, k)
```
artemsok: mx.sym.max(k - slope * i, epsilon) might be conciser?

fhieber:
@acclift I am closing this due to inactivity. Feel free to reopen when relevant again.

Issue #209, opened 2017-11-15
Potential error of "permission denied" when trying to open lock files


xingniu:
A "permission denied" error would be raised if the lock file was owned by another user.
Is moving L552 into try-except a possible solution?
https://github.com/awslabs/sockeye/blob/d436ae8319e9cd531a562434f5022029ff933d6c/sockeye/utils.py#L549-L575

tdomhan:
oh ok good point, we should fix this. In your scenario does `open` through an IOError with errno.EAGAIN?

xingniu:
I got "permission denied", so I think it's errno.EACCES. In this scenario, it's better that Sockeye tries to find another GPU.

bricksdont:
Yes, but in case a specific device was requested by the user with `--device` I don't think sockeye should try to aquire a GPU that was not listed in this option, right?

xingniu:
Right, "find another GPU" if `--device` got an negative value.

tdomhan:
So this PR https://github.com/awslabs/sockeye/pull/268 should fix the issue and retry when we get a permission denied for the lock. Does this work for you?

What was the other point regarding `--device`? Is there some other issue with the way locking works right now?

xingniu:
Thanks! It's ok for me if Sockeye behaves the same when the lock is owned by self and others.

bricksdont:
@tdomhan Sorry I actually meant `--device-ids`, and in the context of xingniu's comment:

>In this scenario, it's better that Sockeye tries to find another GPU.

this should only happen if `--device-ids` has a negative value. But, as far as I can see, that's not an issue in your PR.

tdomhan:
ah ok. Yes, if you specify non-negative device ids with the recent change sockeye will wait for these specific devices to become available.

I guess one potential issue with lock files owned by other people is that in case sockeye gets shut down non-gracefully and the lock files don't get deleted, then someone will need to manually delete the files. This is not the case when there are no permission errors, as there we don't rely on the existence of the file, but rather on the file system lock provided by the operating system, which ceases to exist as soon as a process dies. However, I don't see a better way of handling it, than we do now.

tdomhan:
The code has been merged. Let us know if there are issues.

PR #268, opened 2018-01-12
Retrying gpu locking on permission denied.

tdomhan:
Solves Issue https://github.com/awslabs/sockeye/issues/209.

### DIFF
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -610,7 +610,12 @@ class GpuFileLock:
     def __enter__(self) -> Optional[int]:
         for gpu_id in self.candidates:
             lockfile_path = os.path.join(self.lock_dir, "sockeye.gpu%d.lock" % gpu_id)
-            lock_file = open(lockfile_path, 'w')
+            try:
+                lock_file = open(lockfile_path, 'w')
+            except IOError as e:
+                if errno.EACCES:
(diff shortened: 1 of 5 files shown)


sockeye/utils.py
```diff
@@ -610,7 +610,12 @@ def __init__(self, candidates: List[int], lock_dir: str) -> None:
     def __enter__(self) -> Optional[int]:
         for gpu_id in self.candidates:
             lockfile_path = os.path.join(self.lock_dir, "sockeye.gpu%d.lock" % gpu_id)
-            lock_file = open(lockfile_path, 'w')
+            try:
+                lock_file = open(lockfile_path, 'w')
+            except IOError as e:
+                if errno.EACCES:
+                    logger.warning("GPU %d is currently locked by a different process (Permission denied).", gpu_id)
+                    return None
```
xingniu: Is it better to continue the for loop and try to find another available GPU? The same for L641.

tdomhan: this is what effectively happens. It just happens outside the class in this while loop: https://github.com/awslabs/sockeye/blob/master/sockeye/utils.py#L563

tdomhan: (where we check whether `None` was returned)

xingniu: Got it. Thanks!

tdomhan:
travis seems to be down :/

Issue #234, opened 2017-12-04
Missing params.best


konrad-roze:
If you set low _max_num_epochs_ and _train_early_stop_when_ values then _params.best_ file is often missing although the training has succeeded. This is likely caused by the fact that there was no improvement of the model.
The params.best file should always be present after a model is trained.

fhieber:
This can happen if you there not a single checkpoint made during training, independent whether there is any improvement. 
How many updates are performed in these cases? Can you set the `--checkpoint-frequency` to <= that number?

We can think about always doing at least one checkpoint when exiting the training loop.

konrad-roze:
Decreasing checkpoint frequency in connection with reducing batch size worked for me as a workaround, thanks.
Still you probably need to handle this corner case differently. If you're not planning a fix, feel free to close this issue.

tdomhan:
Could you share the exact CLI parameters and the size of your dataset?

Is it the case that during training the optimized metric did not improve a single time? I would probably make sense to handle this corner case. In any case something is probably fishy if your validation metric does not improve a single time during training. Are you dealing with a very small data set?

konrad-roze:
This is very degenerate case that I cooked up for workflow testing/development. It turned out that my batch size was 2000 but I had only 1000 (or less) lines of input. Consequently there was no single checkpoint during the training. Still I believe that params.best should be produced and contain any/initial parameters.

tdomhan:
I agree that a model should always be valid. With respect to testing it might make sense to take a look at the system copy test we use here:
https://github.com/awslabs/sockeye/blob/master/test/system/test_seq_copy_sys.py 
The test creates sequences of digits and we train small models to copy these sequences. The learning of these can be done in only a few updates.

tdomhan:
This PR should fix the issue by writing out the initial parameters: https://github.com/awslabs/sockeye/pull/269 Let us know if that works for you.

fhieber:
I am closing this for now, feel free to reopen if #269 did not solve the problem.

PR #269, opened 2018-01-12
Model file valid before the first checkpoint.

tdomhan:
This resolves Issue https://github.com/awslabs/sockeye/issues/234. With the change the model folder is valid right from the start (as the params.best file will exist). This way one does no long get more not completely obvious error messages from sockeye.translate when trying to translate with a model that yet has to write the first checkpoint.

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -273,6 +273,8 @@ class TrainingModel(model.SockeyeModel):
         self.module.init_params(initializer=initializer, arg_params=self.params, aux_params=None,
                                 allow_missing=allow_missing_params, force_init=False)
         self._log_params()
+        initial_params_fname = self._save_params(output_folder, checkpoint=0)
+        os.symlink(initial_params_fname, os.path.join(output_folder, C.PARAMS_BEST_NAME))
 
         self.module.init_optimizer(kvstore=kvstore, optimizer=optimizer, optimizer_params=optimizer_params)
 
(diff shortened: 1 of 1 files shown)


chrkell:
I trained a model with Sockeye and my model directory is missing the params.best file. 

The error: sockeye.utils.SockeyeError: No model parameter file found under ../sockeye_training_output/params.best. This is either not a model directory or the first training checkpoint has not happened yet.

These were my parameters during training: 
```
params = {
        '--source': '../sentence_parallel_files/src_train.txt',
        '--target': '../sentence_parallel_files/tgt_train.txt',
        '--validation-source': '../sentence_parallel_files/src_validation.txt',
        '--validation-target': '../sentence_parallel_files/tgt_validation.txt',
        '--output': '../sockeye_training_output',
        '--shared-vocab': '',
        '--num-words': '20000',
        '--num-layers': '5:5',
        '--transformer-model-size': '512:512',
        '--transformer-attention-heads': '4:4',
        '--transformer-feed-forward-num-hidden': '2048:2048',
        '--embed-dropout': '0.3',
        '--label-smoothing': '0.3',
        '--optimized-metric': 'bleu',
        '--checkpoint-interval': '10',
        '--max-samples': '10000000'
    }
```

What do I do?

mjdenkowski:
Can you share the contents of your model directory (`ls sockeye_training_output`) and the tail of the training log (`tail -n20 sockeye_training_output/log`)?

chrkell:
Model directory:
Bildschirm­foto 2023-10-20 um 09 20 03

Tail of training log:
Bildschirm­foto 2023-10-20 um 09 39 50

mjdenkowski:
It looks like training only ran for 1 checkpoint and may have ended with an error, although no error is reported. The model is early enough in training that we wouldn't expect it to produce any useful outputs.

At this point, there isn't much information available to diagnose potential issues. A first thing to try would be rerunning training with a smaller model to test the general setup. You could try `--transformer-model-size 128 --transformer-feed-forward-num-hidden 512 --num-layers 3:3`.

PR #277, opened 2018-01-19
Stable system tests.

tdomhan:
Our system tests currently fail on the Travis nightly cronjob. The reason mainly is that the bounds were originally created on a different machine. Additionally, there is the complexity that the system tests result in different training curves with each and every run on Travis, so there is some variation across runs that must be accounted for. I would argue that the bounds of the tests should be set in a way that given this variation we are sure not to be outside of the bounds. Namely, we will catch any bug where we observe crashes or where the perplexity or BLEU score changes considerably. Having run almost 100 builds on Travis I hope that these values provide a more stable way of running the nightly system tests, by hopefully having no more false positives.


test/system/test_seq_copy_sys.py
```diff
@@ -25,8 +26,14 @@
 _TEST_LINE_COUNT = 110
 _TEST_LINE_COUNT_EMPTY = 10
 _TEST_MAX_LENGTH = 11
-_SEED_TRAIN = 13
-_SEED_DEV = 17
+_SEED_TRAIN_DATA = 13
+_SEED_DEV_DATA = 17
+
+# Training on different systems, as it happens in the Travis system tests, always results in different training curves
```
fhieber: thanks for adding this note!

PR #281, opened 2018-01-25
Fp16 mode - Rough work in progress.  Opened for reference only.

KellenSunderland:


sockeye/arguments.py
```diff
@@ -260,6 +260,11 @@ def add_training_data_args(params, required=False):
                         required=required,
                         type=regular_file(),
                         help='Target side of parallel training data.')
+    params.add_argument(C.TRAINING_ARG_DTYPE, '-dt',
```
fhieber: Is there a good reason to define this as a constant?

sockeye/arguments.py
```diff
@@ -260,6 +260,11 @@ def add_training_data_args(params, required=False):
...
+                        default='float32',
```
fhieber: I'd rather define the allowed values here as constants in constants.py

sockeye/arguments.py
```diff
@@ -260,6 +260,11 @@ def add_training_data_args(params, required=False):
...
+                        choices=['float16', 'float32'],
+                        type=str,
+                        help='Data type to use when training a model, float16 or float32. Default: %(default)s')
```
fhieber: [nitpick]: Period missing at the end :)

sockeye/arguments.py
```diff
@@ -1009,6 +1014,18 @@ def add_inference_args(params):
                                help='Beta factor for the length penalty used in beam search: '
                                     '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
 
+    decode_params.add_argument('--encoder-dtype',
```
fhieber: same comments as above, but more generally: is there a reason to allow different dtypes for encoder/ decoder?
If so I'd suggest going for the 'two int' flag style we have for other arguments as well (see for example `--num-embed`)

sockeye/arguments.py
```diff
@@ -1044,14 +1061,14 @@ def add_build_vocab_args(params):
 
 
 def add_init_embedding_args(params):
-    params.add_argument('--weight-files', '-w', required=True, nargs='+',
-                        help='List of input weight files in .npy, .npz or Sockeye parameter format.')
+    params.add_argument('--embeddings', '-e', required=True, nargs='+',
```
fhieber: that seems to be an artifact of an outdated master

sockeye/constants.py
```diff
@@ -330,6 +331,9 @@
 LARGE_POSITIVE_VALUE = 99999999.
 LARGE_NEGATIVE_VALUE = -LARGE_POSITIVE_VALUE
 
+LARGE_POSITIVE_VALUE_FP_16 = np.finfo(np.float16).max
```
fhieber: this should be based on `--dtype`

sockeye/data_io.py
```diff
@@ -579,7 +579,8 @@ def get_validation_data_iter(data_loader: RawParallelDatasetLoader,
                              max_seq_len_source: int,
```
fhieber: somewhere in data_io.py we are creating the numpy / ndarrays for each bucket. These should probably also have a certain dtype (they are actually int so they should be int anyway), right now they are probably the default dtype of `mx.nd.array()`.

sockeye/inference.py
```diff
@@ -60,12 +65,18 @@ def __init__(self,
...
-                 cache_output_layer_w_b: bool = False) -> None:
+                 cache_output_layer_w_b: bool = False,
+                 encoder_dtype=np.float32,
+                 decoder_dtype=np.float32) -> None:
         self.model_version = utils.load_version(os.path.join(model_folder, C.VERSION_NAME))
         logger.info("Model version: %s", self.model_version)
         utils.check_version(self.model_version)
 
         config = model.SockeyeModel.load_config(os.path.join(model_folder, C.CONFIG_NAME))
+        config.config_encoder.dtype = encoder_dtype
```
fhieber: why implement this through setters and not in the constructor?

sockeye/encoder.py
```diff
@@ -56,11 +56,13 @@ class RecurrentEncoderConfig(Config):
     def __init__(self,
                  rnn_config: rnn.RNNConfig,
                  conv_config: Optional['ConvolutionalEmbeddingConfig'] = None,
-                 reverse_input: bool = False) -> None:
+                 reverse_input: bool = False,
+                 use_fp16: bool = False) -> None:
```
fhieber: I think it would be nicer to rename this to dtype and allow the actualy dtype (imagine we want to train with float64 at some point).

sockeye/arguments.py
KellenSunderland: Spoke with Tobi about this.  This is present only for debugging purposes in this version.  In future versions we propose having a single dtype.  This dtype should actually not need to be set during inference as it should be present in the model config.  Setting it will be enabled for cases where the user wants to override behaviour (for example converting from float32 params in an existing model to float16).

sockeye/encoder.py
KellenSunderland: +1

sockeye/inference.py
KellenSunderland: +1

sockeye/constants.py
KellenSunderland: +1

sockeye/data_io.py
KellenSunderland: Good call out, I'll take a look.

KellenSunderland:
Thanks for the review Felix.  We'll have a more polished version eventually, this is really just testing code.

sockeye/data_io.py
```diff
@@ -1228,14 +1237,14 @@ def __init__(self,
         self.provide_data = [
             mx.io.DataDesc(name=self.source_data_name,
                            shape=(self.bucket_batch_sizes[-1].batch_size, self.default_bucket_key[0]),
-                           layout=C.BATCH_MAJOR),
+                           layout=C.BATCH_MAJOR, dtype=dtype),
```
tdomhan: shouldn't these be int32? I think you specify them as float above?

sockeye/inference.py
```diff
@@ -271,7 +310,11 @@ def _get_decoder_data_shapes(self, bucket_key: Tuple[int, int]) -> List[mx.io.Da
         source_max_length, target_max_length = bucket_key
         return self.decoder_data_shapes_cache.setdefault(
             bucket_key,
-            [mx.io.DataDesc(name=C.TARGET_NAME, shape=(self.batch_size * self.beam_size,), layout="NT")] +
+            [mx.io.DataDesc(
+                name=C.TARGET_NAME,
+                shape=(self.batch_size * self.beam_size,),
+                dtype=self.decoder.dtype,
```
tdomhan: is this float or int? should be int, no?

KellenSunderland:
This PR is outdated, Pavel's is newer.  We'll take the comments into account before submitting final PR.

PR #284, opened 2018-01-30
Fixed sentence-level BLEU calculation with sockeye.evaluate. Also som…

fhieber:
…e code reformatting.

until you can check this box.

### DIFF
--- a/contrib/sacrebleu/sacrebleu.py
+++ b/contrib/sacrebleu/sacrebleu.py
@@ -149,21 +149,20 @@ Originally written by Matt Post.
 The official version can be found at github.com/awslabs/sockeye, under `contrib/sacrebleu`.
 """
 
-import re
-import os
-import sys
-import math
-import gzip
-import tarfile
-import logging
-import urllib.request
 import argparse
-import unicodedata
-import numpy
-
+import gzip
+import logging
+import os
+import re
+import sys
+import tarfile
+import urllib.request
 from collections import Counter, namedtuple
 from itertools import zip_longest
 from typing import List, Iterable, Tuple
+
+import math
+import unicodedata
 
 VERSION = '1.2.1'
 
@@ -769,6 +768,7 @@ TOKENIZERS = {
 }
 DEFAULT_TOKENIZER = '13a'
 
+
 def _open(file, encoding='utf-8'):
     """Convenience function for reading compressed or plain text files.
     :param file: The file to read.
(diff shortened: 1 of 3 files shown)


contrib/sacrebleu/sacrebleu.py
```diff
@@ -149,22 +149,21 @@
...
-import unicodedata
-import numpy
-
 from collections import Counter, namedtuple
 from itertools import zip_longest
 from typing import List, Iterable, Tuple
 
+import math
+import unicodedata
+
```
mjpost: What is the logic behind this reorganization? Why are math and unicodedata pulled out below?

sockeye/evaluate.py
```diff
@@ -38,7 +38,8 @@ def raw_corpus_bleu(hypotheses: Iterable[str], references: Iterable[str], offset
     :param offset: Smoothing constant.
     :return: BLEU score as float between 0 and 1.
     """
-    return sacrebleu.raw_corpus_bleu(hypotheses, [references], smooth_floor=offset).score / 100
+    return sacrebleu.raw_corpus_bleu(hypotheses, [references], smooth_floor=offset).score / 100.0
+
```
mjpost: Python 3 is supposed to be robust to this, right? `//` is for explicitly integer division. This is fine for clarity but it shouldn't have any effect.

sockeye/evaluate.py
```diff
@@ -90,7 +93,7 @@ def main():
             scores = []
             for metric in args.metrics:
                 if metric == C.BLEU:
-                    bleu = raw_corpus_bleu(h, r, args.offset)
+                    bleu = raw_corpus_bleu([h], [r], args.offset)
                     scores.append("%.6f" % bleu)
                 elif metric == C.CHRF:
                     chrf_score = raw_corpus_chrf(h, r)
```
mjpost: I assume this is the bug? This is really a problem with our interface because this mistake is too easy to make. What do you think about adding some stricter type checking? Maybe the problem is unavoidable because I'm not sure that we want to disallow passing in strings, but that might be the better option.

contrib/sacrebleu/sacrebleu.py
fhieber: Just running import organization with an IDE

sockeye/evaluate.py
fhieber: Sure that doesn't make a difference, just for clarity.

sockeye/evaluate.py
fhieber: We could do some stricter type checking but ideally we shouldn't call the function `corpus_bleu` anyway if we want to compute sentence BLEU. My suggestion would be to add a `sentence_bleu` function in sockeye/evaluate.py that calls sacrebleu BLEU code to compute (properly smoothed) sentence-level BLEU scores.

PR #285, opened 2018-01-30
Store beam search history

Deseaus:
First draft of the beam search history storing and visualization to get feedback before further iteration.

During each step in the beam, store the predicted ids, predicted tokens, their parent ids, normalized and unormalized scores. These get returned from the `Translator.translate()` call by passing them along the `Translation` and `TranslatorOuput` namedtuples. From there, the information can be used by one of the output handlers to write it out to json format (there is currently some dummy code to print to a file for testing within the beam search).

Separately, I have a modified version of [OpenNMT's VisTools](https://github.com/OpenNMT/VisTools) that can take the outputted beam histories in `json` and generate interactive graphs similar to these (for a full interactive version, download and have a look at the zipped folder further below):

screen shot 2018-01-30 at 17 23 59

Ideally, I'd like to add this modified version to `sockeye/contrib`, license allowing. The VisTools package has an MIT license, while the `generate_beam_viz.py` has Apache 2.0 and Google copyright. I can add the code to the PR if the licensing seems ok to you.

Known issues:

* `_concat_translations` is a bit awkward to handle, since you can't really conceptually merge beam histories from two different sentence chunks into a whole sentence. For now, the individual beam histories is added as a list in this case.
* In the generated graph, some of the nodes and labels overlap. Had a cursory look and might be complicated to fix, as it goes into how d3js generates the graphs. Should still be useful for now though.

Full example folder, with `json` input file and generated beam visualizations:

[18-01-26.zip](https://github.com/awslabs/sockeye/files/1678467/18-01-26.zip)

until you can check this box.
- [x] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs? (Not modified)
- [x] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -551,13 +552,16 @@ def get_max_output_length(input_length: int):
 :param tokens: List of translated tokens.
 :param attention_matrix: Attention matrix. Shape: (target_length, source_length).
 :param score: Negative log probability of generated translation.
+:param beam_history: List of beam histories. The list will contian more than one
+history if it was split due to exceeding max_length.
 """
 
 TokenIds = List[int]
 Translation = NamedTuple('Translation', [
```
fhieber: Can we make this a proper class to have direct typing in the constructor?
Also we could type `beam_history` to `Optional[List[Dict[str, List]]]`, or maybe make the inner dict a class itself.

sockeye/inference.py
```diff
@@ -986,6 +993,15 @@ def _beam_search(self,
                                dtype='int32')
         sequences[:, 0] = self.start_id
 
+        # Beam history
+        beam_histories = []
+        for _ in range(self.batch_size):
+            beam_histories.append({"scores": [],
```
fhieber: lets make the inner dict either a class or you can use defaultdict

sockeye/inference.py
```diff
@@ -1051,6 +1067,9 @@ def _beam_search(self,
                 scores = scores[:1] / self.length_penalty(lengths[:1])
             else:
                 # renormalize scores by length ...
+
+                # TODO hussed: why keep multiplying back the LP added in the
```
fhieber: its for renormalization since we are always carrying all hypotheses (including finished ones) through. This is being discussed in the pruning PR.

Lets remove the TODOs that ask questions and discuss them in this PR.

sockeye/inference.py
```diff
@@ -986,6 +993,15 @@ def _beam_search(self,
                                dtype='int32')
         sequences[:, 0] = self.start_id
 
+        # Beam history
+        beam_histories = []
+        for _ in range(self.batch_size):
+            beam_histories.append({"scores": [],
+                                 "normalized_scores": [],
```
fhieber: why do we need normalized and unnormalized scores?

sockeye/inference.py
```diff
@@ -1089,22 +1108,55 @@ def _beam_search(self,
             attentions[:, t, :] = attention_scores
             lengths += mx.nd.cast(1 - mx.nd.expand_dims(finished, axis=1), dtype='float32')
 
-            # (6) determine which hypotheses in the beam are now finished
+
+            # (6) save beam history
+            # TODO hussed: check speed, if slow,  condition this block on some CLI flag so it doesn't run in prod
```
fhieber: I think this should be off by default to reduce overhead, for example lets implement another output_handler that enables this feature, otherwise we leave it off.
I would also factor this out into some kind of `_update_beam_history()` function to make the length of _beam_search smaller.

Deseaus: I agree, the problem I had was how to implement the switch. Where should I get the condition the `if store_beam:` should check? Should it just be an extra arg to `Translator().__init__()`? It would have to pass down quite a few functions to reach here.

About factoring out, I would rather leave it here for now, it's not too much code. We had discussed with Matt to move the beam search function to a more flexible and readable class. I'll probably work on that after the visualization is done.

sockeye/inference.py
Deseaus: I was going to add both to the beam visualisation, it's interesting to see how normalization rescores the hypothesis. I imagine it could help to see better why a certain path was chosen due to normalization. What do you think?

fhieber: the reason I am asking is that you then need to unnormalize twice, once in-place in the existing code, and once for the beam_history logging.
I would say its fine to have as long as we completely disable the logging by default and only turn it on via a CLI flag / output-handler type.

sockeye/inference.py
fhieber: I think making this a member for `Translator` is reasonable, just like `restrict_lexicon`. Then you can directly access it it the beam search function, no?

something like:
```
if output-handler-type == `beam_visualization`:
  output_handler = <implementation to handle dumping histories to json>

translator = Translator(..., store_beam=output-handler-type == `beam_visualization`,...)

File: sockeye/inference.py [hunk 6]
```diff
@@ -1089,22 +1108,55 @@ def _beam_search(self,
...
 
-            # (7) update models' state with winning hypotheses (ascending)
+            # (8) update models' state with winning hypotheses (ascending)
             for ms in model_states:
                 ms.sort_state(best_hyp_indices)
 
-        return sequences, attentions, scores_accumulated, lengths
+        # count the number of steps in each beam
+        for history in beam_histories:
+            history["number_steps"] = len(history["predicted_tokens"])
```
fhieber: isn't that exactly the `t` variable from beam search?

sockeye/inference.py
```diff
@@ -1089,22 +1108,55 @@ def _beam_search(self,
...
+
+        # TODO hussed: delete once handled in the output handlers
+        import json
+        with open("iwslt_s16_test.json", "a") as f:
```
fhieber: this should be moved to an output handler

sockeye/inference.py
Deseaus: Sounds good, will do that then!

sockeye/inference.py
Deseaus: Each history corresponds to a sentence. In the batching case, this code ensures that `"number_steps"` is right for each sentence within a batch. `t` would just give the right size for the longest sentence in the batch.

fhieber: So you want the length without padding tokens? Then you can use the `lengths` array in the function maybe.

Deseaus:
I would still need to know if it's ok licence-wise to add my modified version of VisTools to `sockeye/contrib`, if so I can add it in the next iteration of the PR.

sockeye/inference.py
Deseaus: I think Matt was turning some of these namedtuples into classes in his PR, but this one was kept. Is it still a concern that it's not a class?

fhieber:
The VisTools don't have to be necessarily part of this PR, which is mostly concerned with logging the histories in beam search.

sockeye/inference.py
fhieber: I think its fine leaving Translation as a NamedTuple for now. But I think typing the beam history or making it a proper class would be a good idea. 
Also, there is a typo in the docstring: contian -> contain

fhieber: On a second thought ( I didn't see that you pushed an update to this PR): I think it would be easier to make this a proper class because then you can set beam_histories in the constructor to `None` by default and you dont have to pass it everywhere in the tests and in the rest of the code.

Deseaus:
Thanks, in that case I'd consider this feature complete from my side. Just updated the commit, seems github doesn't like it when I just `git commit --amend` the previous commit :)

fhieber:
We will squash and merge anyway so feel free to push  separate commits, for example for fixing unit tests and code checkers (mypy, pylint) :)

sockeye/inference.py
```diff
@@ -639,6 +666,7 @@ def _concat_translations(translations: List[Translation], start_id: int, stop_id
     # Concatenation of all target ids without BOS and EOS
     target_ids = [start_id]
     attention_matrices = []
+    beam_histories = []
```
fhieber: this likely needs type annotation to satisfy mypy

sockeye/inference.py
```diff
@@ -695,13 +724,15 @@ def __init__(self,
                  models: List[InferenceModel],
                  vocab_source: Dict[str, int],
                  vocab_target: Dict[str, int],
-                 restrict_lexicon: Optional[lexicon.TopKLexicon] = None) -> None:
+                 restrict_lexicon: Optional[lexicon.TopKLexicon] = None,
+                 store_beam : Optional[bool] = False) -> None:
```
fhieber: missing :param description in docstirng above

sockeye/inference.py
```diff
@@ -1051,6 +1094,7 @@ def _beam_search(self,
                 scores = scores[:1] / self.length_penalty(lengths[:1])
             else:
                 # renormalize scores by length ...
+
```
fhieber: you added a newline here :)

sockeye/inference.py
```diff
@@ -1127,5 +1193,9 @@ def _get_best_from_beam(self,
             # attention_matrix: (target_seq_len, source_seq_len)
             attention_matrix = np.stack(attention_lists[idx].asnumpy()[:length, :], axis=0)
             score = accumulated_scores[idx].asscalar()
-            result.append(Translation(sequence, attention_matrix, score))
+            if beam_histories:
```
fhieber: should this check for None instead?

sockeye/output_handler.py
```diff
@@ -267,3 +270,32 @@ def handle(self,
...
+               t_input: inference.TranslatorInput,
+               t_output: inference.TranslatorOutput,
+               t_walltime: float = 0.):
+        """
+        :param t_input: Translator input.
+        :param t_output: Translator output.
+        :param t_walltime: Total wall-clock time for translation.
+        """
+        # If the sentence was max_len split, we may have more than one history
+        for h in t_output.beam_histories:
```
fhieber: this will fail if beam_histories is None. Please add a None check.

Deseaus: I would say this is fine if it fails. If this handler is used and no beam is there it's a bug in the code that should be fixed :)

fhieber: ok

Deseaus:
Ok, the travis checks finally pass :)

fhieber:
Cool, looks good from my side, but lets have someone else (@tdomhan @davvil @mjdenkowski ) take a look.

One thing that is still missing is updating the minor version in sockeye/__init__.py and providing an entry in the CHANGELOG.md

Deseaus:
Done!

Deseaus:
Rebased on the latest master.

sockeye/inference.py
```diff
@@ -857,13 +892,15 @@ def __init__(self,
                  models: List[InferenceModel],
                  source_vocabs: List[vocab.Vocab],
                  target_vocab: vocab.Vocab,
-                 restrict_lexicon: Optional[lexicon.TopKLexicon] = None) -> None:
+                 restrict_lexicon: Optional[lexicon.TopKLexicon] = None,
+                 store_beam : Optional[bool] = False) -> None:
```
davvil: That's not really an Optional, it's initialized to False.

sockeye/inference.py
```diff
@@ -813,6 +841,12 @@ def _concat_translations(translations: List[Translation], start_id: int, stop_id
             else:
                 target_ids.extend(translation.target_ids[1:])
                 attention_matrices.append(translation.attention_matrix[1:, :])
+        if translation.beam_history:
+            # Make a list of the individual beam histories
+            beam_histories.append(translation.beam_history[0])
+
+    if not beam_histories:
```
davvil: Could we go forward with an empty list here? These two lines look a bit confusing at first.

sockeye/output_handler.py
davvil: An assert would be in place, then.

Deseaus:
Adressed comments, updated version and rebased.

PR #286, opened 2018-01-31
log epoch in metrics file and tensorboard

fhieber:
until you can check this box.

### DIFF
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -684,8 +684,8 @@ def write_metrics_file(metrics: List[Dict[str, Any]], path: str):
     """
     with open(path, 'w') as metrics_out:
         for checkpoint, metric_dict in enumerate(metrics, 1):
-            metrics_str = "\t".join(["%s=%s" % (name, str(value)) for name, value in sorted(metric_dict.items())])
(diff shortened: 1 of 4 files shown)


sockeye/utils.py
```diff
@@ -684,7 +684,7 @@ def write_metrics_file(metrics: List[Dict[str, Any]], path: str):
     """
     with open(path, 'w') as metrics_out:
         for checkpoint, metric_dict in enumerate(metrics, 1):
-            metrics_str = "\t".join(["%s=%.6f" % (name, value) for name, value in sorted(metric_dict.items())])
+            metrics_str = "\t".join(["%s=%s" % (name, str(value)) for name, value in sorted(metric_dict.items())])
```
davvil: Why this conversion to string? Because epoch is an int? I think it would look cleaner with a format() call.

fhieber: with string conversion each type is printed more nicely. Ints without a digit and floats with as many digits as needed.

davvil: "{}={}".format(name, value) would do the same. It's nitpicking, but I prefer the syntax.

fhieber: I typically prefer the old-style format strings, I find the `{}`'s hard to read. Will change though.

PR #287, opened 2018-01-31
[WIP] draft of float16 dtype support in training/inference

logogin:
This is a first draft of complete training/inference with `float16` support. We can keep discussion and further updates within this thread.
For training use `--dtype <float16|float32>` (`float32` is default).
For inference, encoder and decoder may have different dtypes. For encoder `--encoder-dtype <float16|float32>`, and for decoder `--decoder-dtype <float16|float32>` respectively.

### Known issues/limitations:
* Almost all operations and symbols (except `mx.sym.FullyConnected` and activations) seems to be 2-3 times slower in `float16` than `float32`.
* `mx.sym.batch_dot` has no support for `float16` in upstream mxnet. Patch will be provided by @KellenSunderland at some moment.
* Only `mx.sym.FullyConnected` seems to have proper support for tensor cores. Other symbols that may require matrix multiplication should be *simulated* via FC layer, i.e.`mx.nd.norm`. 
* `mx.sym.Softmax` numerical stability is limited for `float16`.
* Embeddings should run in `int32`, since `float16` has only integers up to ±2K.
* No unit tests and style checks. Tests may fail during the build.

### Open design questions
* `dtype` can be passed between app layers as numpy type or as plain string. Typed object has advantages as it encapsulates more information and internally all string type names get  converted into `numpy.dtype` at some moment. However, string name is easier to store in config files and seems like neither numpy api nor mxnet has problems with accepting type as plain string.
* Transfering `dtype` from call to call as argument pollutes the code. It could be more effective to store it once in a global config object. However encoders and decoders have polymorphic implementations. Should `dtype` be a singleton across all encoders and decoded or every instance can be configured with its *own* `dtype`?
* What should be the contract between training and inference `dtype`? Currently I can suggest next one:
1. During training `dtype` is stored in config file.
2. For inference `dtype` is read from config.
3. User may override `dtype` for inference with cli arguments.
What about `dtype` for *resumed* training?

### Testing/profiling
There are `test/translate.py` and `test/train.py` scripts which bootstrap minimal inference and training. Models can be used from [tutorials](https://github.com/awslabs/sockeye/tree/master/tutorials) for seq2seq and WMT data.
Those scripts can be used to debug/profile simple cases. Uncomment/modify necessary parts to adjust for your workspace or use-cases.
To dump parameters from layer to layer, uncomment monitors in `inference.py` line 163,168.
`test/compare_params.py` can be used to plot precision drifts between layers.


logogin:
Comments from #281 can be taken into consideration for open questions discussion.

sockeye/training.py
```diff
@@ -41,9 +41,23 @@
...
+# no fp16 support for mx.nd.norm
+def sym_norm(arr):
+    result = norm_symbol.eval(ctx=mx.gpu(0), x=arr.expand_dims(axis=0))[0]
+    return result.reshape((1,))
+
+
 def global_norm(ndarrays: List[mx.nd.NDArray]) -> float:
     # accumulate in a list, as asscalar is blocking and this way we can run the norm calculation in parallel.
-    norms = [mx.nd.square(mx.nd.norm(arr)) for arr in ndarrays if arr is not None]
+    norms = [sym_norm(arr) for arr in ndarrays if arr is not None]
```
fhieber: why use the symbolic API here? Binding an executor every single time seems rather expensive.

KellenSunderland:
I was wondering if it would make sense for us to break the embeddings into a different PR.  Should be mergable to master independently of the dtype work right?

sockeye/training.py
logogin: Symbol gets created once - it is static. But in general, this hack should go away as soon as mxnet will have proper matrix multiplication for float16.

fhieber: True, the symbol only exists once, but you still bind it every single time. What's the advantage over the imparative API here?

logogin: At the moment, this is the only way to compute `x*x` with float16.

fhieber: This does not work?
```
def imperative_norm(arr):
  arr = mx.nd.flatten(arr)
  return mx.nd.FullyConnected(data=arr, weight=arr, no_bias=True, flatten=True, num_hidden=1)

imperative_norm(mx.nd.array([[2,2]], dtype='float16'))
```

fhieber:
@logogin  this PR seems rather out-of-date. Do you want to keep it open or can we close this for now?

logogin:
I think, we can close it for now. Some of the changes were not back-ported yet. But I can track those in my repo.

PR #291, opened 2018-02-02
Fix numpy warning in test_coverage_attention

fhieber:
This removes a warning when running pytest:
```
test/unit/test_attention.py::test_coverage_attention[tanh-4]
 [...]/test/unit/test_attention.py:124: RuntimeWarning: divide by zero encountered in true_divide
    expected_probs = np.ones_like(source_length_vector) /source_length_vector.reshape((batch_size, 1))

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

until you can check this box.

### DIFF
--- a/test/unit/test_attention.py
+++ b/test/unit/test_attention.py
@@ -15,9 +15,9 @@ import mxnet as mx
 import numpy as np
 import pytest
 
-import sockeye.rnn_attention
 import sockeye.constants as C
 import sockeye.coverage
+import sockeye.rnn_attention
 from test.common import gaussian_vector, integer_vector
 
 attention_types = [C.ATT_BILINEAR, C.ATT_DOT, C.ATT_DOT_SCALED, C.ATT_LOC, C.ATT_MLP]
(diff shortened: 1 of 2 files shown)


test/unit/test_attention.py
```diff
@@ -119,7 +119,7 @@ def test_coverage_attention(attention_coverage_type,
     attention_prob_result = exec_output[1].asnumpy()
     dynamic_source_result = exec_output[2].asnumpy()
 
-    expected_probs = (1 / source_length_vector).reshape((batch_size, 1))
+    expected_probs = (1. / source_length_vector).reshape((batch_size, 1))
```
mjpost: I think this kind of thing is deprecated in Python 3. It's no longer necessary since floating point division is the default, and you have to force integer division with `//`.

fhieber: Sure but it also doesn't hurt right? Anyway, source_length_vector is a numpy array so I am not sure the change in Python3 would also apply to numpy.

fhieber:
@mjpost What exactly should I test here? The fact that lengths are now always > 0? I think thats fine since we never have 0-length data in Sockeye anyway.

PR #293, opened 2018-02-09
Data cleanup (towards source_factors PR)

fhieber:
_This is a PR against the source factors branch_ to include another backwards-incompatible change:
Separates DataConfig into DataConfig and DataInfo.
DataInfo contains only stuff that is relevant for training (and prepared data folders). It is saved to a file called `data.info`.
DataConfig is part of ModelConfig and contains values relevant for inference.

until you can check this box.
- [ ] Updated CHANGELOG.md


sockeye/data_io.py
```diff
@@ -556,13 +556,17 @@ def prepare_data(source_fnames: List[str],
...
-                             num_shards=num_shards,
-                             data_statistics=data_statistics,
+    data_info = DataInfo(sources=[os.path.abspath(fname) for fname in source_fnames],
+                         target=os.path.abspath(target_fname),
+                         source_vocabs=source_vocab_paths,
+                         target_vocab=target_vocab_path,
+                         shared_vocab=shared_vocab,
+                         num_shards=num_shards)
+    data_info_fname = os.path.join(output_prefix, C.DATA_INFO)
+    logger.info("Writing data config to '%s'", data_info_fname)
```
tdomhan: technically "writing data info..." ;)

sockeye/data_io.py
```diff
@@ -909,21 +919,30 @@ def __init__(self,
...
+
+    def __init__(self,
+                 data_statistics: DataStatistics,
+                 max_seq_len_source: int,
+                 max_seq_len_target: int) -> None:
+        super().__init__()
         self.data_statistics = data_statistics
         self.max_seq_len_source = max_seq_len_source
         self.max_seq_len_target = max_seq_len_target
-        self.num_source_factors = len(sources)
```
tdomhan: wouldn't it make sense to keep the num_source_factors here, as it's something that is necessary at inference time?

sockeye/data_io.py
fhieber: fixed

sockeye/data_io.py
fhieber: At inference time we currently get them from the embedding config:
```
@property
    def num_source_factors(self) -> int:
        """
        Returns the number of source factors of this InferenceModel (at least 1).
        """
        return self.config.config_embed_source.num_factors
```
Which one do you prefer?

sockeye/data_io.py
```diff
@@ -898,9 +908,9 @@ def describe_data_and_buckets(data_statistics: DataStatistics, bucket_batch_size
                         bucket_batch_size.average_words_per_batch)
 
 
-class DataConfig(config.Config):
+class DataInfo(config.Config):
     """
-    Stores data paths from training.
+    Stores training data information that is not relevant for inference
```
davvil: Missing period at the end.

sockeye/data_io.py
davvil: I think it is cleaner in DataConfig, but it doesn't really make a difference...

fhieber: agreed, I changed it

sockeye/train.py
```diff
@@ -604,7 +607,7 @@ def create_model_config(args: argparse.Namespace,
     config_decoder = create_decoder_config(args, encoder_num_hidden)
 
     source_factor_configs = None
-    if config_data.num_source_factors > 1:
+    if len(source_vocab_sizes) > 1:
```
tdomhan: it probably makes sense to add a check for the args.source_factors_num_embed being of equal size to source_factor_vocab_sizes.

PR #294, opened 2018-02-09
Do not throw exceptions while translating

fhieber:
Ensures that the translate CLI does not throw exceptions but instead logs errors.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -710,7 +710,10 @@ Translation = NamedTuple('Translation', [
     ('score', float)
 ])
 
-EMPTY_TRANSLATION = Translation(target_ids=[], attention_matrix=np.asarray([[0]]), score=-np.inf)
+
+def empty_translation() -> Translation:
+    return Translation(target_ids=[], attention_matrix=np.asarray([[0]]), score=-np.inf)
+
 
 TranslatedChunk = NamedTuple('TranslatedChunk', [
     ('id', int),
@@ -920,11 +923,11 @@ class Translator:
 
             # bad input
             if isinstance(trans_input, BadTranslatorInput):
-                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=EMPTY_TRANSLATION))
+                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=empty_translation()))
 
             # empty input
             elif len(trans_input.tokens) == 0:
-                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=EMPTY_TRANSLATION))
+                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=empty_translation()))
 
             # oversized input
             elif len(trans_input.tokens) > self.max_input_length:
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -857,7 +857,7 @@ class Translator:
         self.vocab_target_inv = vocab.reverse_vocab(self.vocab_target)
         self.restrict_lexicon = restrict_lexicon
         self.start_id = self.vocab_target[C.BOS_SYMBOL]
-        utils.check_condition(C.PAD_ID == 0, "pad id should be 0")
+        assert C.PAD_ID == 0, "pad id should be 0"
         self.stop_ids = {self.vocab_target[C.EOS_SYMBOL], C.PAD_ID}  # type: Set[int]
         self.models = models
         self.interpolation_func = self._get_interpolation_func(ensemble_mode)
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -654,7 +654,7 @@ def make_input_from_factored_string(sentence_id: int,
         pieces = token.split(delimiter)
 
         if not all(pieces) or len(pieces) != model_num_source_factors:
-            logger.error("Failed to parse %s factors at position %d ('%s') in '%s'" % (model_num_source_factors,
+            logger.error("Failed to parse %d factors at position %d ('%s') in '%s'" % (model_num_source_factors,
                                                                                        token_id, token,
                                                                                        factored_string.strip()))
             return _bad_input(sentence_id)
@@ -994,7 +994,8 @@ class Translator:
             factors = trans_input.factors if trans_input.factors is not None else []
             num_factors = 1 + len(factors)
             if num_factors != self.num_source_factors:
-                logger.warning("Input has not enough factors, %d, but expected %d",num_factors, self.num_source_factors)
+                logger.warning("Input has not enough factors, %d, but expected %d", num_factors,
+                               self.num_source_factors)
             for i, factor in enumerate(factors[:self.num_source_factors - 1], start=1):
                 # fill in as many factors as there are tokens
                 source[j, :num_tokens, i] = data_io.tokens2ids(factor, self.source_vocabs[i])[:num_tokens]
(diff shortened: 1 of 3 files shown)


sockeye/inference.py
```diff
@@ -652,15 +670,19 @@ def make_input_from_multiple_strings(sentence_id: int, strings: List[str]) -> Tr
...
     Returns a TranslatorInput object from multiple strings, where the first element corresponds to the surface tokens
     and the remaining elements to additional factors. All strings must parse into token sequences of the same length.
+
     :param sentence_id: An integer id.
     :param strings: A list of strings representing a factored input sequence.
     :return: A TranslatorInput.
     """
-    utils.check_condition(bool(strings), "List of strings cannot be empty.")
+    if not bool(strings):
+        return TranslatorInput(sentence_id=sentence_id, tokens=[], factors=None)
```
davvil: TranslatorInput or _bad_input()?

sockeye/inference.py
```diff
@@ -688,6 +710,8 @@ def make_input_from_multiple_strings(sentence_id: int, strings: List[str]) -> Tr
     ('score', float)
 ])
 
+EMPTY_TRANSLATION = Translation(target_ids=[], attention_matrix=np.asarray([[0]]), score=-np.inf)
```
davvil: Wouldn't it be better to create a new object (similar to _bad_input() above). It doesn't happen at the moment AFAIK, but if at some point the returned object is modified we can get unexpected results (objects are passed by reference).

sockeye/inference.py
davvil: Update: This corresponds to the case where the input is indeed empty, right? Then it should be ok (is also handled with EMPTY_TRANSLATION below).

sockeye/inference.py
```diff
@@ -952,20 +982,19 @@ def _get_inference_input(self, trans_inputs: List[TranslatorInput]) -> Tuple[mx.
         """
         bucket_key = data_io.get_bucket(max(len(inp.tokens) for inp in trans_inputs), self.buckets_source)
 
-        utils.check_condition(C.PAD_ID == 0, "pad id should be 0")
```
davvil: This could be an assert.

sockeye/inference.py
```diff
@@ -952,20 +982,19 @@ def _get_inference_input(self, trans_inputs: List[TranslatorInput]) -> Tuple[mx.
...
-                utils.check_condition(len(factor) == num_tokens,
-                                      "Factor sequence has different length than tokens: %s" % trans_input)
-                source[j, :num_tokens, i + 1] = data_io.tokens2ids(factor, self.source_factor_vocabs[i])
+            source[j, :num_tokens, 0] = data_io.tokens2ids(trans_input.tokens, self.source_vocabs[0])
+
+            factors = trans_input.factors if trans_input.factors is not None else []
+            num_factors = 1 + len(factors)
+            if num_factors != self.num_source_factors:
+                logger.warning("Input has not enough factors, %d, but expected %d",num_factors, self.num_source_factors)
+            for i, factor in enumerate(factors[:self.num_source_factors - 1], start=1):
```
davvil: Wouldn't this fail if num_factors < self.num_source_factors?

sockeye/inference.py
fhieber: I guess this is a case where the input is indeed empty so its fine as is.

sockeye/inference.py
fhieber: Probably better yes, will change.

sockeye/inference.py
fhieber: I moved it to the constructor. We anyway require this throughout the project so i wanted to remove the assert being called all the time.

sockeye/inference.py
fhieber: This is fine in python slicing:
```
In [2]: list(range(2))[:14]
Out[2]: [0, 1]
```

sockeye/inference.py
davvil: Ah! Hadn't seen it. Nevertheless this should be an assert, as it is a programming error and not a user error.

sockeye/inference.py
davvil: Right, that is slicing and not accessing!

sockeye/inference.py
tdomhan: it was already a check_condition before, but I do agree, we should convert to the assert now that we're touching the code.

fhieber: done (but I think you were referring to the other comment below :))

sockeye/inference.py
```diff
@@ -579,6 +580,17 @@ def chunks(self, chunk_size: int) -> Generator['TranslatorInput', None, None]:
...
 
 
+class BadTranslatorInput(TranslatorInput):
+
+    def __init__(self, sentence_id, tokens):
+        super().__init__(sentence_id=sentence_id, tokens=tokens, chunk_id=-1, factors=None)
+
+
+def _bad_input(sentence_id: int) -> BadTranslatorInput:
+    logger.warning("Bad input (%d) will return empty output.", sentence_id)
```
tdomhan: it would probably make sense to log the actual input rather than the sentence id, so that one could rerun a bad input from the log file.

sockeye/inference.py
```diff
@@ -639,8 +652,13 @@ def make_input_from_factored_string(sentence_id: int,
     factors = [[] for _ in range(model_num_source_factors - 1)]  # type: List[Tokens]
     for token_id, token in enumerate(data_io.get_tokens(factored_string)):
         pieces = token.split(delimiter)
-        utils.check_condition(all(pieces) and len(pieces) == model_num_source_factors,
-                              "Failed to parse %s factors from token '%s'" % (model_num_source_factors, token))
+
+        if not all(pieces) or len(pieces) != model_num_source_factors:
+            logger.error("Failed to parse %s factors at position %d ('%s') in '%s'" % (model_num_source_factors,
```
tdomhan: the format string seems of. model_num_source_factors is an int, no? in the format string you have `%s`

sockeye/inference.py
fhieber: I am doing this in the individual functions because its hard to decide what the logged string should be (json, multiple stirngs, factored string).

sockeye/inference.py
fhieber: fixed

sockeye/inference.py
tdomhan: make_input_from_json_string actually does not log the translation as far as I can see. Alternatively could have a field `BadTranslatorInput.reason` where you fill in a string from each one of the methods and then just do the logging of the entire input string along with the reason one level up.

sockeye/inference.py
```diff
@@ -598,19 +610,20 @@ def make_input_from_json_string(sentence_id: int, json_string: str) -> Translato
...
+        factors = jobj.get(C.JSON_FACTORS_KEY)
+        if isinstance(factors, list):
+            factors = [list(data_io.get_tokens(factor)) for factor in factors]
+        else:
+            factors = None
+        return TranslatorInput(sentence_id=sentence_id, tokens=tokens, factors=factors)
 
-    return TranslatorInput(sentence_id=sentence_id, tokens=tokens, factors=factors)
+    except Exception as e:
+        logger.exception(e, exc_info=True) if not is_python34() else logger.error(e)  # type: ignore
```
tdomhan: this is what I was referring to: here the string is not logged.

fhieber: changed

PR #307, opened 2018-03-01
Use stack instead of expand_dims, concat for memory efficiency.

fhieber:
This is a memory improvement for RNN decoders.
As stated in https://github.com/apache/incubator-mxnet/pull/9934 the use of `mx.sym.stack` is more memory efficient than `mx.sym.expand_dims` followed by `mx.sym.concat`.
I ran a quick test between these two patterns by stacking/concatenating an array of size (128, 100, 512) 100 times, and the memory used on the GPU went from 5249MB to 2736MB when using `mx.sym.stack`.

I searched through the codebase if there are other cases for this, but the combination of RNN decoder hidden states over time seems to be the only place where this is relevant. Note that this only applies to cases where you want to concatenate on a separate axis. 

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/encoder.py
+++ b/sockeye/encoder.py
@@ -455,7 +455,7 @@ class AddSinCosPositionalEmbeddings(PositionalEncoder):
         sin = mx.sym.sin(scaled_positions)
         cos = mx.sym.cos(scaled_positions)
 
-        # (batch_size, num_embed/2)
+        # (batch_size, num_embed)
         pos_embedding = mx.sym.concat(sin, cos, dim=1)
 
         if self.scale_up_input:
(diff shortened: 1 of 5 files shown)


sockeye/encoder.py
```diff
@@ -455,7 +455,7 @@ def encode_positions(self,
         sin = mx.sym.sin(scaled_positions)
         cos = mx.sym.cos(scaled_positions)
 
-        # (batch_size, num_embed/2)
+        # (batch_size, num_embed)
```
davvil: Was that a mistake earlier? It shouldn't change with this PR, should it?

fhieber: yeah the comment was wrong

PR #311, opened 2018-03-02
[1.18] restrict-lexicon: Allow specification of topk at runtime.

fhieber:
This allows selecting the number of elements considered for each source word at runtime, if the `--restrict-lexicon` functionality is used.
- Added a new flag `--restrict-lexicon-topk` that controls this. If the lexicon on disk contains fewer elements than this value, a warning is given and the maximum number of elements from the lexicon on disk is used.
- Changed serialization of TopKLexicon to numpy arrays as this simplifies the logic. @mjdenkowski in order to allow the selection at runtime, I removed the sorting by trg_ids. The lexicon rows are now sorted by decreasing logprob. As far as I can tell this is ok.
- Changed lexicon CLI to expose two subcommands: "create" and "inspect". The former behaves as the previous CLI. The latter allows inspecting an existing lexicon by passing input tokens to STDIN. The CLI then lists the set of unique target tokens allowed by that top-k lexicon.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ### Changed
 - Change default target vocab name in model folder to `vocab.trg.0.json`
 - Changed serialization format of top-k lexica to pickle/Numpy instead of JSON.
+- `sockeye-lexicon` now supports two subcommands: create & inspect.
+  The former provides the same functionality as the previous CLI.
+  The latter allows users to pass source words to the top-k lexicon to inspect the set of allowed target words.
 
 ### Added
 - Added ability to choose a smaller `k` at decoding runtime for lexicon restriction.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ## [1.18.0]
 ### Changed
 - Change default target vocab name in model folder to `vocab.trg.0.json`
+- Changed serialization format of top-k lexica to pickle/Numpy instead of JSON.
+
+### Added
+- Added ability to choose a smaller `k` at decoding runtime for lexicon restriction.
 
 ## [1.17.4]
 ### Added
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -1033,8 +1033,13 @@ def add_inference_args(params):
     decode_params.add_argument('--restrict-lexicon',
                                type=str,
                                default=None,
-                               help="Specify top-k lexicon to restrict output vocabulary based on source.  See lexicon "
+                               help="Specify top-k lexicon to restrict output vocabulary based on source. See lexicon "
                                     "module. Default: %(default)s.")
+    decode_params.add_argument('--restrict-lexicon-topk',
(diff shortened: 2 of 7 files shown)


tdomhan:
The change to the format is orthogonal to exposing the topk parameter, right? I'm not sure that changing the serialization format is worth another change in the major version. Could we separate the two things and put the format change on a list of things for the next major version?

fhieber:
Actually its not orthogonal. The previous serialization format was storing only the src-trg pairs in an (unordered) dictionary. So if you de-serialize this you have no way of selecting a smaller top-k from this. So we could either extend the dictionary to store the probabilities, or change to store the numpy matrix with rows ordered by decreasing prob. Both would entail a format/major version change.

tdomhan:
oh ok. makes sense.

mjdenkowski:
In principle this should be fine and make it a little less expensive to run experiments with different values of K.  Sorting by target id was chosen to minimize work for NumPy's set operations that sort arrays as part of each operation.  Have you run speed comparisons?  I don't expect this to make a noticeable difference but we could still sort by value (target id) after loading.

One question: is NumPy's array format human readable?  One reason for going with JSON was that a couple of grep commands would let a human see whether translations allowed for a particular source word made sense.

fhieber:
Thanks for the comments!
I added sorting of each row by increasing target id when loading the lexicon. This should then be fully equivalent to the previous implementation.

Numpy's binary format is not human readable but is the fastest to load. Depending on the size of the lexicon, this can make a difference in startup time of the translator. The other options would be again JSON (but using lists to keep the ordering) or Numpy's savetxt. Both would be readable (though only show the vocab ids of course), but slower to load.
The use case you are mentioning makes sense to me, but grepping and mapping words to ids already requires some expert knowledge. 
What do you think about leaving the numpy binary format and adding a small CLI that takes vocabularies and a serialized lexicon and reads words from stdin? This would be similar to our embeddings CLI.

mjdenkowski:
NumPy binary format with a CLI for checking entries sounds perfect--faster and easier to use than straight JSON.  Otherwise everything looks good!

fhieber:
I modified the lexicon CLI to support this use case. 
Here is some example output when the CLI is used for inspection (creation works the same way, but requires the subcommand `create`):
```
echo "le maison" | python3 -m sockeye.lexicon inspect -l lexicon -m model/ -k 5
[INFO:inspect] Sockeye version 1.18.0 commit unknown
[INFO:inspect] Inspecting top-k lexicon at "lexicon"
[INFO:sockeye.vocab] Vocabulary (30065 words) loaded from "model/vocab.src.0.json"
[INFO:sockeye.vocab] Vocabulary (30065 words) loaded from "model/vocab.trg.json"
[INFO:inspect] Loaded top-5 lexicon from "lexicon".
[INFO:inspect] Reading from STDIN...
Input:  n=2 le(19) maison(3698)
Output: n=14 <pad>(0) <unk>(1) <s>(2) </s>(3) ,(4) of(8) le(19) &apos;&apos;(51) Le(54) M@@(394) 06(4812) aison(8059) EMENT(9004) SOCI@@(28093)
```

fhieber:
@mjdenkowski ping :)

PR #314, opened 2018-03-06
Freeze embeddings

davvil:
Added an option for freezing (input/output) embeddings.
Probably use only in combination with sockeye.init_embedding.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -156,6 +156,19 @@ def multiple_values(num_values: int = 0,
     return parse
 
 
+def string2bool(string : str) -> bool:
+    """
+    Parses a string as a bool value. This function can be used as the
+    'data_type' argument of multiple_values.
+    """
+    if string in ["True", "true", "1"]:
+        return True
+    elif string in ["False", "false", "0"]:
+        return False
+    else:
+        raise argparse.ArgumentTypeError("Could not parse boolean value '%s'" % string)
+
+
 def file_or_stdin() -> Callable:
     """
     Returns a file descriptor from stdin or opening a file from a given path.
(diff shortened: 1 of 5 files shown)


sockeye/arguments.py
```diff
@@ -156,6 +156,19 @@ def parse(value_to_check):
     return parse
 
 
+def string2bool(string : str) -> bool:
```
fhieber: nitpick: extra space between "string" and ":"

sockeye/encoder.py
```diff
@@ -281,13 +281,15 @@ def __init__(self,
                  vocab_size: int,
                  num_embed: int,
                  dropout: float,
+                 frozen: bool = False,
```
fhieber: I find this name a bit confusing (given that we have the hidden member in configs called _frozen). 
But more general, I wonder if we should implement fixing certain parameters during training with the functionality in the Module API (http://mxnet.incubator.apache.org/api/python/module/module.html) and its member `fixed_param_names`, which right now defaults to None.
This would allow a more general way of skipping optimization for certain parameter sets.

davvil: My original name was "freeze" which lead to a funny error message about bool not being callable... I'm open to naming suggestions.

fixed_param_names is also a good way to go (I used it in another branch), but we'll probably should discuss the user interface for it.

fhieber: How about the simplest interface possible, an additional CLI argument (next to `--allow-missing-params` that takes a list of strings that correspond to the names of the parameters to be fixed during training. We then simply pass this to the Module constructor. This is transparent, does not touch the symbol generating code (encoder.py) but is limited to the training code, and users only need to know the parameter names, which they can get easily by starting the training and looking at the log.

fhieber:
closing in favor of #320

PR #315, opened 2018-03-06
rnn_attention config and factory improvements

logogin:
Prerequisite for #287. 
`dtype` in `AttentionConfig`. `get_attention` factory has been improved to utilise OO approach.

until you can check this box.
- [ ] Updated CHANGELOG.md


tdomhan:
this seems to include the dtype change, is this intended?

logogin:
Updated PR

sockeye/rnn_attention.py
```diff
@@ -49,7 +52,8 @@ def __init__(self,
                  query_num_hidden: int,
                  layer_normalization: bool,
                  config_coverage: Optional[coverage.CoverageConfig] = None,
-                 num_heads: Optional[int] = None) -> None:
+                 num_heads: Optional[int] = None,
+                 dtype: Optional[str] = C.DTYPE_FP32) -> None:
```
tdomhan: is `dtype` indeed `Optional`, meaning it can be `None` given that you have a default `str` argument? You use it below in `np.finfo`, which defaults to `float64` when given None on my system.

sockeye/rnn_attention.py
```diff
@@ -274,20 +295,20 @@ class DotAttention(Attention):
...
     """
 
     def __init__(self,
                  input_previous_word: bool,
                  source_num_hidden: int,
                  query_num_hidden: int,
                  num_hidden: int,
-                 scale: Optional[float] = None) -> None:
-        super().__init__(input_previous_word)
+                 dtype: Optional[str] = C.DTYPE_FP32) -> None:
```
tdomhan: why not keep scale as an optional parameter defaulting to None?

sockeye/rnn_attention.py
```diff
@@ -343,33 +364,65 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
...
+                 query_num_hidden: int,
+                 num_hidden: int,
+                 scale: Optional[float] = None,
+                 dtype: Optional[str] = C.DTYPE_FP32) -> None:
+        super().__init__(input_previous_word,
+                         source_num_hidden,
+                         query_num_hidden,
+                         num_hidden,
+                         dtype=dtype)
+        self.scale = scale if scale else num_hidden ** -0.5
```
tdomhan: if you kept scale as a parameter on `DotAttention` you could pass it directly in the constructor.

sockeye/rnn_attention.py
logogin: fixed

sockeye/rnn_attention.py
logogin: As discussed, `scale` parameter will be now supported in `DotScaleAttantion`, but not `DotAttention`

sockeye/rnn_attention.py
```diff
@@ -59,6 +63,26 @@ def __init__(self,
         self.layer_normalization = layer_normalization
         self.config_coverage = config_coverage
         self.num_heads = num_heads
+        self.dtype = dtype
+
+
+def _instantiate(cls, params):
+    """
+    Helper to instantiate Attention classes from parameters. Warns in log if parameter is not supported
```
fhieber: Not really specific to Attention is it? Would it make sense to move this to utils.py and generalize the doc string (and the additional args you pass, e.g. max_seq_len)?

sockeye/rnn_attention.py
```diff
@@ -59,6 +63,26 @@ def __init__(self,
...
+    by class constructor.
+
+    :param cls: Attention class
```
fhieber: [nit] missing period at the end of param/return descriptions.

sockeye/rnn_attention.py
```diff
@@ -196,6 +214,7 @@ def make_input(self,
         return AttentionInput(seq_idx=seq_idx, query=query)
 
 
+@Attention.register(C.ATT_BILINEAR)
```
fhieber: I love that!

sockeye/rnn_attention.py
fhieber: I agree with Tobi's comment. Its confusing to set self.scale to None in `DotAttention` just to have inheritance. I would be in favor of a single `DotAttention` class that takes a scale parameter (that can be None).

sockeye/rnn_attention.py
```diff
@@ -608,7 +666,8 @@ def on(self, source: mx.sym.Symbol, source_length: mx.sym.Symbol, source_seq_len
         :return: Attention callable.
         """
 
-        coverage_func = self.coverage.on(source, source_length, source_seq_len) if self.coverage else None
+        if self.coverage:
```
fhieber: explicit check for None

fhieber: don't you need to set coverage_func to None if self.coverage is None? I can imagine that the style checkers would complain about possibly uninitialized variables.

sockeye/rnn_attention.py
```diff
@@ -689,16 +748,46 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
...
 
+@Attention.register(C.ATT_COV)
+class MlpCovAttention(MlpAttention):
+    """
+    MlpAttention with optional coverage config.
+
+    :param input_previous_word: Feed the previous target embedding into the attention mechanism.
+    :param num_hidden: Number of hidden units.
+    :param layer_normalization: If true, normalizes hidden layer outputs before tanh activation.
+    :param config_coverage: Optional coverage config.
```
fhieber: If you really want to have this string inheritance (MLPAttention -> MLPCovAttention), then the coverage config is not optional here.

sockeye/rnn_attention.py
logogin: no style check likes that :)

sockeye/rnn_attention.py
fhieber: sure, fine

sockeye/rnn_attention.py
logogin: fixed

sockeye/rnn_attention.py
logogin: changed

sockeye/rnn_attention.py
fhieber: I would leave it as is: scaled dot attention is still conceptually dot attention, so it makes sense to have a single class DotAttention that can either scale or not scale. The responsibility of these Attention classes is to compute attention scores in a certain way. The scaling is a minor detail that is well defined within the responsibility of the dot attention class and reduces code complexity.
The question is how we want to expose this to the user. Right now there is "dot" and "dot-scaled" which was added quickly by me without thinking about it very long.  Maybe it might make sense to remove "dot-scaled" and allow configuring the "dot" attention to scale or not via the command line (for example by adding another flag `--rnn-scale-dot-attention` or something)?
Another way could be to annotate the DotAttention class with two registered names? Is it possible to have a 2:1 mapping with this new type of instantiation?
@tdomhan what's your opinion?

tdomhan: conceptually having a single DotAttention where the scale can optionally be given would make the most sense. The only downside is that this would actually require a major version bump. If we agree that this the way to go, this should be coordinated with the top-k lexicon PR from Felix to just have a single bump for both features.

sockeye/rnn_attention.py
```diff
@@ -343,33 +364,65 @@ def attend(att_input: AttentionInput, att_state: AttentionState) -> AttentionSta
...
     def __init__(self,
                  input_previous_word: bool,
-                 num_hidden: int,
-                 heads: int) -> None:
-        super().__init__(input_previous_word)
-        utils.check_condition(num_hidden % heads == 0,
-                              "Number of heads (%d) must divide attention depth (%d)" % (heads, num_hidden))
-        self.num_hidden = num_hidden
-        self.heads = heads
+                 source_num_hidden: int,
```
tdomhan: renaming these arguments means that models trained with the previous sockeye version will no longer be compatible. We should either leave the names the same, or bump (which might anyway be necessary if we decide to remove the scaled dot attention)

sockeye/rnn_attention.py
```diff
@@ -707,7 +795,7 @@ def get_context_and_attention_probs(values: mx.sym.Symbol,
                                  axis=1,
                                  use_sequence_length=True,
                                  sequence_length=length,
-                                 value=C.LARGE_NEGATIVE_VALUE)
+                                 value=np.finfo(dtype).min)
```
tdomhan: also technically a reason to bump the major version. Practically it probably does not make a difference. However, if we don't bump this should be confirmed.

tdomhan:
Given what me discussed I think it makes sense to include this in the backwards-incompatible 1.18 release, which is now tracked in the branch `v1.18`. Could you create a PR based on your changes for that branch, when can then merge this change and other changes first into the `v1.18` branch and then into master.

sockeye/rnn_attention.py
logogin: I renamed constructor args, to match `AttentionConfig` state variables. Internal class fields as well as instantiation of `AttentionConfig` are untouched. So the change is backwards-compatible.

sockeye/rnn_attention.py
logogin: I can revert to the original constant to avoid version bump

sockeye/rnn_attention.py
logogin: Let me try once again to argue against single class. As everyone agrees, we have two types of attention, normal and scaled. Scaled requires a parameter, which is by default calculated as `num_hidden ** -0.5`. Now, it is not possible to supply `scale` param via cli arguments. Attention type acts as a flag to substitute this param into `DotAttention` constructor.
So here is a contradiction to the quote:
> The responsibility of these Attention classes is to compute attention scores in a certain way. The scaling is a minor detail that is well defined within the responsibility of the dot attention class and reduces code complexity.

`DotAttention` is not responsible for the all required logic - the factory must pre-compute `scale` param, set it to `non None` only then `DotAttention` will scale. Just having a type is not enough. The complexity grew up. It is spread across two parts of code - class implementation and factory method.
There are dozens of design patterns to solve that: Decorator, Template class/method, Builder etc. Having more arguments in a constructor than less is a wrong direction. Having branching logic based on optional parameters does not reduce the complexity. 
But back to the proposed solutions:
1. **Unfortunately**, two annotations on a single class will work :). But then default scale param logic should be moved back to the factory method.
2. In case you decide to deprecate one of types, the place for `scale` parameter calculation still remains questionable.

If you all agree with drawback for the single class approach - I will go for 1st option to preserve compatibility.

logogin: additionally `is_scaled` flag can be added to `DotAttention` instead of `scale` argument.

sockeye/rnn_attention.py
tdomhan: oh ok. I mean there were anyway some backwards incompatible changes in the pipeline, so that we bump (see below), so either bumping shouldn't be a big issue.

sockeye/rnn_attention.py
tdomhan: as above, I think a version bump is fine.

sockeye/rnn_attention.py
tdomhan: I think the latter approach with the boolean CLI parameter, which I think Felix also argued for with enabling scaling through `--rnn-scale-dot-attention` is the way to go as it's the simplest solution that covers both use cases covered in the previous code versions: If it's not given we don't scale and if it is given we scale be the number of hidden units. Having the logic of how to scale in the constructor of the DotAttention class makes sense to me. It's simple, clean and works :)

logogin:
rebased with v1.18

test/unit/test_arguments.py
```diff
@@ -114,7 +115,52 @@ def test_device_args(test_params, expected_params):
               conv_embed_num_highway_layers=4,
               conv_embed_pool_stride=5,
               conv_embed_add_positional_encodings=False,
-              rnn_attention_in_upper_layers=False))])
+              rnn_attention_in_upper_layers=False)),
+    ('--rnn-scale-dot-attention', dict(params=None,
```
fhieber: can we remove this test? We stripped down all of these test_arguments tests to the default case ('') to not have people have to add arguments all over the place.

logogin: done

PR #304, opened 2018-02-26
Separate TrainingModel into TrainingModel and EarlyStoppingTrainer classes.

fhieber:
- split TrainingModel class into TrainingModel and Trainer class to separate model from fitting logic better
- simplified training resumption logic
- TrainState stores more information about training state
- removed the quite complicated and convoluted callback logic
- restructured fit method in Trainer to make it more readable
- introduced OptimizerConfig to bundle optimizer-related settings
- CheckpointDecoder now returns overall decoding time, not only sent/sec.
- Fixed a bug with gradient norms not being written to tensorboard correctly
- Fixed the same bug as in #302 
- simplified constructors of {Sockeye,Training,Inference}Model

#### Pull Request Checklist ##
- [x] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]'
until you can check this box.
- [x] Unit tests pass (`pytest`)
- [x] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?
- [x] System tests pass (`pytest test/system`)
- [x] Passed code style checking (`./pre-commit.sh` or manual run of pylint & mypy)
- [x] You have considered writing a test
- [x] Updated major/minor version in `sockeye/__init__.py`. Major version bump if this is a backwards incompatible change.
- [x] Updated CHANGELOG.md

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #317, opened 2018-03-08
Bug when loading external parameters for initialization

davvil:
(Bug introduced in #304)

Calling set_params directly after loading the params from file does not
honor the allow_missing flag. I allowed skipping the set_params call and
let the handling of allow_missing to init_params, as it was done
pre-#304. However set_params also accepts an allow_missing parameter. I
am not sure of the difference between the two variants, though,
therefore the [WIP].

We should probably consider adding a test for this.

(description of the change)

- [x] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]' until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -305,7 +305,7 @@ class TrainingModel(model.SockeyeModel):
         self.aux_params = aux_params
         super().save_params_to_file(fname)
 
-    def load_params_from_file(self, fname: str, set_params: bool = True):
+    def load_params_from_file(self, fname: str, allow_missing_params: bool = False):
         """
         Loads parameters from a file and sets the parameters of the underlying module and this model instance.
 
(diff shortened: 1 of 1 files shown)


sockeye/training.py
```diff
@@ -305,16 +305,17 @@ def save_params_to_file(self, fname: str):
         self.aux_params = aux_params
         super().save_params_to_file(fname)
 
-    def load_params_from_file(self, fname: str, set_params: bool = True):
+    def load_params_from_file(self, fname: str, allow_missing_params: bool = False):
         """
         Loads parameters from a file and sets the parameters of the underlying module and this model instance.
 
         :param fname: File name to load parameters from.
         :param set_params: Set the params in the module.
```
fhieber: Update doc string:
":param allow_missing_params: If True, the parameters in fname are allowed to be a subset of the Module parameters."

PR #320, opened 2018-03-09
Fix params

davvil:
Added parameter --fix-params for fixing parameters during training. If approved this supersedes #314.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -895,7 +895,7 @@ def add_training_args(params):
                               choices=[C.RNN_INIT_ORTHOGONAL, C.RNN_INIT_ORTHOGONAL_STACKED, C.RNN_INIT_DEFAULT],
                               help="Initialization method for RNN parameters. Default: %(default)s.")
 
-    train_params.add_argument('--fix-params',
+    train_params.add_argument('--fixed-param-names',
                               default=[],
                               nargs='*',
                               help="Names of parameters to fix at training time. Default: %(default)s.")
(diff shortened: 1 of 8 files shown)


sockeye/model.py
```diff
@@ -60,7 +60,7 @@ def __init__(self,
                  config_loss: loss.LossConfig,
                  weight_tying: bool = False,
                  weight_tying_type: Optional[str] = C.WEIGHT_TYING_TRG_SOFTMAX,
-                 weight_normalization: bool = False) -> None:
+                 weight_normalization: bool = False):
```
fhieber: "-> None" missing

sockeye/arguments.py
```diff
@@ -895,6 +895,11 @@ def add_training_args(params):
                               choices=[C.RNN_INIT_ORTHOGONAL, C.RNN_INIT_ORTHOGONAL_STACKED, C.RNN_INIT_DEFAULT],
                               help="Initialization method for RNN parameters. Default: %(default)s.")
 
+    train_params.add_argument('--fix-params',
```
fhieber: I would vote for calling this "fixed-params" or even "fixed-param-names", which is consistent with what MXNet expects.

sockeye/train.py
```diff
@@ -640,7 +640,8 @@ def create_training_model(config: model.ModelConfig,
                                             provide_label=train_iter.provide_label,
                                             default_bucket_key=train_iter.default_bucket_key,
                                             bucketing=not args.no_bucketing,
-                                            gradient_compression_params=gradient_compression_params(args))
+                                            gradient_compression_params=gradient_compression_params(args),
+                                            fixed_params=args.fix_params)
```
fhieber: I would call both the value and the parameter "fixed_param_names"

sockeye/arguments.py
davvil: I was tempted to do so as well, but I have the feeling that most of our arguments (if appropriate) refer to actions: --restrict-lexicon, --fill-up, --use-tensorboard, etc.

fhieber: True, but to me `--fix-params` sounds like a boolean flag (fix params if this flag is given) and doesn't encourage me to put a list of names after it).

PR #321, opened 2018-03-09
Added --dry-run parameter

davvil:
Initialize models but do not perform any training iteration or write anything to disk.

Useful e.g. to find out parameter names, data sizes, etc.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -921,6 +921,11 @@ def add_training_args(params):
                               default=-1,
                               help='Keep only the last n params files, use -1 to keep all files. Default: %(default)s')
 
+    train_params.add_argument('--dry-run',
+                              action='store_true',
+                              help="Do not perform any actual training, but print statistics about the model"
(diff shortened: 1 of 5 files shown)


fhieber:
I am not sure about this one. Is this actually useful? It seems to touch a lot of places in the code, adding various if statements (to recently cleaned up pieces) and the benefit is not entirely clear to me.

davvil:
The use case for me was finding out parameter names (see #320), but it can be used as a "sanity check" to test if sockeye will do what you expect it to do. Sure, you can always start a training and stop it, but I found adding the option was cleaner.

In the beginning I was hopping that the changes would be minor (ideally only comment out the call to fit()), but it turns out that files are written in several places. That's why it spread out to several files.

No big deal if it doesn't get in, I'll just retreat to a corner and cry in silence ;-)

davvil:
Preferred much less intrusive implementation in #322.

PR #322, opened 2018-03-09
Added option --dry-run (light version)

davvil:
(Nearly) Equivalent to #321 but much less intrusive changes.

If --dry-run is given, the output directory is set to a temporary and max_updates is set to 0.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/train.py
+++ b/sockeye/train.py
@@ -19,6 +19,7 @@ import json
 import os
 import shutil
 import sys
+import tempfile
 from contextlib import ExitStack
 from typing import Any, cast, Optional, Dict, List, Tuple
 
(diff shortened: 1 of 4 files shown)


sockeye/train.py
```diff
@@ -723,6 +724,13 @@ def main():
     arguments.add_train_cli_args(params)
     args = params.parse_args()
 
+    if args.dry_run:
+        # Modify arguments so that we write to a temporary directory and
+        # perform 0 training iterations
+        temp_dir = tempfile.TemporaryDirectory()  # Will be automatically removed
```
fhieber: Are you sure that this gets removed?

davvil: From the documentation: "On completion of the context or destruction of the temporary directory object the newly created temporary directory and all its contents are removed from the filesystem." (https://docs.python.org/3/library/tempfile.html)

I also checked after completion and it was indeed removed.

fhieber: Ok, thanks!

PR #324, opened 2018-03-13
[1.18] Change default target vocab name in model folder to vocab.trg.0.json

fhieber:
until you can check this box.
- [x] Updated CHANGELOG.md


davvil:
I have no issues with the implementation itself, but why do we want it? Just for consistency with the source vocabs?

fhieber:
Mostly for parallelism of vocabulary names. Also, if we would add target factors we may be able to avoid a MV bump.

sockeye/constants.py
```diff
@@ -195,7 +195,8 @@
 JSON_SUFFIX = ".json"
 VOCAB_SRC_PREFIX = "vocab.src"
 VOCAB_SRC_NAME = VOCAB_SRC_PREFIX + ".%d" + JSON_SUFFIX
-VOCAB_TRG_NAME = "vocab.trg" + JSON_SUFFIX
```
mjpost: As long as we're changing this, shouldn't it be ".tgt"?

fhieber: I am a big 'trg' believer :)

Issue #270, opened 2018-01-15
smallest_k_mx just partitions but does not sort


tdomhan:
Writing this done before I forget about it: _beam_search expects the first hypothesis to be the best hypothesis (see `_get_best_from_beam`). The function `utils.smallest_k` guarantees this by sorting the partition returned by numpy. `smallest_k_mx` does not, as it takes the output of mxnet's `topk`. Right now this is not an issue because 1. `smallest_k_mx` is not used and 2. mxnet's `topk` actually sorts. However, we can not rely on a concrete implementation of `topk` and should change `smallest_k_mx` accordingly..

sxjscience:
The implementation in mxnet.topk is to first sort and the take the top k elements and should be fine to use.

tdomhan:
that makes sense. I guess the question is though, whether this is the guaranteed behavior of topk, order whether this is a side effect of the current implementation?

mjpost:
I see. Looking at [the documentation](https://mxnet.incubator.apache.org/api/python/symbol.html#mxnet.symbol.topk), no guarantee is made that the top `k` elements will be sorted, so MXNet's `topk` actually only guarantees something similar to numpy's `argpartitiion()`. This might just be an error with the documentation, though. Perhaps we could get a clarification?

tdomhan:
yeah, so I think we either need to update the documentation around guarantees in mxnet or add a second sorting step in smallest_k_mx just to make sure future changes to topk won't break our beam search.

sxjscience:
@tdomhan It's currently the side effect of the implementation. Sorting after topk is a good solution. Another solution is to combine slice_axis and sort:
```python
out = nd.slice_axis(nd.sort(data, axis=-1), axis=-1, begin=0, end=K)
```

mjpost:
What about just leaving it like it is until the topk() in MXNet is updated? I think our test cases would catch a situation where this gets updated without us noticing. My speed tests (which I'll post tomorrow) show much better performance of topk() on both p2's and p3's, but *much worse* performance on CPU's.

Alternately, we could use @sxjscience's proposal here (slice + sort) and then revert to topk() when it is no longer sorting.

sxjscience:
@mjpost Great! I can later help submit a feature request in MXNet quoting the speed tests.

mjpost:
[...]

PR #333, opened 2018-03-22
Mxnet topk unless cpu

mjpost:
Uses MXNet's `topk()` operator when working on GPU's, where the cost of copying memory off the GPU for use with numpy is more expensive than MXNet's current inefficient sort-based implementation.

As [discussed here](https://github.com/awslabs/sockeye/issues/270#issuecomment-375286651), based on [graph here](https://github.com/awslabs/sockeye/pull/230#issuecomment-375263091).

until you can check this box.
- [ ] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -889,7 +889,10 @@ def __init__(self,
                  source_vocabs: List[vocab.Vocab],
                  target_vocab: vocab.Vocab,
                  restrict_lexicon: Optional[lexicon.TopKLexicon] = None,
-                 store_beam : bool = False) -> None:
+                 store_beam: bool = False,
+                 use_mxnet_topk: bool = False) -> None:
+        self.use_mxnet_topk = use_mxnet_topk
+        self.smallest_k_func = utils.smallest_k_mx if self.use_mxnet_topk else utils.smallest_k
```
fhieber: how about we branch based on `context`? Then we don't need the use_mxnet_topk flag at all (neither as a CLI argument nor in this constructor)

sockeye/inference.py
```diff
@@ -1278,13 +1281,16 @@ def _beam_search(self,
 
             # (3) get beam_size winning hypotheses for each sentence block separately
             # TODO(fhieber): once mx.nd.topk is sped-up no numpy conversion necessary anymore.
-            scores = scores.asnumpy()  # convert to numpy once to minimize cross-device copying
+            if not self.use_mxnet_topk:
```
fhieber: lets branch on context here

mjpost:
Sorry, I branched and then didn't commit my changes before pushing. This addresses basically all your comments :)

sockeye/inference.py
mjpost: See my comment below. I had made all these changes and forgot to commit them before pushing. I think you'll like it now.

sockeye/inference.py
```diff
@@ -1278,13 +1279,16 @@ def _beam_search(self,
 
             # (3) get beam_size winning hypotheses for each sentence block separately
             # TODO(fhieber): once mx.nd.topk is sped-up no numpy conversion necessary anymore.
-            scores = scores.asnumpy()  # convert to numpy once to minimize cross-device copying
+            if self.context == mx.cpu():
+                scores = scores.asnumpy()  # convert to numpy once to minimize cross-device copying
```
tdomhan: minor suggestion: we could avoid this if branch by pulling it into utils.smallest_k, so that both methods expect an mxnet.NDArray.

fhieber: not sure this would be beneficial as we would then numpify slices of the large tensor.

PR #336, opened 2018-03-23
Make sure all model parameters are prefixed by a model prefix. Default: ''

fhieber:
Most of our layer implementations already support naming parameters with a prefix. This change propagates this pattern up to the SockeyeModel class to allow prefixing all parameter names of a model with a certain name. By default the model prefix is the empty string, so nothing changes. 
In the future this allows to create multiple models / model components with different parameter sets.

I added a check in SockeyeModel.load_params_from_file() to ensure the loaded parameters match a prefix.

until you can check this box.

### DIFF
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -453,7 +453,9 @@ class RecurrentDecoder(Decoder):
         # TODO: implement variant without input feeding
         self.config = config
         self.rnn_config = config.rnn_config
-        self.attention = rnn_attention.get_attention(config.attention_config, config.max_seq_len_source)
+        self.attention = rnn_attention.get_attention(config.attention_config,
+                                                     config.max_seq_len_source,
+                                                     prefix + C.ATTENTION_PREFIX)
         self.prefix = prefix
 
         self.num_hidden = self.rnn_config.num_hidden
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -34,13 +34,13 @@ logger = logging.getLogger(__name__)
 DecoderConfig = Union['RecurrentDecoderConfig', transformer.TransformerConfig, 'ConvolutionalDecoderConfig']
 
 
-def get_decoder(config: DecoderConfig) -> 'Decoder':
+def get_decoder(config: DecoderConfig, prefix: str) -> 'Decoder':
     if isinstance(config, RecurrentDecoderConfig):
-        return RecurrentDecoder(config=config, prefix=C.RNN_DECODER_PREFIX)
+        return RecurrentDecoder(config=config, prefix=prefix + C.RNN_DECODER_PREFIX)
     elif isinstance(config, ConvolutionalDecoderConfig):
-        return ConvolutionalDecoder(config=config, prefix=C.CNN_DECODER_PREFIX)
+        return ConvolutionalDecoder(config=config, prefix=prefix + C.CNN_DECODER_PREFIX)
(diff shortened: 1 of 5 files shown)


sockeye/decoder.py
```diff
@@ -34,13 +34,13 @@
 DecoderConfig = Union['RecurrentDecoderConfig', transformer.TransformerConfig, 'ConvolutionalDecoderConfig']
 
 
-def get_decoder(config: DecoderConfig) -> 'Decoder':
+def get_decoder(config: DecoderConfig, prefix: str) -> 'Decoder':
```
tdomhan: should we set the default here as well?

sockeye/model.py
```diff
@@ -166,6 +169,10 @@ def load_params_from_file(self, fname: str):
                                                      "This is either not a model directory or the first training "
                                                      "checkpoint has not happened yet." % fname)
         self.params, self.aux_params = utils.load_params(fname)
+        utils.check_condition(all(name.startswith(self.prefix) for name in self.params.keys()),
```
tdomhan: is it important to check this? What do we want to guard against?

sockeye/decoder.py
fhieber: done

sockeye/model.py
fhieber: At this point not really, but I want to guard against future problems. Right now, the model prefix is hardcoded/defaults to an empty string (See SockeyeModel constructor), but if we would create multiple SockeyeModels in one graph in the future and name them differently, the parameters would be tied to that name. This is not necessarily what we want so maybe we would want to strip the prefix when loading the parameters in the future, but in order to not forget this, I am putting in this check now.

sockeye/encoder.py
```diff
@@ -126,20 +128,20 @@ def get_recurrent_encoder(config: RecurrentEncoderConfig) -> 'Encoder':
...
 
 
-def get_convolutional_encoder(config: ConvolutionalEncoderConfig) -> 'Encoder':
+def get_convolutional_encoder(config: ConvolutionalEncoderConfig, prefix: str) -> 'Encoder':
     """
     Creates a convolutional encoder.
 
     :param config: Configuration for convolutional encoder.
-    :param embed_weight: Optionally use an existing embedding matrix instead of creating a new one.
+    :param prefix: Prefix.
```
davvil: This docstring is not specially helpful ;-) Expand to something like "Prefix for variable names.".

sockeye/model.py
```diff
@@ -91,33 +91,36 @@ class SockeyeModel:
...
 
         # encoder & decoder first (to know the decoder depth)
-        self.encoder = encoder.get_encoder(self.config.config_encoder)
-        self.decoder = decoder.get_decoder(self.config.config_decoder)
+        self.encoder = encoder.get_encoder(self.config.config_encoder, prefix=self.prefix)
+        self.decoder = decoder.get_decoder(self.config.config_decoder, prefix=self.prefix)
 
         # source & target embeddings
-        embed_weight_source, embed_weight_target, out_weight_target = self._get_embed_weights()
+        embed_weight_source, embed_weight_target, out_weight_target = self._get_embed_weights(prefix)
```
davvil: Nit: Before you used 'self.prefix' and here and the lines below just 'prefix'. I would prefer using self.prefix overall for consistency.

sockeye/model.py
davvil: Nit: This should be an assert. (I can be boring, I know ;-) )

PR #338, opened 2018-03-26
Batched top-k implementation

Deseaus:
In the current numpy-based implementation of top-k, we use `np.argpartition` to get the smallest `k` scores. We then do a resort because we assumed that the returned partition might not be in sorted order (discussed [here](https://github.com/awslabs/sockeye/issues/270)). It turns out that `np.argpartition` can return an already sorted partition, but it's a bit hidden in the docs, found out about it on [stack overflow](https://stackoverflow.com/a/42185645).

In microbenchmarks, it's nearly 2x faster that the current approach in master:

```
In [2]: import numpy as np

In [3]: long = np.random.rand(5, 32000)

In [4]: long
Out[4]:
array([[ 0.19846017,  0.95168003,  0.6515433 , ...,  0.23239723,
         0.05825158,  0.19316375],
       [ 0.56247234,  0.12604064,  0.63571076, ...,  0.2767887 ,
         0.7993372 ,  0.64724466],
       [ 0.19306748,  0.06372986,  0.90068909, ...,  0.70291493,
         0.65553369,  0.01743039],
       [ 0.00679545,  0.83289007,  0.09616722, ...,  0.92498974,
         0.19441304,  0.68613256],
       [ 0.98585787,  0.56107536,  0.63116259, ...,  0.10633627,
         0.43929976,  0.51507055]])

In [5]: long = long.flatten()

In [6]: long.shape
Out[6]: (160000,)

In [10]: %%timeit
    ...: args = np.argpartition(long, 5)[:5];
    ...: args = args[np.argsort(long[args])]
    ...:
1.89 ms ± 197 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [11]: %%timeit
    ...: args = np.argpartition(long, range(5))[:5]
    ...:
977 µs ± 88.1 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

I also translated a large (5431 sent) test set on GPU on a p2.xlarge and compared the total times:

* Mxnet-based top-k: `Total time: 954.4973, sec/sent: 0.1757, sent/sec: 5.6899`
* Numpy-based top-k: `Total time: 1031.7326, sec/sent: 0.1900, sent/sec: 5.2640`
* Simplified numpy top-k (this PR): `Total time: 946.1822, sec/sent: 0.1742, sent/sec: 5.7399`

There were no differences in the translation output.

Given that the numbers look better than the mxnet-based one, I would recommend switching over to the simplified numpy version pending further speed tests. I'm currently running the same test on CPU, but it will take a while. I'll update the PR with the results. I think @mjpost has a nice setup for speed testing, would be nice if you could run the code in this PR :)

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
I love these kinds of small but relevant changes, great finding, thanks!
Let's wait for the CPU times and have @mjpost chime in on whether we should revert the smallest_k_mx change (using mxnet topk) we just introduced.

Deseaus:
A small update with initial numbers on CPU:

* Numpy master: `Total time: 5022.9249, sec/sent: 0.9249, sent/sec: 1.0812`
* Numpy simplified: `Total time: 4993.4917, sec/sent: 0.9194, sent/sec: 1.0876`

Slight improvement, this was only run on a p2.xlarge though, no exactly the most powerful CPU machine :)

fhieber:
One thing we could do is to decouple the revert of the cpu/gpu branching and only change the utils.smallest_k() function. We could then merge this PR as is and revisit the context branching again later.

Deseaus:
I pushed a new version where I'm batching the top_k computation in a single call. While this doesn't make much of a difference in single-sentence translation, it does bring a nice speed-up in batch translation. I've consolidated the various scores below (*simplified numpy top-k* was the first iteration of this PR adding `range()` to `argpartition`; *batched numpy top-k* is the current version).

The code is a mess and would need to be cleaned up, but just so you get an idea of the approach. I haven't found a cleaner way of offsetting the indexes in `best_hyp_indices`. Still needs further testing for sure to find the overall speed differences. At batch size 10, there doesn't seems to be much of a difference between *simplified numpy top-k* and *master numpy-based top-k*, for example.

## GPU results:

```
Master mxnet-based top-k     [GPU] [Batch 1]: `Total time: 954.4973, sec/sent: 0.1757, sent/sec: 5.6899`
Master numpy-based top-k     [GPU] [Batch 1]: `Total time: 1031.7326, sec/sent: 0.1900, sent/sec: 5.2640`
Simplified numpy top-k       [GPU] [Batch 1]: `Total time: 946.1822, sec/sent: 0.1742, sent/sec: 5.7399`
Batched numpy top-k          [GPU] [Batch 1]: `Total time: 932.8663, sec/sent: 0.1718, sent/sec: 5.8218`
```

```
Master mxnet-based top-k       [GPU] [Batch 10]:  ???
Master numpy-based top-k       [GPU] [Batch 10]: `Total time: 452.3622, sec/sent: 0.0833, sent/sec: 12.0059`
Simplified numpy top-k         [GPU] [Batch 10]: `Total time: 452.3817, sec/sent: 0.0833, sent/sec: 12.0053`
Batched numpy top-k            [GPU] [Batch 10]: `Total time: 273.6624, sec/sent: 0.0504, sent/sec: 19.8456`
```

## CPU results:

```
Master numpy-based top-k     [CPU] [Batch 1]: `Total time: 5022.9249, sec/sent: 0.9249, sent/sec: 1.0812`
Simplified numpy top-k       [CPU] [Batch 1]: `Total time: 4993.4917, sec/sent: 0.9194, sent/sec: 1.0876`
Batched numpy top-k          [CPU] [Batch 1]: ???
```

sockeye/inference.py
```diff
@@ -302,11 +302,6 @@ def run_decoder(self,
                     prev_word: mx.nd.NDArray,
                     bucket_key: Tuple[int, int],
                     model_state: 'ModelState') -> Tuple[mx.nd.NDArray, mx.nd.NDArray, 'ModelState']:
-        """
```
fhieber: this shouldn't be removed

sockeye/inference.py
```diff
@@ -1279,16 +1274,31 @@ def _beam_search(self,
...
+            flat_idxs = np.argpartition(folded_scores, range(self.beam_size))[:,:self.beam_size]
+            # Score values: (batch_size, beam_size)
+            folded_value_scores = folded_scores[np.arange(folded_scores.shape[0])[:, None], flat_idxs]
+
+            scores_accumulated[:,0] = folded_value_scores.ravel()
+            (best_hyp_indices[:], best_word_indices[:]) = np.unravel_index(flat_idxs.ravel(), scores.shape)
+
+            # TODO hussed: best_hyp_indices are not offset row-wise
+            # use the previous approach, find a cleaner way
             for sent in range(self.batch_size):
```
fhieber: I was actually starting to do the same thing (batched topk) yesterday. One idea I had was to adapt the indices by broadcast adding a mask we can create once in beam search.

fhieber:
I too the freedom to take your code and merge it with what I had for batched mxnet topk. How about adding a function in utils.py that looks like this:
```python
def topk(scores: mx.nd.NDArray, k: int, batch_size: int, t: int, use_mxnet_topk: bool):
    # scores: (batch_size * beam_size, target_vocab_size)
    if t == 1:  # beam size is effectively 1
        folded_scores = scores[::k, :]
    else:
        # (batch_size, beam_size * vocab_size)
        folded_scores = scores.reshape((batch_size, k * scores.shape[-1]))

    if use_mxnet_topk:
        values, indices = mx.nd.topk(folded_scores, axis=1, k=k, ret_typ='both', is_ascend=True)
        best_hyp_indices, best_word_indices = np.unravel_index(indices.astype(np.int32).asnumpy().ravel(), scores.shape)
        values = values.reshape((-1,))
    else:
        folded_scores = folded_scores.asnumpy()
        # Get the scores
        # Indexes into folded_scores: (batch_size, beam_size)
        flat_idxs = np.argpartition(folded_scores, range(k))[:, :k]
        # Score values: (batch_size, beam_size)
        values = folded_scores[np.arange(folded_scores.shape[0])[:, None], flat_idxs].ravel()
        best_hyp_indices, best_word_indices = np.unravel_index(flat_idxs.ravel(), scores.shape)
```

This entirely gets rid of the previous for loop and would be called in `_beam_search()` like this:
```python
best_hyp_indices[:], best_word_indices[:], scores_accumulated[:, 0] = utils.topk(scores, self.beam_size, self.batch_size, t=t, use_mxnet_topk=False)  # TODO: determine when to switch
```

We could use something like this to benchmark all variations (CPU/GPU, batch/non-batch) etc.

Deseaus:
I ran some speed tests on a p2.xlarge on GPU with a variety of setups, here are the graphs:

* The green and orange lines correspond to baselines in master, with the commit ID in brackets.
* Red and blue are the batched topK implementations that @fhieber referenced above.
* The purple line is a two-step topK explained below.

## Split by batch size

speedbybatch


## Split by beam size

speedbybeam

## Discussion

* It seems that on GPU the batched topK implementations don't negatively affect the batch=1 use case, as was expected.
* Batched topK brings nice speed-ups for most batch-beam size combinations, with some coming in roughly the same as the baseline.
* Mxnet and numpy batched topK versions are not terribly different, with the mxnet version performing better for `beam_size>5` and the numpy version for `beam_size<5`.
* The batched topK versions show a nearly linear speed-up as batch sizes increases in the greedy-case (batch=1). 

This last point led me to think that the bottleneck in the topK was the larger vocabulary dimension, which wasn't scaling well. For larger beam sizes we'd be performing topK row-wise on a matrix of size `(batch_size, beam_size * vocab_size)`, where adding more rows (i.e. sentences) scaled well as seen in the graphs for beam_size=1. So I tried the purple line, which is a two-step topK: first run topK row-wise on the matrix `(batch_size * beam_size, vocab_size)`, so we'd end up with the top k entries for each *hypothesis*. Then, run another topK on the reshaped result, which would only be `(batch_size, k*k)`. Unfortunately, this didn't show much improvement.

All in all, it seems the batched topK operators work well on GPU decoding, leaning towards the mxnet one. We should check there's no degradation on CPU for non-batched decoding.

fhieber:
Fantastic, thanks a lot for this extensive benchmarking!!
Based on your first picture, I would go for using mxnet-based batched topk on GPUs as it outperforms at beam size 5 the numpy version (beam size 5 being a common default setting). 
Can you run the CPU benchmarks as well? I think we should then go ahead and implement this asap.

Deseaus:
I'm running some CPU benchmarks now on a c5.xlarge. Only with batch sizes 1 and 10 now, since batching doesn't make much sense on a CPU. But wanted to try batch 10 and see what it looks like :)

Deseaus:
Here are the CPU speed tests on a c5.xlarge. Although unlikely as a production deployment, I also added `batch=10` numbers out of curiosity. These are the setups (couldn't get the colours to match the previous graphs, sorry!):

* The red and orange lines correspond to baselines in master, with the commit ID in brackets.
* Purple and blue are the batched topK implementations that @fhieber referenced above.
* The green line is a two-step topK explained below.

## Split by batch size

cpu_speedbybatch

## Split by beam size

cpu_speedbybeam

## Discussion

* The mxnet batched topK turns out worse in all cases except `beam=1,batch=1`.
* The numpy batched topK seems slightly better in the `batch=1` case (as CPU translation would probably be deployed), especially in `beam=5`.

Given that the greedy setting (`beam=1`) is unlikely, I would go for the numpy-based batched topK for CPU inference and keep the mxnet-based one for GPU. What do you think? I can update this PR with the code after.

fhieber:
Great to see this, thanks again!
I agree with you that we should switch to a batched topk implementation, using mxnet topk for GPUs & numpy topk for CPUs for now.
Feel free to update the PR whenever you find time. I will make time to review so that we can get merge this in a timely manner.

Deseaus:
Rebased and updated the PR with the changes.

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
...
-    """
-    if only_first_row:
-        matrix = mx.nd.reshape(matrix[0], shape=(1, -1))
-
-    # pylint: disable=unbalanced-tuple-unpacking
-    values, indices = mx.nd.topk(matrix, axis=None, k=k, ret_typ='both', is_ascend=True)
-
-    return np.unravel_index(indices.astype(np.int32).asnumpy(), matrix.shape), values
+    if use_mxnet_topk:
+        values, indices = mx.nd.topk(folded_scores, axis=1, k=k, ret_typ='both', is_ascend=True)
```
fhieber: You probably need to add `# pylint: disable=unbalanced-tuple-unpacking` above this line to make pylint happy.

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
         else:
             return self._M2 / self._count
 
-
-def smallest_k(matrix: np.ndarray, k: int,
-               only_first_row: bool = False) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
+def topk(scores: mx.nd.NDArray, k: int, batch_size: int, t: int, use_mxnet_topk: bool):
     """
     Find the smallest elements in a numpy matrix.
```
fhieber: Lets update this docstring to reflect what this function is doing.

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
...
 
-    :param matrix: Any matrix.
-    :param k: The number of smallest elements to return.
+    :param scores: Vocabulary scores for the next beam step. (beam_size * batch_size, target_vocabulary_size)
```
fhieber: lets define the shape here as `(batch_size * beam_size, target_vocabulary_size)`

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
...
+    :param k: The number of smallest scores to return.
+    :param batch_size: Number of sentences being decoded at once.
+    :param t: Time step in the beam search
     :param only_first_row: If true the search is constrained to the first row of the matrix.
```
fhieber: can be removed

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
...
+        folded_scores = folded_scores.asnumpy()
+        # Get the scores
+        # Indexes into folded_scores: (batch_size, beam_size)
+        flat_idxs = np.argpartition(folded_scores, range(k))[:, :k]
+        # Score values: (batch_size, beam_size)
+        values = folded_scores[np.arange(folded_scores.shape[0])[:, None], flat_idxs].ravel()
+        best_hyp_indices, best_word_indices = np.unravel_index(flat_idxs.ravel(), scores.shape)
+
+    # Offsetting the indices to match the shape of the scores matrix
+    best_hyp_indices += np.repeat(np.arange(0, batch_size * k, k), k)
```
fhieber: can we pass this offset mask into this function optionally (the function could create it based on batch_size and k if not given). We can create this once in `_beam_search`, for example were we also create the pad_dist and scores arrays, and then reuse it.

sockeye/utils.py
```diff
@@ -241,45 +241,41 @@ def variance(self) -> float:
         else:
             return self._M2 / self._count
 
-
-def smallest_k(matrix: np.ndarray, k: int,
-               only_first_row: bool = False) -> Tuple[Tuple[np.ndarray, np.ndarray], np.ndarray]:
+def topk(scores: mx.nd.NDArray, k: int, batch_size: int, t: int, use_mxnet_topk: bool):
```
fhieber: can we add typing for the return value, i.e. `Tuple[np.ndarray, np.ndarray, Union[np.ndarray, mx.nd.NDArray]]` ?

sockeye/inference.py
```diff
@@ -1253,6 +1252,12 @@ def _beam_search(self,
                 models_output_layer_w.append(m.output_layer_w.take(vocab_slice_ids))
                 models_output_layer_b.append(m.output_layer_b.take(vocab_slice_ids))
 
+        # Select topK to use
```
fhieber: self-explanatory, I would remove the comment

sockeye/inference.py
```diff
@@ -1253,6 +1252,12 @@ def _beam_search(self,
...
+        if self.context == mx.cpu():
+            use_mxnet_topk = False
```
fhieber: I'd prefer one-line setting for this simple boolean: `use_mxnet_topk = self.context == mx.cpu()`

sockeye/utils.py
Deseaus: Do you think there'd be a lot of time/memory saving from this? It'll make the function quite big, at that point I think it's cleaner to just use a partial:

```
topk = partial(utils.topk, self.beam_size, self.batch_size, use_mxnet_topk=use_mxnet_topk, offset=offset)

# inside beam search loop
_ = topk(scores, t)
```

What do you think?

fhieber: yes, good idea!

PR #341, opened 2018-03-28
[WIP] Scoring of existing translations

bricksdont:
Adding classes and cli for scoring existing translations. Work in progress, your feedback is very welcome.

until you can check this box.
- [ ] Updated CHANGELOG.md


fhieber:
this looks interesting, thanks!
Could you rebase onto latest master to remove diffs that are not part of this PR? Thank you!

bricksdont:
I'm glad to do that. Does this mean

git pull --rebase upstream master

If I'm on my scoring branch, and defined upstream as the awslabs master?

fhieber:
That should work. You might need to do `git fetch --all` first. If rebasing is a pain because of many commits in your branch, you can also simply merge. If we accept this PR after reviewing, we squash anyway, so you feature branch commit history is not that important.

bricksdont:
Thanks, I had to merge because rebasing resulted in several conflicts (probably because I messed up the commit history earlier).

sockeye/data_io.py
```diff
@@ -437,12 +437,14 @@ def load(self,
...
         num_pad_target = 0
+        map_buckets2sentence_ids = defaultdict(lambda: defaultdict(int))
 
         # Bucket sentences as padded np arrays
-        for sources, target in zip(zip(*sources_sentences), target_sentences):
+        for sentence_id, (sources, target) in enumerate(zip(zip(*sources_sentences), target_sentences)):
             source_len = len(sources[0])
             target_len = len(target)
             buck_index, buck = get_parallel_bucket(self.buckets, source_len, target_len)
+            ## TODO does this ever happen if we use maxlens from test set? check
```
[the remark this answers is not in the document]
a-rios: I was wondering what should happen in case we have maxlen given on CLI and source or target are > args.maxlen.. in this case, get_parallel_bucket will return None, should we drop the sentence pair..? What should be the output, empty line, None..? If maxlen is not given as argument, the data_iterator gets the maximum lengths from the data set itself, so get_parallel_bucket will always return a bucket index.. could also remove maxlen from arguments for scoring?

sockeye/data_io.py
```diff
@@ -1432,11 +1432,11 @@ def __init__(self,
                  target_data_name=C.TARGET_NAME,
                  label_name=C.TARGET_LABEL_NAME,
                  num_factors: int = 1,
-                 no_shuffle: Optional[bool] = False,
+                 shuffle: Optional[bool] = False,
```
fhieber: I think the default should be True here as well. Also, what about the ShardedIter?

bricksdont: Oh yes, I meant to set it to True there as well. For the sharded iter, is it enough to just also supply the `shuffle` argument to the underlying iterator `self.shard_iter`? The shard files (`self.shards_fnames`) are also shuffled currently.

sockeye/data_io.py
```diff
@@ -463,8 +465,14 @@ def load(self,
             # Once MXNet allows item assignments given a list of indices (probably MXNet 1.0): e.g a[[0,1,5,2]] = x,
             # we can try again to compute the label sequence on the fly in next().
             data_label[buck_index][sample_index, :target_len] = target[1:] + [self.eos_id]
-
+            
+            if buck_index in map_buckets2sentence_ids:
+                map_buckets2sentence_ids[buck_index].update({sample_index:sentence_id})
```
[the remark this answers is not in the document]
bricksdont: Time is roughly linear in the number of samples to score.

50 samples: 0.7 seconds
500 samples: 4.4 seconds
5000 samples: 41 seconds
50000 samples: 403 seconds

Depends on the particular model of course. Is that too slow? If yes I can rewrite.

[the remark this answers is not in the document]
fhieber: My point is not about the speed for scoring. This change would also affect any dataset you create during training, where the dataset is much much larger than what you typically score (for example 100m lines of parallel data). In this case, a dict with many elements might not be the ideal choice. Using an array of constant size (which you know in advance) would have O(1) insertion time and probably uses less memory overall.

sockeye/data_io.py
```diff
@@ -463,9 +466,13 @@ def load(self,
             # Once MXNet allows item assignments given a list of indices (probably MXNet 1.0): e.g a[[0,1,5,2]] = x,
             # we can try again to compute the label sequence on the fly in next().
             data_label[buck_index][sample_index, :target_len] = target[1:] + [self.eos_id]
-
+            
+            map_buckets2sentence_ids[buck_index][sample_index] = sentence_id
+            
             bucket_sample_index[buck_index] += 1
 
+        self.map_buckets2sentence_ids = map_buckets2sentence_ids
```
fhieber: thanks for the change!
You could directly operate on `self.map_buckets2sentence_ids`. One thing I would do is to add it in the constructor as optional (so that all class members are defined there) as such:
```
def __init__():
...
self.map_buckets2sentence_ids = None  # type: Optional[List[np.ndarray]]
```

bricksdont: Sure, let me know if the newest commit is what you had in mind.

sockeye/arguments.py
```diff
@@ -1109,3 +1109,80 @@ def add_init_embedding_args(params):
                         help='File to write initialized parameters to.')
     params.add_argument('--encoding', '-c', type=str, default=C.VOCAB_ENCODING,
                         help='Open input vocabularies with specified encoding. Default: %(default)s.')
+
+def add_scoring_args(params):
+    add_device_args(params)
+    add_logging_args(params)
+    scoring_params = params.add_argument_group("Scoring CLI")
+
+    scoring_params.add_argument('--source', '-s',
```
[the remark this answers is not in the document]
bricksdont: I refactored some of the arguments, the others are exclusive to scoring, in my opinion.

sockeye/data_io.py
fhieber: looks good, thanks!

sockeye/scoring.py
```diff
@@ -0,0 +1,481 @@
...
+        """
+        scored_batch = []
+
+        model.module.forward(batch, is_train=False)
+        outputs = model.module.get_outputs()
+
+        # split output array into probs per batch
+        sample_length = int(len(outputs[0]) / self.batch_size)
+
+        probs = mx.nd.array(outputs[0], ctx=self.context)  # shape is (t_len*batch_size, t_vocab)
```
fhieber: isnt `outputs[0]` already an ndarray?

fhieber: also, the comment is wrong I think, the shape should be `(batch_size * t_len, t_vocab)`
(a tensor of `(batch, len, vocab)` reshaped to `(batch*len, vocab)` which is a different ordering of data)

sockeye/scoring.py
```diff
@@ -0,0 +1,481 @@
...
+        probs_per_batch = [probs[i*sample_length:(i+1)*sample_length] for i in range(self.batch_size)]
```
fhieber: I think you could simplify this by just reshaping the module output back into `(batch, len, t_vocab)`

sockeye/scoring.py
```diff
@@ -0,0 +1,481 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""
+Implement scoring of existing translations.
```
fhieber: I think here or somewhere in the help string for this CLI it should be made a bit clearer what scoring means for the user, i.e. what are the scores returned and how are they obtained.

sockeye/arguments.py
```diff
@@ -1109,3 +1121,51 @@ def add_init_embedding_args(params):
...
+                               required=True,
+                               nargs='+',
+                               help='Model folder(s). Use multiple for ensemble scoring.')
+    scoring_params.add_argument('--checkpoints', '-c',
+                               default=None,
+                               type=int,
+                               nargs='+',
+                               help='If not given, chooses best checkpoints for model(s). '
+                                    'If specified, must have the same length as --models and be integer')
+    scoring_params.add_argument('--output-type',
```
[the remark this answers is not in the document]
tdomhan: I guess one use case would be to plot the alignment of a given translation, no? (not sure how important it is and would probably be fine either way)

sockeye/arguments.py
```diff
@@ -1109,3 +1121,51 @@ def add_init_embedding_args(params):
...
+                               default=C.OUTPUT_HANDLER_SCORE,
+                               # scoring does not result in beam history
+                               choices=[H for H in C.OUTPUT_HANDLERS if H != C.OUTPUT_HANDLER_BEAM_STORE],
```
tdomhan: variables should be lower case.

sockeye/data_io.py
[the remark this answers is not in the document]
tdomhan: that's a good question. For translation we split a sentence into multiple smaller pieces if it exceeds the maximum input length and translate them separately. For scoring it would make sense to either do something similar or reject those sentences, return an empty line and document this behavior in the CLI.

sockeye/arguments.py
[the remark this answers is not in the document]
bricksdont: I added this argument because we could probably extend the list of handlers in the future. Yes, plotting the alignment would be useful, either producing plots directly or output a format that can be turned into plots (https://github.com/M4t1ss/SoftAlignments#how-to-get-alignment-files-from-nmt-systems). That said, only RNN models have attention scores right now and I can also drop the argument for now.

sockeye/scoring.py
```diff
@@ -0,0 +1,481 @@
...
+                        self.config.config_data.max_seq_len_source, self.config.config_data.max_seq_len_target)
+            symbol, _, __ = sym_gen(data_iter.buckets[0])
+            return mx.mod.Module(symbol=symbol,
+                                 data_names=data_names,
+                                 label_names=label_names,
+                                 logger=logger,
+                                 context=self.context)
+
+
+Tokens = List[str]
```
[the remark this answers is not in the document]
bricksdont: Could you clarify: where would this be defined?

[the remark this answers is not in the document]
fhieber: https://github.com/awslabs/sockeye/blob/master/sockeye/inference.py#L535

Might make sense to move this type definition to a more central place (vocab or data_io maybe).

sockeye/scoring.py
bricksdont: My newest commit contains more information here. Let me know if this is what you expected.

sockeye/data_io.py
```diff
@@ -1333,10 +1349,11 @@ def __init__(self,
                  target_data_name=C.TARGET_NAME,
                  label_name=C.TARGET_LABEL_NAME,
                  num_factors: int = 1,
+                 shuffle: Optional[bool] = True,
```
tdomhan: `Optional` means it can be `None`, I guess that's not intended here. Use `bool` instead.

sockeye/score.py
```diff
@@ -0,0 +1,138 @@
...
+        logger = setup_main_logger(__name__,
+                                   console=not args.quiet,
+                                   file_logging=False)
+    utils.log_basic_info(args)
+
+    if args.checkpoints is not None:
+        check_condition(len(args.checkpoints) == len(args.models), "must provide checkpoints for each model")
+
+    with ExitStack() as exit_stack:
+        context = translate._setup_context(args, exit_stack)
```
tdomhan: minor: might make sense to remove the underscore to make it explicit that the method is used outside of the module.

sockeye/scoring.py
```diff
@@ -0,0 +1,474 @@
...
+                                           bucket_batch_sizes,
+                                           shuffle=False)
+
+    return data_iter, config_, map_buckets2sentence_ids
+
+
+def get_max_source_and_target(args: argparse.Namespace) -> Tuple[int, int]:
+    source_lines = utils.smart_open(args.source).readlines()
+    target_lines = utils.smart_open(args.target).readlines()
+    max_len_source = max([len(line.rstrip().split()) for line in source_lines])
```
tdomhan: you could use `data_io.read_content` here and maybe a generator instead of constructing the full list:
```
max([len(sent) for sent in data_io.read_content(source))
```

sockeye/scoring.py
```diff
@@ -0,0 +1,474 @@
...
+                             data_iter=data_iter,
+                             config=config,
+                             checkpoint=checkpoint,
+                             bucketing=bucketing)
+        models.append(model)
+
+    return models
+
+
+def create_data_iter_and_vocab(args: argparse.Namespace,
```
tdomhan: So scoring in general has things in common with both training (a single MXNet Module capturing the entire computation graph) and inference (translation one sentence at a time and making sure to have the correct order on the output).

In the code below you make use of a lot of the training data loading code. I'm wondering if it would be easier to factor out some of the logic in `translate.py` and reuse it here instead. For example you could make use of `translate.py.Translator._get_inference_input`, so that you could simply have a for loop over sentences and then from each sentence (or list of sentences) construct an NDArray. This would get around using RawParallelDatasetLoader and putting things in the correct order in the end.

What do you think?

mjpost:
Thanks for submitting this! I have a use for this and it would be really handy to have merged in. I just took a look over the code, though, and it seems like the changes here are many more than you need. Why no

sockeye/scoring.py
mjpost: Thanks for submitting this! I think it would be very useful and would like to have this merged in. However, looking over the code, it seems much more extensive than necessary. Why not just augment `translate.py` and then force the choice of the word at each point? To sketch this out:

- Require the JSON format, using something like:

      { 'text': "This is a test', 'force': "This is its translation" }

   You then pass `--json-input` and it will parse this out, making a `force` field (ha) available

- In `inference.py`, you'd add a field to `TranslatorInput` and turn the `force` field (still funny) to IDs

- Then, instead of calling `_beam_search()`, you could call something like `_force_decode()`, which would copy some code from `_beam_search()` (like the loop over `t` and the call to `_decode_step()`) to get the probs, but would then choose the appropriate column instead of calling `topk()`

That would eliminate a lot of extra files and redundancy.

bricksdont: Hi Tobi and Matt,
Thanks for your comments - I did not forget about this PR;
I agree that  much of the inference code could be reused, hopefully I will soon find time to implement your suggestions.

tdomhan: In general I don't mind having a separate module for scoring. Scoring has a lot more in common with training than it was with inference, as like in training you can create the complete computation graph. Force decoding in `inference.py` adds a lot of unnecessary overhead. At the same time we could try to share the logic for creating batches from inference. So to some extend it makes sense to share code from both training (computation graph construction) and inference (data_io). 
So in general I'd vote for keeping the separate scoring module, but maybe trying to factor out the needed functionalities from the training and inference modules.

davvil: Just a wild shot here without any specific pointers at code, but when PPL is computed during training this in fact force decoding. Could we take advantage of it?

a-rios: the main reason to re-use the data iterator from training was to not have to deal with bucketing for source-target pairs - the only difference to training is that we need to keep track of the original input order. scoring really has not much in common with inference, except the ordering, and a 'forced' decoding seems overly complicated when you can just use decode_sequence?

mjpost: I agree that conceptually it's closer to training since the source and target are fixed. But I think this could be done in about 20 lines of code as an inference task (where bucketing and batching will come along for free), and I think it would take a lot more as an inference task, and that appeals to me.

tdomhan:
While conceptually scoring is closer to training with the single computation graph, I do like the simplicity of your approach, Matt. It does make it much easier to reuse all existing translate.py/inference.py logic.

mjpost:
Okay, based on discussions, we should:

- get rid of `_force_decode()` and add the forced decoding functionality to `_beam_search()`, ideally by setting `topk()` to some version of `pick()` so that everything is as seamless as possible.
- added a separate CLI, `sockeye.score`, that does scoring, and accepts source and target input via file flags (`-s` and `-t`) or as pasted input on STDIN. This means we can also get rid of the extra field in the JSON object and the special processing that requires.

Sorry if I have jumped in and snatched this from you, @bricksdont: I had an immediate need for this and have been neck deep in the inference code for a while, so it was a quick idea to test.

mjpost:
There has been no update on this and I think it's superseded by #413, so I'm going to close it for now. Feel free to reopen it, if someone wants to try the computation graph approach some time in the future.

bricksdont:
Hi Matt,
All fine for me - your implementation makes a lot of sense. Sorry I did not get back to you on this.

PR #343, opened 2018-04-03
use hard link instead of soft

mjpost:
Is there any reason we're using a soft link for `params.best`? A hard link would not add space and would allow safer cleanup, e.g.,

    rm model/params.0*

after training completed would not risk deleting the best model file.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -817,7 +817,7 @@ class EarlyStoppingTrainer:
         actual_best_params_fname = C.PARAMS_NAME % self.state.best_checkpoint
         if os.path.lexists(best_params_path):
             os.remove(best_params_path)
-        os.symlink(actual_best_params_fname, best_params_path)
+        os.link(actual_best_params_fname, best_params_path)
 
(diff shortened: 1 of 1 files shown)


fhieber:
I don't have a strong opinion for either. What I like about a symlinks is that it is very transparent that symlinks are symlinks and what they link to. You can easily tell what checkpoint is the current best.
I don't know enough about hard links to judge if there are any limitations w.r.t file systems where/how model folders are packaged/zipped/archived etc. so maybe someone else can chime in on this?

mjpost:
Yeah, that's true that a hard loses you that information. Also, the tests failed :)

Maybe a better route is to add a utility that would clean up a trained model directory.

tdomhan:
at some point we thought about adding the averaging functionality to sockeye.train at the end of training, so that we would run averaging on the parameters and then discard all the parameters that are no longer needed.

mjpost:
It would be a good idea to have an easy way to clean up (lots of runs start to accumulate disk space pretty fast), but I'm not sure that Sockeye should do it without being told.

PR #346, opened 2018-04-05
mxboard: Only construct one SummaryWriter to reduce I/O overhead

fhieber:
Iteratively constructing multiple SummaryWriters from mxboard has overhead (see https://github.com/awslabs/mxboard/issues/10#issuecomment-378679195). This change makes sure to create only one instance.

until you can check this box.


fhieber:
Didn't notice anything in terms of performance. These logging events only happen at checkpoint time, so it doesn't affect training speed significantly anyway.

As a side note: it'd be great to approve changes through Github review first before merging PRs, even if they don't go into master. :)

sockeye/training.py
```diff
@@ -967,41 +967,38 @@ def _load_training_state(self, train_iter: data_io.BaseParallelSampleIter):
...
-        except ImportError:
+        if self.sw is None:
             return
 
+        for name, value in metrics.items():
+            if isinstance(value, mx.nd.NDArray):
+                self.sw.add_histogram(tag=name, values=value, bins=100, global_step=checkpoint)
+            else:
+                self.sw.add_scalar(tag=name, value=value, global_step=checkpoint)
+        self.sw.flush()
```
reminisce: One minor comment, `flush()` would block the main thread and may affect the performance if logging heavily. Suggest not using it.

sockeye/training.py
```diff
@@ -967,41 +967,38 @@ def _load_training_state(self, train_iter: data_io.BaseParallelSampleIter):
...
     def log_graph(self, symbol: mx.sym.Symbol):
-        try:
-            from mxboard import SummaryWriter
-            with SummaryWriter(logdir=self.logdir) as sw:
-                sw.add_graph(symbol)
-        except ImportError:
+        if self.sw is None:
             return
+        self.sw.add_graph(symbol)
+        self.sw.flush()
```
reminisce: Same as above.

sockeye/training.py
fhieber: good point. I will reconsider this if we start logging events more frequently. As it is right now, logging occurs only at checkpoints (e.g. every 4k updates). Overall loss in training speed should be negligible.

PR #348, opened 2018-04-06
WIP: Clamp vocab to be a multiple of 16

KellenSunderland:
Opening a PR for discussion.  I'll have to fix some unit tests before it's merged.

This PR clamps vocab to a multiple of 16.  Vocab size at this stage should be somewhat arbitrary already compared to the user's input during training right?  Would you guys be opposed to us clamping the vocab to the most frequent tokens up to a multiple of 16?  Having a vocab be a multiple of 16 is a requirement for tensorcores, but I think it should have a minimal impact for other models.  If you're not comfortable with it as the default behaviour we could also make it an option that's enabled when fp16 mode is active.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/vocab.py
+++ b/sockeye/vocab.py
@@ -67,7 +67,8 @@ def build_vocab(data: Iterable[str], num_words: int = 50000, min_count: int = 1)
     pruned_vocab = sorted(((c, w) for w, c in raw_vocab.items() if c >= min_count), reverse=True)
 
     vocab = islice((w for c, w in pruned_vocab), num_words)
-
+    div_16_len = int(len(pruned_vocab) / 16) * 16 - len(C.VOCAB_SYMBOLS)
+    vocab = islice(vocab, div_16_len)
     word_to_id = {word: idx for idx, word in enumerate(chain(C.VOCAB_SYMBOLS, vocab))}
     logger.info("Vocabulary: types: %d/%d/%d/%d (initial/min_pruned/max_pruned/+special) " +
                 "[min_frequency=%d, max_num_types=%d]",


mjdenkowski:
It sounds like this will allow our systems to be significantly faster overall so we should find a way to make it work.  I have a couple of general questions:

- Are performance improvements from tensorcores only applicable to fp16 mode?  Will there be any benefit from having vocabulary be a multiple of 16 in the general case?

- One issue with rounding down is that dropping vocab items in the training data can potentially confuse users as to why their system can't translate something it's seen in training.  An alternative would be rounding up and adding padding symbols.  This would also catch the edge case @fhieber mentioned.

The choices I see are rounding up vs down and enabling for fp16 only vs in general.  Do we have a favorite?

KellenSunderland:
To your first question, in general when running on GPUs it's good to partition your kernels to be a multiple of 32 to match warp size, but realistically I don't think it'll impact performance in the average case.  The real benefit is only apparent when using Tensor Cores.  

Padding does sound like it would make sense.  I'd vote we do that.  I'm also starting to lean towards just making this an optional behaviour.

fhieber:
Yes, padding and rounding up to a multiple of 16 would be the right choice. I don't see how some additional never-observed items in the embedding parameters would make a difference. I would though make this behaviour optional and enable it via an extra flag (maybe related to `--num-words`, something like `--num-words-multiple-of-16` or whatever).

Another question would be how to represent these bogus items in the vocabulary. Since they are never observed they are conceptually padding symbols, but vocabs are 1:1 mappings , so we cannot reuse `<pad>`. I would suggest using `<pad1>` ... `<pad15>` and adding them to the end of the vocab (largest ids). IDs are assigned by frequency, so assigning `<padX>` a large id is fine.

tdomhan:
+1 sounds like the right solution. I would also not remove any entries from the vocab, but rather round up. Given there isn't a huge downside we could consider always doing it, but it might make sense to be very explicitly about it in a log message.

fhieber:
closing in favor of #466

PR #351, opened 2018-04-09
allow `--tokenize none`

martinpopel:
This may be useful with a custom test set which is already tokenized.
Either the detokenized version of the reference may not be available
or you want to explore what is the effect of de/tokenization
e.g. when replicating an experiment from a paper
which used multi-bleu.perl on tokenized output
(which is a bad practice, I know).

- [*] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]'
until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/contrib/sacrebleu/sacrebleu.py
+++ b/contrib/sacrebleu/sacrebleu.py
@@ -1292,7 +1292,7 @@ def main():
                             help='use case-insensitive BLEU (default: actual case)')
     arg_parser.add_argument('--smooth', '-s', choices=['exp', 'floor', 'none'], default='exp',
                             help='smoothing method: exponential decay (default), floor (0 count -> 0.01), or none')
-    arg_parser.add_argument('--tokenize', '-tok', choices=[x for x in TOKENIZERS.keys() if x != 'none'], default='13a',
+    arg_parser.add_argument('--tokenize', '-tok', choices=TOKENIZERS.keys(), default='13a',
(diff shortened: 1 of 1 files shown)


martinpopel:
I understand the fear of violating the spirit of sacreBLEU and for that reason I included the warning message. BTW: SacreBLEU allows `--refs`, which can lead to similar problems - if someone uses a slightly different version of wmt14 references provided via `--refs`, the only rescue is reporting the signature where you can see that `test.wmt14` is missing and take the result with a grain of salt. Similarly, you can see `tok.none` in the signature.
My use case: someone [reported](https://github.com/tensorflow/tensor2tensor/pull/611#issuecomment-379687796) that [t2t-bleu](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/bin/t2t_bleu.py) (which should be equivalent to `sacrebleu -tok intl`) gives very different results than `multi-bleu.perl` on a test set which was indeed tokenized. I suggested to use sacreBLEU with `-tok 13a`, `-tok intl` and `-tok none` in order to explore what is the reason behind the difference (theoretically it could be also smoothing). My goal was to show this way why sacreBLEU is much better than multi-bleu.perl for public reports because it does the tokenization automatically by default, so my goal was to promote the spirit of sacreBLEU.:-)

mjpost:
This was merged in but squashed. Thanks, @martinpopel!

PR #360, opened 2018-04-18
Transformer source attention: removed combined linear projection for keys & values

fhieber:
This changes the transformer architecture to use separate linear projections for keys and values in the source attention layer.
During inference this yields a 5% speed improvement. As this changes the parameter set of a transformer decoder, I added (temporary) logic to `utils.load_params` to make this change backwards-compatible with existing models.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/utils.py
```diff
@@ -167,7 +167,16 @@ def load_params(fname: str) -> Tuple[Dict[str, mx.nd.NDArray], Dict[str, mx.nd.N
     for k, v in save_dict.items():
         tp, name = k.split(':', 1)
         if tp == 'arg':
-            arg_params[name] = v
+            """TODO(fhieber):
```
tdomhan: how about creating a 1.19 branch with this splitting removed, that we can use for accumulating other changes and also for not forgetting do remove this.

fhieber: given the benefits of the change lets get it in right now. I will keep it on my list to remove the logic/this TODO once we bump to 1.19.

PR #362, opened 2018-04-19
Allow empty sequences in train/validation data and skip them

fhieber:
Previously sockeye was throwing an exception when training or validation data contained empty segments.
With this change we simply skip over them (which included a little bit more logic to make sure parallel readers remain in sync). Overall this also simplified the iterator logic a little bit.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -16,6 +16,7 @@ Implements data iterators and I/O related functions for sequence-to-sequence mod
 """
 import bisect
 import logging
+import math
 import os
 import pickle
 import random
@@ -24,7 +25,6 @@ from collections import OrderedDict
 from contextlib import ExitStack
 from typing import Any, cast, Dict, Iterator, Iterable, List, Optional, Sequence, Sized, Tuple
 
-import math
 import mxnet as mx
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


sockeye/data_io.py
```diff
@@ -1025,45 +1017,45 @@ def __init__(self,
...
-        check_condition(bool(sequence), "Empty sequence in file %s" % self.path)
-
-        if vocab is not None and self.add_bos:
-            sequence.insert(0, self.vocab[C.BOS_SYMBOL])
+        for tokens in read_content(self.path, self.limit):
+            if self.vocab is not None:
+                sequence = tokens2ids(tokens, self.vocab)
+            else:
+                sequence = strids2ids(tokens)
+            if not sequence:
```
tdomhan: I'd personally explicitly check for length `if len(sequence) == 0`

sockeye/data_io.py
```diff
@@ -1025,45 +1017,45 @@ def __init__(self,
...
-        if not self._iterated_once:
-            self.count += 1
 
-        # fetch next element
-        self._next = next(self._iter, None)
-        if self._next is None:
-            self._iter = None
-            if not self._iterated_once:
-                self._iterated_once = True
+def parallel_iter(source_iters: Sequence[Iterable], target_iterable: Iterable):
```
tdomhan: we could even make this `Sequence[Iterable[Optional[Any]]]` to denote that we have the expectation that some are none. Anyway, not super important..

sockeye/data_io.py
```diff
@@ -1081,14 +1073,15 @@ def get_parallel_bucket(buckets: List[Tuple[int, int]],
                         length_target: int) -> Optional[Tuple[int, Tuple[int, int]]]:
     """
     Returns bucket index and bucket from a list of buckets, given source and target length.
-    Returns (None, None) if no bucket fits.
+    Returns (None, None) if no bucket fits or one or both lengths are 0.
```
tdomhan: this seems to be a doc left-over, right?

Issue #366, opened 2018-04-20
lr_scheduler seems to be reset if training is resumed


julian90:
Hi,

if training is resumed the learning rate scheduler seems to be reset.

This seems to be the case because the class OptimizerConfig contains the learning rate scheduler both directly in lr_scheduler and also in the dict params, and the module.init_optimizer function relies on the dict OptimizerConfig.params but not on OptimizerConfig.lr_scheduler.

But here:
https://github.com/awslabs/sockeye/blob/da518f1c6e90e08e2a1f8d2888608121f75b3d67/sockeye/training.py#L989-L993
only the lr_scheduler is updated but not the params dict.

A quick fix might be adding

`self.optimizer_config.params['lr_scheduler'] = self.optimizer_config.lr_scheduler`

before

https://github.com/awslabs/sockeye/blob/da518f1c6e90e08e2a1f8d2888608121f75b3d67/sockeye/training.py#L993

But overall it looks dangerous to have the learning rate scheduler both in lr_scheduler and in the dict params within the class OptimizerConfig.

fhieber:
Hi @julian90, thanks for submitting this issue! This is clearly a bug and I agree about the double reference to the scheduler instance in OptimizerConfig.
Does this [change](https://github.com/awslabs/sockeye/commit/778bdcf0a839e44fdf45c323e667bb900a52fef0) fixes the problem for you?

PR #369, opened 2018-04-22
Proper reloading of learning rate scheduler state at training resumption

fhieber:
until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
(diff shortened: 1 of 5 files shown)


fhieber:
This should fix #366 .

fhieber:
`Trainer.initialize_optimizer()` is just a wrapper for `self.model.initialize_optimizer`, but this is not related to the bug. The bug was that `OptimizerConfig` had two references to the `lr_scheduler` instance, one as a member of `OptimizerConfig`, and once in the optimizer_params (which are used by MXNet). At training resumption we unpickled the lr_scheduler instance, but only replaced `OptimizerConfig.lr_scheduler` with that pickle object, not the optimizer_params for MXNet.
This change makes sure there is only one reference held.
In what line would you like a comment?

PR #373, opened 2018-04-25
Log loading time of models and topklexicon at inference

fhieber:
(self-descriptive)

until you can check this box.

### DIFF
--- a/sockeye/lexicon.py
+++ b/sockeye/lexicon.py
@@ -16,6 +16,7 @@ import collections
 import operator
 import os
 import sys
+import time
 from typing import Dict, Generator, Tuple, Optional
 
 import mxnet as mx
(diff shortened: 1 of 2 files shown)


sockeye/lexicon.py
```diff
@@ -12,12 +12,13 @@
...
 
 import argparse
-import collections
 import operator
 import os
 import sys
+import time
 from typing import Dict, Generator, Tuple, Optional
 
+import collections
```
mjdenkowski: Looks like an automatic tool accidentally moved `collections` out of the standard library import section.

fhieber: Actually thats the 'optimize import' function of PyCharm, which is supposed to standardize the order. I don't know exactly what the order is (maybe collections is not seens as a built-in module), but since we are using it heavily I'd like to rely on it. Are you ok with this?

mjdenkowski: I checked the documentation for PyCharm and optimize imports is just supposed to follow the PEP8 style guide: https://www.python.org/dev/peps/pep-0008/#imports.  I haven't seen this issue before so it may be a one-off glitch.  What about manually editing it this time and revisiting if it happens again with other files?

fhieber: ok, will do

PR #374, opened 2018-04-26
experimental option to override transformer dtype during inference

logogin:
First functional change to override transformer dtype. Marked cli option as experimental.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -1087,6 +1087,11 @@ def add_inference_args(params):
                                type=float,
                                help='Beta factor for the length penalty used in beam search: '
                                     '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
+    decode_params.add_argument('--override-dtype',
+                               default=None,
+                               type=str,
+                               help='EXPERIMENTAL: may be changed or removed in future. Overrides training dtype of '
+                                    'encoders and decoders during inference. Default: %(default)s')
 
 def add_evaluate_args(params):
     eval_params = params.add_argument_group("Evaluate parameters")
--- a/sockeye/encoder.py
+++ b/sockeye/encoder.py
@@ -200,6 +200,7 @@ class Encoder(ABC):
 
     @abstractmethod
     def __init__(self, dtype):
+        logger.info('{}.{} dtype: {}'.format(self.__module__, self.__class__.__name__, dtype))
         self.dtype = dtype
 
     @abstractmethod
@@ -932,6 +933,7 @@ class TransformerEncoder(Encoder):
         :param seq_len: Maximum sequence length.
         :return: Encoded versions of input data data, data_length, seq_len.
         """
+        data = utils.cast_conditionally(data, self.dtype)
         if self.config.dropout_prepost > 0.0:
             data = mx.sym.Dropout(data=data, p=self.config.dropout_prepost)
 
@@ -941,11 +943,12 @@ class TransformerEncoder(Encoder):
                                                                        num_heads=self.config.attention_heads,
                                                                        fold_heads=True,
                                                                        name="%sbias" % self.prefix), axis=1)
-
+        bias = utils.cast_conditionally(bias, self.dtype)
         for i, layer in enumerate(self.layers):
             # (batch_size, seq_len, config.model_size)
             data = layer(data, bias)
         data = self.final_process(data=data, prev=None)
+        data = utils.uncast_conditionally(data, self.dtype)
         return data, data_length, seq_len
 
     def get_num_hidden(self) -> int:
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -352,7 +352,8 @@ def load_models(context: mx.context.Context,
                 softmax_temperature: Optional[float] = None,
                 max_output_length_num_stds: int = C.DEFAULT_NUM_STD_MAX_OUTPUT_LENGTH,
                 decoder_return_logit_inputs: bool = False,
-                cache_output_layer_w_b: bool = False) -> Tuple[List[InferenceModel],
+                cache_output_layer_w_b: bool = False,
+                override_dtype: Optional[str] = None) -> Tuple[List[InferenceModel],
                                                                List[vocab.Vocab],
                                                                vocab.Vocab]:
     """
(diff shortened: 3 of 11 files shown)


sockeye/arguments.py
```diff
@@ -1067,6 +1067,11 @@ def add_inference_args(params):
                                type=float,
                                help='Beta factor for the length penalty used in beam search: '
                                     '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
+    decode_params.add_argument('--override-dtype',
+                               default=None,
+                               type=str,
+                               help='EXPERIMENTAL: may be changed or removed in future. Forces dtype of encoders and '
```
fhieber: 'Forces' -> 'Enforces' or 'Overrides training dtype of encoders and decoders during inference.'

sockeye/arguments.py
```diff
@@ -1067,6 +1067,11 @@ def add_inference_args(params):
...
+                                    'decoders during inference. Default: %(default)s')
```
fhieber: [supernit] Period missing :)

sockeye/encoder.py
```diff
@@ -932,6 +933,7 @@ def encode(self,
         :param seq_len: Maximum sequence length.
         :return: Encoded versions of input data data, data_length, seq_len.
         """
+        data = mx.sym.cast(data=data, dtype=self.dtype)
```
fhieber: as said in my previous comment, this code is also used in training and without the override-dtype flag, so we should make sure that (no-op) casting does not incur any performance penalties.

sockeye/transformer.py
```diff
@@ -291,14 +293,15 @@ def __init__(self, max_length: int) -> None:
     def forward(self, is_train, req, in_data, out_data, aux):
         # lengths: (batch_size,)
         lengths = in_data[0]
+        dtype = lengths.dtype
 
         # (batch_size, max_length)
-        data = mx.nd.zeros((lengths.shape[0], self.max_length), ctx=lengths.context)
+        data = mx.nd.zeros((lengths.shape[0], self.max_length), dtype=dtype, ctx=lengths.context)
         data = mx.nd.SequenceMask(data=data,
```
fhieber: the bias tensor is added to the data in the encoder. Are you sure we are not causing any fp16 overflow AFTER this addition?

sockeye/transformer.py
```diff
@@ -291,14 +293,15 @@ def __init__(self, max_length: int) -> None:
     def forward(self, is_train, req, in_data, out_data, aux):
```
fhieber: transformer.py also contains `AutoRegressiveBias` which also uses a large negative value to ensure auto-regressiveness in the Transformer Decoder. This also needs to be fixed.

sockeye/inference.py
```diff
@@ -371,6 +372,8 @@ def load_models(context: mx.context.Context,
                                         vocabulary.  Used when logits/softmax are handled separately.
     :param cache_output_layer_w_b: Models cache weights and biases for logit computation as NumPy arrays (used with
                                    restrict lexicon).
+    :param override_dtype: Overrides dtype of encoder and decoder to a different from one, defined in config
```
fhieber: Maybe "Overrides dtype of encoder and decoder defined at training time to a different one."?

sockeye/transformer.py
```diff
@@ -69,6 +69,7 @@ class TransformerEncoderBlock:
     def __init__(self,
                  config: TransformerConfig,
                  prefix: str) -> None:
+        self.dtype = config.dtype
```
fhieber: why is this needed? couldn't we use the (potentially overriden) `self.config.dtype`?

sockeye/transformer.py
```diff
@@ -120,6 +121,7 @@ class TransformerDecoderBlock:
...
```
fhieber: same here

sockeye/transformer.py
logogin: you are right, missed that config is a member already.

sockeye/encoder.py
logogin: I will try to double-ckeck that. Will be something to report as a bug in MxNet if no-op casting will have penalties.

sockeye/arguments.py
logogin: where? After "experimental"?

sockeye/transformer.py
logogin: actually, this is just a fix for the custom operator. Due to the zeros symbol without dtype it was forced always to be fp32. But in the change, bias calculated in fp32 and casted to fp16 before addition.

sockeye/arguments.py
fhieber: 'Default: %(default)s' -> 'Default: %(default)s.' :)

sockeye/transformer.py
fhieber: I am not sure I get what you are saying. I understand that the custom op now creates data in the correct dtype. The problem is that you are masking this fp16 data in some places with the smallest possible value in the fp16 range. Now when you actually use this bias tensor in the transformer model, it gets added to the encoder representation tensors. Depending on how they look like, this addition might cause an overflow: `fp16.min + -x`. 
So instead of using a value for masking that is right at the limit of the representable numeric range, it might be better to use something less extreme.

sockeye/transformer.py
fhieber: Its the same argument as in my comment above. you need to guard against hitting the fp16 range at masking time, and when using the bias tensor in the model.

logogin: added unit test and support for dtype argument

sockeye/transformer.py
logogin: Changed `float16` constant to be something close to maximum, but not maximum possible number.

sockeye/transformer.py
```diff
@@ -399,9 +401,10 @@ def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
 @mx.operator.register("auto_regressive_bias")
 class AutoRegressiveBiasProp(mx.operator.CustomOpProp):
 
-    def __init__(self, length: str) -> None:
+    def __init__(self, length: str, dtype: Optional[str] = C.DTYPE_FP32) -> None:
```
fhieber: I don't think `dtype` could be None, so `Optional` does not apply here.

logogin: fixed

PR #376, opened 2018-05-03
Simplified computation of smoothed cross entropy loss in CrossEntropyMetric

fhieber:
found a way to compute smoothed cross entropy without explicitly representing the smoothed distribution. This PR uses this alternative way in the CrossEntropyMetric, which is for now only relevant for the Eve optimizer.

All existing tests still pass.

until you can check this box.

### DIFF
--- a/sockeye/loss.py
+++ b/sockeye/loss.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2017, 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not
 # use this file except in compliance with the License. A copy of the License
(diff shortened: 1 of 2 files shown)


sockeye/loss.py
```diff
@@ -145,40 +145,48 @@ def __init__(self,
...
 
     def update(self, labels, preds):
         for label, pred in zip(labels, preds):
             batch_size = label.shape[0]
             label = label.as_in_context(pred.context).reshape((label.size,))
-            # Ignore padding
-            # TODO: contribute ignoring padding for cross-entropy back to MXNet
-            ignore = (label == C.PAD_ID).astype(dtype=pred.dtype)
 
+            logprob = mx.nd.log(pred + 1e-8)
```
tdomhan: would it make sense to do this with `mx.sym.clip`?

fhieber: Ok, will use the same pattern as in MXNET's perplexity metric: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/metric.py#L749

fhieber:
ping @tdomhan

PR #381, opened 2018-05-04
Updating default parameters with values from our arXiv paper.

tdomhan:
(description of the change)

until you can check this box.
- [X] Updated CHANGELOG.md


tdomhan:
I've been meaning to do this for quite some time now. I think it's good to have the parameter values used in the arXiv paper as the default values to give a better out-of-the-box experience for users. I think it's fine to not bump the major version for this as models trained with the previous version can still be used at inference time.

fhieber:
Agreed that it makes sense to update the default parameters to more recent best practices.
Does the default of `sockeye.train` still fit on a single commodity GPU device though? It might be useful to have a default configuration that is not only runnable on a huge cloud computing instance :)

Also, you might need to update the README.md and WMT/seqcopy tutorials with adjusted arguments

tdomhan:
yeah, good points. I'll test the memory consumption and update the WMT tutorials next week.

pengzhao-intel:
Does this default configuration coverage to the state-of-the-art in the paper?

mjdenkowski:
For the WNMT18 baseline, we used a version of the arXiv settings adapted to use rate decay and early stopping vs a set number of steps: https://github.com/awslabs/sockeye/blob/wnmt18/wnmt18/train.sh.  These settings are also tailored to fit on a single GPU.

tdomhan:
So the 6 layer Transformer seems to run fine on a p2 with the WMT tutorial training data. So I'd vote for having this as the default. What do you think?

fhieber:
What about `--weight-tying --weight-tying-type=src_trg_softmax`?

tdomhan:
to use weight tying in the tutorial?

fhieber:
No, just asking because we refer to the new default settings as used in the arxiv paper, which had weight tying enabled.

tdomhan:
oh ok, good point. The "problem" is that `--weight-tying` is a boolean parameter, so we would have to rename it to `--no-weight-tying`, enabling it by default, but being able to turn it off. I guess it could make sense to do that.

tdomhan:
While in general having the Transformer and the learning parameters changed probably makes sense across language pairs and seq2seq problems, the same is probably not true for weight-tying, as we can't always share the source and target vocabs. Additionally, as one typically does not see a huge boost in BLEU from weight tying, but rather just needs to learn fewer parameters, I'd leave weight-tying as is by default (off), to cause fewer headaches when people use Sockeye for use cases where it does not apply.

tdomhan:
makes sense, done!

PR #402, opened 2018-05-18
Added support for config files

davvil:
**Update:**

As per comments, I updated the format to YAML. The corresponding updated example for the config file is now:

```
source: /Users/dvilar/corpora/iwslt/tokenize/de-en/train.tags.de-en.de
target: /Users/dvilar/corpora/iwslt/tokenize/de-en/train.tags.de-en.en
validation_source: /Users/dvilar/corpora/iwslt/tokenize/de-en/IWSLT16.TED.dev2010.de-en.de.txt
validation_target: /Users/dvilar/corpora/iwslt/tokenize/de-en/IWSLT16.TED.dev2010.de-en.en.txt
output: out
use_cpu: true
```

Fun fact: the original json config file is also correctly parse as yaml :-)

**Original description:**

Specify a config file with --config. Command line parameters have precedence over the values read from the config file, as expected.

The config file is a json serialization of the namespace returned by argparse, casted to a dictionary. The file args.json produced by the training can be used as config file.

The config file does not need to be complete. Missing parameters will be read from the command line or will take default values.

Minimal working example (`python -m sockeye.train --config config.json` starts a training, cpu activated):
```
{
  "source": "/Users/dvilar/corpora/iwslt/tokenize/de-en/train.tags.de-en.de",
  "target": "/Users/dvilar/corpora/iwslt/tokenize/de-en/train.tags.de-en.en",
  "validation_source": "/Users/dvilar/corpora/iwslt/tokenize/de-en/IWSLT16.TED.dev2010.de-en.de.txt",
  "validation_target": "/Users/dvilar/corpora/iwslt/tokenize/de-en/IWSLT16.TED.dev2010.de-en.en.txt",
  "output": "out",
  "use_cpu": true
}
```

Additionally the functionality introduced in 5a3bf5f8 has been reimplemented, accessing the argument_definitions member of ConfigArgumentParser.

### Note about the PR ###
This PR includes two commits. The first one mainly reverts arguments.py to pre-5a3bf5f8. The second one is the actual "meat" of this PR. Looking at the commits separately probably will provide a better overview.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -17,26 +17,11 @@ Defines commandline arguments for the main CLIs with reasonable defaults.
 import argparse
 import os
 import sys
-from functools import partial
-from typing import Any, Callable, Dict, Tuple, Optional
+from typing import Callable, Optional
 
 from sockeye.lr_scheduler import LearningRateSchedulerFixedStep
 from . import constants as C
 from . import data_io
-
-# Data structure to store argument definitions for various CLIs,
-# populated when CLI args are defined via argparse in the function below.
-ARGUMENT_DEFINITIONS = dict()  # type: Dict[Tuple[Any], Dict[str, Any]]
-
-
-def add_argument(params, *args, **kwargs):
-    """
-    Simple wrapper around argparse.ArgumentParser.add_argument() to store the argument specifications in another
-    data structure.
-    """
-    assert hasattr(params, 'add_argument')
-    params.add_argument(*args, **kwargs)
-    ARGUMENT_DEFINITIONS[args] = kwargs
 
 
 def regular_file() -> Callable:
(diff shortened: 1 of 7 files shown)


sockeye/arguments.py
```diff
@@ -15,14 +15,58 @@
...
+import types
 from typing import Callable, Optional
 
 from sockeye.lr_scheduler import LearningRateSchedulerFixedStep
 from . import constants as C
 from . import data_io
 
+# Adapted from https://stackoverflow.com/questions/28579661/getting-required-option-from-namespace-in-python
+class ConfigArgumentParser(argparse.ArgumentParser):
+    def __init__(self, *args, **kwargs):
```
fhieber: add '-> None'

sockeye/arguments.py
```diff
@@ -15,14 +15,58 @@
...
+        super().__init__(*args, **kwargs)
+        self.argument_definitions = {}
+        self.argument_actions = []
+        self._overwrite_add_argument(self)
+        self.add_argument("--config", help="Config file in JSON format.", type=str)  # Not FileType so that we can get the path here
```
fhieber: overly long line I think

sockeye/arguments.py
```diff
@@ -15,14 +15,58 @@
...
+
+    def register_argument(self, _action, *args, **kwargs):
```
fhieber: will arguments also be registered if there is no group?

sockeye/arguments.py
```diff
@@ -15,14 +15,58 @@
...
 import sys
+import types
 from typing import Callable, Optional
 
 from sockeye.lr_scheduler import LearningRateSchedulerFixedStep
 from . import constants as C
 from . import data_io
 
+# Adapted from https://stackoverflow.com/questions/28579661/getting-required-option-from-namespace-in-python
+class ConfigArgumentParser(argparse.ArgumentParser):
```
fhieber: This could use a docstring explaining what it does and why :)

fhieber:
This deserves a new minor version and a changelog entry as it is potentially interesting to many users.

fhieber:
@yx3zhang: FYI

yx3zhang:
Just a quick comment, this might be solving the 10% on top of the 90% this PR already covers. It would be more user friendly to support .yaml so documentation of each parameter can be provided in line.

fhieber:
Good point, we already depend on `pyyaml`, so switching from JSON to YAML shouldn't be difficult. We have been dumping the arguments to `<model>/args.json` only for logging purposes for a while, but I don't think users rely on it at this point.
So switching to YAML with a minor version bump should be fine.

fhieber:
@davvil #403 is a PR on top of yours to change to YAML

davvil:
@fhieber hadn't seen your commit, I implemented it myself independently.

As you pointed out, I originally chose json because it was the format we were dumping the args originally. I also found yaml more convenient.

sockeye/arguments.py
davvil: Yes they will. Line 34 calls _overwrite_add_argument on itself for supporting the non-group case. What is not supported (but not used in sockeye) are sub-parsers [and possibly other exotic argparse use cases].

davvil:
Sure, go ahead!

PR #413, opened 2018-05-25
[WIP] Forced decoding / scoring

mjpost:
This introduces a scoring module that will score an input against a reference:

    paste source_file ref_file | python3 -m sockeye.score -m model ...

It can also take a JSON object of the form:

    { 'text': 'the text to translate', 'target': 'the reference translation' }

It works with all the features — source factors, restricted lexicon, multiple models.

There is a lot of redundancy in this code and I am not sure whether to refactor it or not. The main redundancy is in `inference.py` between `_beam_search()` and `_force_decode()`. Refactoring for efficiency would require passing the target into `_beam_search()` and adding a lot of special case checking.

until you can check this box.
- [ ] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -574,22 +574,25 @@ class TranslatorInput:
...
                  factors: Optional[List[Tokens]] = None,
                  constraints: Optional[List[Tokens]] = None,
+                 target_tokens: Optional[Tokens] = None,
                  chunk_id: int = -1) -> None:
         self.sentence_id = sentence_id
         self.chunk_id = chunk_id
         self.tokens = tokens
         self.factors = factors
         self.constraints = constraints
+        self.target_tokens = target_tokens
```
tdomhan: would it make sense to add an assertion to the `chunks` methods to make sure it's not called when target_tokens are present (as currently their are not carried forward and it is any unclear what one should do in the case when there are target_tokens)?

sockeye/inference.py
```diff
@@ -1043,6 +1052,57 @@ def _log_linear_interpolation(predictions):
...
+            if isinstance(trans_input, BadTranslatorInput):
+                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=empty_translation()))
+
+            # empty input
+            elif len(trans_input.tokens) == 0:
+                translated_chunks.append(TranslatedChunk(id=input_idx, chunk_id=0, translation=empty_translation()))
+
+            # regular input
+            else:
+                # TODO(mjpost): max length checking
```
tdomhan: what should we go with long inputs in the case of scoring? Truncate + log?

sockeye/inference.py
```diff
@@ -1043,6 +1052,57 @@ def _log_linear_interpolation(predictions):
...
+            if rest > 0:
+                batch_translations = batch_translations[:-rest]
+            for chunk, translation in zip(batch, batch_translations):
+                translated_chunks.append(TranslatedChunk(chunk.sentence_id, chunk.chunk_id, translation))
+
+        # Sort by input idx
+        translations = [chunk.translation for chunk in sorted(translated_chunks)]
+
+        # Construct results
+        results = [self._make_result(x, y) for x, y in zip(trans_inputs, translations)]  # type: List[TranslatorOutput]
```
tdomhan: maybe don't call it `x` and `y`.

sockeye/inference.py
```diff
@@ -1574,6 +1645,110 @@ def _beam_search(self,
 
         return sequences, attentions, scores_accumulated, lengths, constraints, beam_histories
 
+    def _force_decode(self,
```
tdomhan: it'd probably be good to make this part of beam search.

mjpost:
Still WIP. I need to add test cases and the ability to have a global (translator-level instead of sentence-level) block list. This is half done, just need to find a few hours to finish it off.

sockeye/inference.py
```diff
@@ -1249,8 +1320,8 @@ def _decode_step(self,
                      step: int,
                      source_length: int,
                      states: List[ModelState],
-                     models_output_layer_w: List[mx.nd.NDArray],
-                     models_output_layer_b: List[mx.nd.NDArray]) \
+                     models_output_layer_w: List[mx.nd.NDArray] = list(),
```
tdomhan: mhh if anything we should set these to an empty tuple, as there is only one object per argument+function, so that modifying will lead to nasty side-effects.

tdomhan:
sounds good :)

mjpost:
Closing this, got a better idea (more soon).

PR #423, opened 2018-05-30
Refactored pruning logic and vectorized it

fhieber:
This cleans up the pruning logic a little bit and continues work started in #422. Thanks @mjpost for various explanations!

Changes include:
1) the modifications on various data structures as a result of pruning are now more local to where it matters.
2) vectorized the pruning function and moved it to `utils.py` (similar to `topk()`). Also using a partial now. Vectorization may help us in moving operations to HybridBlocks in the future.

until you can check this box.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1430,7 +1430,6 @@ class Translator:
             # (2) Special treatment for finished and inactive rows. Inactive rows are inf everywhere;
             # finished rows are inf everywhere except column zero, which holds the accumulated model score
             scores += scores_accumulated
-            scores_accumulated = mx.nd.where(inactive, self.inf_array, scores_accumulated)
             # Items that are finished (but not inactive) get their previous accumulated score for the <pad> symbol,
             # infinity otherwise.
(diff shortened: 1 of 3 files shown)


sockeye/inference.py
```diff
@@ -1515,22 +1484,23 @@ def _beam_search(self,
 
             # (6) Prune out low-probability hypotheses. Pruning works by setting entries `inactive`.
             if self.beam_prune > 0.0:
-                scores_accumulated, best_word_indices, inactive, finished = self._prune(scores_accumulated,
-                                                                                        best_word_indices,
-                                                                                        inactive,
-                                                                                        finished)
+                inactive = self.prune(scores_accumulated, finished)
+                best_word_indices = mx.nd.where(inactive, self.zeros_array, best_word_indices)
```
mjpost: I like the simplified signature for `prune()`. Since we set `best_word_indices` here after the call, it seems to me now that it would be clearer and simpler to do that here for `scores_accumulated`, instead of its reassignment in two separate places (below, for the same t, and above, at the next t).

PR #425, opened 2018-06-01
Refactor beam search to use gluon hybrid blocks

fhieber:
This refactors beam search to group fixed-size operations in beam search into cached ops through Gluon HybridBlocks.
My testing showed ~3% speed improvement. Not much, but consistent.

Another change that is included here is to not use columns from `sequences` to pass into the decoder step module, but use `best_word_indices` from the previous iteration. NDArray indexing seems expensive and ideally we should aim for avoiding all indexing ops in an iteration.

until you can check this box.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -842,7 +842,7 @@ class ModelState:
         self.states = [mx.nd.take(ds, best_hyp_indices) for ds in self.states]
 
 
-class LengthPenalty:
+class LengthPenalty(mx.gluon.HybridBlock):
     """
     Calculates the length penalty as:
     (beta + len(Y))**alpha / (beta + 1)**alpha
@@ -854,12 +854,21 @@ class LengthPenalty:
     :param beta: The beta factor for the length penalty (see above).
     """
 
-    def __init__(self, alpha: float = 1.0, beta: float = 0.0) -> None:
+    def __init__(self, alpha: float = 1.0, beta: float = 0.0, **kwargs) -> None:
+        super().__init__(**kwargs)
         self.alpha = alpha
         self.beta = beta
         self.denominator = (self.beta + 1.) ** self.alpha
 
-    def __call__(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray, float]:
+    def hybrid_forward(self, F, lengths):
+        if self.alpha == 0.0:
+            return F.ones_like(lengths)
+        else:
+            numerator = self.beta + lengths if self.beta != 0.0 else lengths
+            numerator = numerator ** self.alpha if self.alpha != 1.0 else numerator
+            return numerator / self.denominator
+
+    def get(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray, float]:
         """
         Calculate the length penalty for the given vector of lengths.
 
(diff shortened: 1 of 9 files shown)


mjpost:
Can you pull in the bug fix I just pushed and make sure the tests still pass? Some of the tests were broken for constrained decoding, which this might affect.

fhieber:
Yes will do that soon

fhieber:
Tests still pass. Also ran the system tests but not sure if they even cover constraints.

sockeye/inference.py
```diff
@@ -854,12 +854,21 @@ class LengthPenalty:
...
-    def __call__(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray, float]:
+    def hybrid_forward(self, F, lengths):
+        if self.alpha == 0.0:
+            return F.ones_like(lengths)
+        else:
+            numerator = self.beta + lengths if self.beta != 0.0 else lengths
+            numerator = numerator ** self.alpha if self.alpha != 1.0 else numerator
+            return numerator / self.denominator
+
+    def get(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray, float]:
```
tdomhan: Could we merge `get` and `hybrid_forward` somehow?  e.g.: 
```
   def get(...):
       return hybird_forward(None, lengths)
   def hybrid_forward(F, lengths):
        if F is None:
             return 1
         else:
             return F.ones_like(...)
```

sockeye/inference.py
```diff
@@ -1339,10 +1362,12 @@ def _beam_search(self,
         # General data structure: each row has batch_size * beam blocks for the 1st sentence, with a full beam,
         # then the next block for the 2nd sentence and so on
 
+        best_word_indices = mx.nd.full((self.batch_size * self.beam_size,), val=self.start_id, ctx=self.context,
+                                       dtype='int32')
         # sequences: (batch_size * beam_size, output_length), pre-filled with <s> symbols on index 0
         sequences = mx.nd.full((self.batch_size * self.beam_size, max_output_length), val=C.PAD_ID, ctx=self.context,
                                dtype='int32')
-        sequences[:, 0] = self.start_id
+        sequences[:, 0] = best_word_indices
```
tdomhan: is this faster, or why did you change this?

sockeye/inference.py
```diff
@@ -1464,35 +1484,38 @@ def _beam_search(self,
...
-            finished = all_finished
+            finished, scores_accumulated = self._normalize_finished.forward(best_word_indices,
+                                                                            finished,
+                                                                            scores_accumulated,
+                                                                            lengths)
 
             # (6) Prune out low-probability hypotheses. Pruning works by setting entries `inactive`.
             if self.beam_prune > 0.0:
-                inactive = self.prune(scores_accumulated, finished)
+                inactive = self._prune_hyps.forward(scores_accumulated, finished, self.inf_array)
```
tdomhan: why not also pull in the where's below?

sockeye/inference.py
fhieber: sure, done

sockeye/inference.py
fhieber: yes, as said in the PR text it seems that indexing is slow. This actually gave a small speedup.

sockeye/inference.py
fhieber: We just had changed this [recently](https://github.com/awslabs/sockeye/commit/fc440df9cccce7d933b520f320a546cecf82325b) to make the interaction between `inactive` and affected data structures more clear.
I could change it but the trouble is that hybrid where's are more strict w.r.t to shape: thinks like `where((5,), (5,1), (5,)` don't work in HybridBlocks but imperatively apparently. 
I tried for a while to get the shapes of all the auxiliary arrays and `scores_accumulated` etc. to work fine, but its adding more reshapes here and there.

sockeye/inference.py
fhieber: actually, scratch that. I misread the position you were commenting on.
I can revert this.

sockeye/inference.py
fhieber: will do

sockeye/inference.py
tdomhan: I mean it's fine, I was just wondering.

PR #427, opened 2018-06-03
[WIP, RFC] Implement multisource ensemble

sliedes:
The new --multisource parameter to sockeye.translate allows giving
different source sentences to each model in the ensemble. Moreover,
the models no longer need to have the same source vocabulary. This
makes it possible to build an ensemble of models with different source
languages but the same target language.

The source sentences should be separated by '|' in the input to
sockeye.translate.

----

While the code is basically in a working state, my first question is if there is at all interest in merging this feature or if it is out of scope for sockeye; I'm aware multisource translation is a niche feature. If there is interest, I could write a test – I'm thinking a copy of the seq copy test, but with half of the output being in one source and the other half in another source.

Possible things to pay attention to during review:

- Is separating the source sentences by '|'  a good option?
  - Also, so far I have only implemented string-based input.
- I can't see how chunking would make any sense here. Ideas?
- Where TranslatorInput previously took a list of tokens, it now takes a list of tokens for each source language, i.e. a list of lists of tokens. Is this too convoluted? What would be a cleaner way?
- I have not yet worked on supporting factors and restricted lexicons, which are largely alien features to me (indeed this is the first time I touch sockeye internals or mxnet).

until you can check this box.
- [ ] Updated CHANGELOG.md


tdomhan:
Thanks for the contribution! With respect to niche usage and whether this is useful to add I think in general as long as the behavior can be changed from the CLI and it does not make other parts of the code more complicated I do not see any reason not to include this. Others might be interested using this feature. I'll take a closer look at the changes and leave review comments in the coming days!

sockeye/inference.py
```diff
@@ -1199,7 +1227,7 @@ def _make_result(self,
 
         target_string = C.TOKEN_SEPARATOR.join(
             tok for target_id, tok in zip(target_ids, target_tokens) if target_id not in self.strip_ids)
-        attention_matrix = attention_matrix[:, :len(trans_input.tokens)]
+        attention_matrix = attention_matrix[:, :len(trans_input.tokens[0])]
```
tdomhan: Shouldn't this be done according to the longest sentence in the input?

sockeye/inference.py
```diff
@@ -1242,11 +1270,12 @@ def _encode(self, sources: mx.nd.NDArray, source_length: int) -> List[ModelState
...
         Returns a ModelState for each model representing the state of the model after encoding the source.
 
-        :param sources: Source ids. Shape: (batch_size, bucket_key, num_factors).
+        :param sources: Source ids. Shape: (batch_size, num_input_langs, bucket_key, num_factors).
         :param source_length: Bucket key.
         :return: List of ModelStates.
         """
-        return [model.run_encoder(sources, source_length) for model in self.models]
+
+        return [model.run_encoder(sources[:, i], source_length) for i, model in enumerate(self.models)]
```
tdomhan: this wouldn't work for single source ensemble models as sources would just contain the data for a single language, would it?

fhieber:
Closing due to inactivity for now. Feel free to reopen.

PR #430, opened 2018-06-08
Fix: Word based batching memory with longer source sentences.

tdomhan:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ## [1.18.22]
 ### Fixed
 - Make sure the default bucket is large enough with word based batching when the source is longer than the target (Previously
-there as an edge case where the memory usage sub-optimal with word based batching and longer source than target sentences).
+there was an edge case where the memory usage was sub-optimal with word based batching and longer source than target sentences).
 
 ## [1.18.21]
 ### Fixed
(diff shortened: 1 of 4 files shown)


tdomhan:
This regression was introduced in the following PR by me: 
https://github.com/awslabs/sockeye/pull/378 

While everything else in `define_bucket_batch_sizes` is defined on the target side we need to make sure that in scenarios where the source is large than the target the source input data of shape `(batch_size, len_source)` can be covered by the default bucket batch size. Otherwise MXNet is not able to reuse memory:
```
 Bucketing: data source has a shape [59,40,1], which is larger than already allocated shape [22,100,1]. Need to re-allocate. Consider putting default bucket key to be the bucket taking the largest input for better memory sharing.
```

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.22]
+### Fixed
+- Make sure the default bucket is large enough with word based batching when the source is longer than the target (Previously
+there as an edge case where the memory usage sub-optimal with word based batching and longer source than target sentences).
```
davvil: "there was"

davvil: "memory usage was sub-optimal"

PR #433, opened 2018-06-12
added negative constraints

mjpost:
This implements negative constraints in beam search, i.e., phrases that are not permitted to appear in the output. It works by tracking the progress of each negative constraint; if the last word is reached, the corresponding entry in the `scores` array is set to `np.inf` prior to the call to `topk()`.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -691,12 +691,22 @@ def make_input_from_json_string(sentence_id: int, json_string: str) -> Translato
                 return _bad_input(sentence_id, reason=json_string)
         else:
             factors = None
+
+        # List of phrases that must appear in the output
         constraints = jobj.get(C.JSON_CONSTRAINTS_KEY)
         if isinstance(constraints, list) and len(constraints) > 0:
```
davvil: Why the `isinstance`? Wouldn't it just work with `if constraints:`? Then you catch the case where it is `None` and len() == 0 in one go.

sockeye/lexical_constraints.py
```diff
@@ -441,12 +542,20 @@ def main():
 
         { "text": "Dies ist ein Test .", "constraints": ["This is", "test"] }
 
+    If you pass `--avoid` to the script, the constraints will be generated as negative constraints, instead:
+
+        echo -e "Dies ist ein Test .\tThis is\ttest" | python3 -m sockeye.lexical_constraints --avoid
```
davvil: Ask the german guys, but I think "Das ist ein Test" sounds better.

sockeye/lexical_constraints.py
```diff
@@ -25,6 +25,104 @@
 RawConstraintList = List[List[int]]
 
 
+class AvoidPhrase:
+    """
+    Represents a negative phrasal constraint. Tracks progress through a phrase
+    and returns a word ID if that word ID would complete the phrase.
```
davvil: Returns the word ID in the `avoid()` method.

sockeye/lexical_constraints.py
```diff
@@ -25,6 +25,104 @@
...
+
+    def consume(self, word_id: int) -> 'AvoidPhrase':
+        """
+        Consumes a word, and updates the tracking based on it. If the word is the next words in the phrase,
+        update last_consumed, otherwise, reset.
+
+        :param word_id: The word that was just generated.
+        """
+        if self.phrase[self.last_consumed + 1] == word_id:
+            return AvoidPhrase(self.phrase, self.last_consumed + 1)
```
davvil: Do you need to return a new object? Couldn't you just update your last_consumed index as in the else: case? If so, you wouldn't actually need to return anything and the loop in AvoidBatch would be more efficient.

sockeye/lexical_constraints.py
```diff
@@ -25,6 +25,104 @@
...
+        else:
+            self.last_consumed = -1
+            return self
+
+    def avoid(self) -> int:
+        """
+        Returns an integer ID that should be avoided.
```
davvil: I would write "word ID".

sockeye/lexical_constraints.py
```diff
@@ -25,6 +25,104 @@
...
+        """
+        Returns a list of indices in a 1-dimensional reshaping of the scores array that should be set to null.
+
+        :return: A list of indices.
+        """
+        to_avoid = []
+        for i, items in enumerate(self.avoid_list):
+            for phrase in items:
+                word_id = phrase.avoid()
+                if word_id:
```
davvil: Not strictly necessary, but I would be explicit here and put `if word_id > 0:', as it cannot be None.

sockeye/lexical_constraints.py
```diff
@@ -455,14 +564,30 @@ def main():
...
-        if len(constraints) > 0:
+        constraints = []
+        avoid_list = []
+        for item in restrictions:
+            if args.avoid:
+                avoid_list.append(item)
+            else:
+                constraints.append(item)
+
+        if len(constraints):
```
davvil: `if constraints:`?

sockeye/lexical_constraints.py
```diff
@@ -455,14 +564,30 @@ def main():
...
             obj['constraints'] = constraints
+        if len(avoid_list):
```
davvil: `if avoid_list:`?

sockeye/inference.py
mjpost: Hmm, good point.

sockeye/lexical_constraints.py
mjpost: Not German but agree.

sockeye/lexical_constraints.py
mjpost: Yes, clearer.

sockeye/lexical_constraints.py
mjpost: AvoidPhrase objects can be shared so long as they have the same state. This is accomplished by returning a new object every time the state advances. This way every hypothesis has its own record of its progress through each phrase.

Put another way, the AvoidPhrase object is a reference shared across hypotheses in the beam. When any hypothesis advances, it gets a new version of the object that is specific to itself.

sockeye/lexical_constraints.py
```diff
@@ -25,6 +25,106 @@
...
+        """
+        Consumes a word, and updates the tracking based on it. If the word is the next words in the phrase,
+        update last_consumed, otherwise, reset.
+
+        :param word_id: The word that was just generated.
+        """
+        if self.phrase[self.last_consumed + 1] == word_id:
+            return AvoidPhrase(self.phrase, self.last_consumed + 1)
+        elif self.last_consumed != -1:
+            return AvoidPhrase(self.phrase, self.last_consumed)
```
davvil: self.last_consumed or -1?

davvil:
It would be useful to have a test for this. Would it be easy to adapt the ones for constrained decoding?

sockeye/lexical_constraints.py
mjpost: good catch.

mjpost:
agree on the need for a test, will write one.

sockeye/lexical_constraints.py
```diff
@@ -24,106 +24,139 @@
 # Represents a list of raw constraints for a sentence. Each constraint is a list of target-word IDs.
 RawConstraintList = List[List[int]]
 
+class AvoidTrie:
```
davvil: Hmm, it is a bit of a strange Trie. The nodes at level i contains a set of phrases the of length i+1 (with root being level 0), instead of the implicit representation of the path up to the node. Is this because of some optimization?

 **Update:** Probably this is to support the avoid() function below, right?

sockeye/lexical_constraints.py
```diff
@@ -24,106 +24,139 @@
...
+
+    def step(self, word_id: int) -> 'AvoidTrie':
+        return self.children.get(word_id, None)
+
+    def final(self) -> Set[int]:
+        return self.final_ids
 
-class AvoidPhrase:
+
+class AvoidState:
```
davvil: Do you need a separate class? Can't you just store AvoidTrie (perhaps renamed as AvoidTrieNode for clarity) in AvoidBatch?

sockeye/lexical_constraints.py
mjpost: Yes, to support `avoid()`. At each trie node, we record all words that should not be generated from that state. A more principled trie would generate arcs leading to terminal states, but this is more efficient for our purposes. Arcs are only generated if they lead to internal states of a phrase that must be avoided.

sockeye/lexical_constraints.py
mjpost: The state object tracks a position in the trie for the current hypothesis. We could just keep this position, but there are a number of associated operations that make sense to bundle together in a class.

davvil: Yes, the question is if you can't bundle these operations in the trie and avoid yet another wrapper class. As you are already using a non-general, partly non-standard trie I think this would help clarity. But up to you, it's not a big issue.

mjpost: It would be simpler to have one fewer class. I'll give it some thought this evening.

mjpost:
Okay, this is now updated to allow both a global block list (`--avoid-list /path/to/list`) plus sentence-specific ones (via the JSON `avoid` keyword). I wrote test cases and rebased on master. It is ready to merge pending CR.

sockeye/inference.py
```diff
@@ -1441,6 +1473,12 @@ def _beam_search(self,
             # finished rows are inf everywhere except column zero, which holds the accumulated model score
             scores = self._update_scores.forward(scores, finished, inactive, scores_accumulated, self.inf_array, pad_dist)
 
+            # Mark entries that should be blocked as having a score of np.inf
+            if self.global_avoid_trie or any(raw_avoid_list):
+                indices = mx.nd.array(avoid_states.avoid(), ctx=self.context)
+                if any(indices):
+                    scores.reshape((self.batch_size * self.beam_size * scores.shape[-1]))[indices] = np.inf
```
fhieber: This looks strange. Is `scores` after this line reshaped or not?

mjpost: It isn't reshaped. Here is the problem: say word ID 74 is always blocked, batch_size is 1, and beam_size is 5, and the maximum vocabulary size is 10,000. We need to block ID 74 in every row of `scores` (by setting it to `np.inf`). The easiest way to do this is to reshape `scores` into a single-dimension array, and block ID 74, ID 10074, 20074, 30074, and 40074. That's what this code does.

fhieber: I understand what the code does, but was just wondering about the clarity of this expression (using fluent reshape and inplace setting of values).

mjpost: I see. 2d indices in the first place would make more sense, if MXNet supports that.

sockeye/inference.py
```diff
@@ -974,8 +987,9 @@ def __init__(self,
         assert C.PAD_ID == 0, "pad id should be 0"
         self.stop_ids = {self.vocab_target[C.EOS_SYMBOL], C.PAD_ID}  # type: Set[int]
         self.strip_ids = self.stop_ids.copy()  # ids to strip from the output
+        UNK_ID = self.vocab_target[C.UNK_SYMBOL]
```
fhieber: This is not a global variable, so it shouldn't be capitalized.

sockeye/inference.py
```diff
@@ -1198,7 +1223,14 @@ def _get_inference_input(self,
                 raw_constraints[j] = [data_io.tokens2ids(phrase, self.vocab_target) for phrase in
                                       trans_input.constraints]
 
-        return source, bucket_key, raw_constraints
+            if trans_input.avoid_list is not None:
+                raw_avoid_list[j] = [data_io.tokens2ids(phrase, self.vocab_target) for phrase in
+                                     trans_input.avoid_list]
+                UNK_ID = self.vocab_target[C.UNK_SYMBOL]
```
fhieber: same here. Why not 'self' it in the Translator constructor, similar to `self.start_ids` etc.

sockeye/inference.py
```diff
@@ -1574,6 +1574,8 @@ def _beam_search(self,
             for ms in model_states:
                 ms.sort_state(best_hyp_indices)
 
+#            self._print_beam(sequences, scores_accumulated, finished, inactive, constraints, t)
```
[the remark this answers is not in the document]
fhieber: should be removed

sockeye/inference.py
```diff
@@ -1441,6 +1482,12 @@ def _beam_search(self,
             # finished rows are inf everywhere except column zero, which holds the accumulated model score
             scores = self._update_scores.forward(scores, finished, inactive, scores_accumulated, self.inf_array, pad_dist)
 
+            # Mark entries that should be blocked as having a score of np.inf
+            if self.global_avoid_trie or any(raw_avoid_list):
+                indices = avoid_states.avoid()
+                if any(indices):
+                    scores[indices] = np.inf
```
fhieber: this is much clearer thanks. 
At some point (not scope of this PR of course) we should think about how to move the constrained/avoid logic to proper operators/gluon blocks that are potentially hybridizable. As seen recently, these imperative indexing ops are quite slow.

sockeye/lexical_constraints.py
```diff
@@ -14,7 +14,7 @@
 import copy
 import logging
 from operator import attrgetter
-from typing import List, Optional, Tuple, Set
+from typing import List, Optional, Tuple, Set, Dict
```
fhieber: [nit] could be ordered alphabetically :)

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
 
+class AvoidTrie:
+    """
+    Represents a set of phrasal constraints for an input sentence.
+    These are organized into a trie.
+    """
+    def __init__(self,
+                 raw_phrases: Optional[RawConstraintList] = None) -> None:
+        self.final_ids = set()  # type: Set[int]
+        self.children = {}      # type: Dict[int,'AvoidTrie']
```
fhieber: [nit] just 2 spaces after code for inline comments (pep8 formatting)

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+        """
+        if len(phrase) == 1:
+            self.final_ids.add(phrase[0])
+        else:
+            next_word = phrase[0]
+            if not next_word in self.children:
+                self.children[next_word] = AvoidTrie()
+            self.step(next_word).add_phrase(phrase[1:])
+
+    def step(self, word_id: int) -> 'AvoidTrie':
```
fhieber: `Optional['AvoidTrie']`

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+    def __init__(self,
+                 raw_phrases: Optional[RawConstraintList] = None) -> None:
+        self.final_ids = set()  # type: Set[int]
+        self.children = {}      # type: Dict[int,'AvoidTrie']
+
+        if raw_phrases:
+            for phrase in raw_phrases:
+                self.add_phrase(phrase)
+
+    def __str__(self):
```
fhieber: -> str

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+                 batch_size: int,
+                 beam_size: int,
+                 avoid_list: Optional[List[RawConstraintList]] = None,
+                 global_avoid_trie: Optional[AvoidTrie] = None) -> None:
+
+        self.global_avoid_states = []  # type: List[AvoidState]
+        self.local_avoid_states = []  # type: List[AvoidState]
+
+        # Store the global trie for each hypothesis
+        if global_avoid_trie:
```
fhieber: i'd explicitly check for None here

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+            self.global_avoid_states = [AvoidState(global_avoid_trie)] * batch_size * beam_size
+
+        # Store the sentence-level tries for each item in their portions of the beam
+        if avoid_list:
```
fhieber: same here. if the list is empty the for loop is not entered

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+        for i, state in enumerate(self.global_avoid_states):
+            for word_id in state.avoid():
+                if word_id > 0:
+                    to_avoid.add((i, word_id))
+        for i, state in enumerate(self.local_avoid_states):
+            for word_id in state.avoid():
+                if word_id > 0:
+                    to_avoid.add((i, word_id))
+
+        x_list = []
```
fhieber: could use some typing annotation

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+    def avoid(self) -> Tuple[List[int], List[int]]:
+        """
+        Assembles a list of per-hypothesis words to avoid. The indices are (x, y) pairs into the scores
+        array, which has dimensions (beam_size, target_vocab_size). These values are then used by the caller
+        to set these items to np.inf so they won't be selected. Words to be avoided are selected by
+        consulting both the global trie of phrases and the sentence-specific one.
+
+        :return: Two lists of indices: the x coordinates and y coordinates.
+        """
+        to_avoid = set()
```
fhieber: could use some typing annotation

test/unit/test_constraints.py
```diff
@@ -191,3 +190,98 @@ def test_constraints_init_batch(raw_constraint_lists):
...
+"""
+@pytest.mark.parametrize("global_raw_phrase_list, raw_phrase_list, batch_size, beam_size, prefix, expected_avoid", [
+    (["5 6 7 8"], None, 1, 3, '17', []),
+    (["5 6 7 12"], None, 1, 4, '5 6 7', [(0,12), (1,12), (2,12), (3,12)]),
+    (["5 6 7 8", "9"], None, 1, 2, '5 6 7', [(0,8), (0,9), (1,8), (1,9)]),
+    (["5 6 7 8", "13"], [[[10]]], 1, 2, '5 6 7', [(0,8), (0,10), (0,13), (1,8), (1,10), (1,13)]),
+    # first two hypotheses blocked on 19 (= 19 and 119), next two on 20 (= 220 and 320)
+    (None, [[[19]], [[20]]], 2, 2, '', [(0,19), (1,19), (2,20), (3,20)]),
+    # same, but also add global block list to each row
+    (['74'], [[[19]], [[20]]], 2, 2, '', [(0,19), (0,74), (1,19), (1,74), (2,20), (2,74), (3,20), (3,74)]),
```
fhieber: pep8 formatting

sockeye/inference.py
```diff
@@ -1226,8 +1226,8 @@ def _get_inference_input(self,
             if trans_input.avoid_list is not None:
                 raw_avoid_list[j] = [data_io.tokens2ids(phrase, self.vocab_target) for phrase in
                                      trans_input.avoid_list]
-                UNK_ID = self.vocab_target[C.UNK_SYMBOL]
-                if any([UNK_ID in phrase for phrase in raw_avoid_list[j]]):
+                unk_id = self.vocab_target[C.UNK_SYMBOL]
```
fhieber: self?

mjpost:
I'm not sure why these tests are failing. They pass locally and the Travis CI reported errors have nothing to do with my changes.

tdomhan:
yeah, something is up with Travis! Master is also failing now. Seems to be unrelated to this and the other PRs.

tdomhan:
somehow `default_bucket_key` is a list...

tdomhan:
looks like pyyaml suddenly loads tuples as lists (the buckets).

mjpost:
Hi folks, I think this is ready to merge if someone can give it a last pass.

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+        # Store the global trie for each hypothesis
+        if global_avoid_trie is not None:
+            self.global_avoid_states = [AvoidState(global_avoid_trie)] * batch_size * beam_size
+
+        # Store the sentence-level tries for each item in their portions of the beam
+        if avoid_list is not None:
+            for raw_phrases in avoid_list:
+                self.local_avoid_states += [AvoidState(AvoidTrie(raw_phrases))] * beam_size
+
+    def reorder(self, indices: mx.nd.array) -> None:
```
tdomhan: to be consistent with the rest of the code these should probably be ```mx.nd.NDArray``` (also below).

sockeye/lexical_constraints.py
```diff
@@ -24,6 +24,200 @@
...
+        for i, state in enumerate(self.global_avoid_states):
+            for word_id in state.avoid():
+                if word_id > 0:
+                    to_avoid.add((i, word_id))
+        for i, state in enumerate(self.local_avoid_states):
+            for word_id in state.avoid():
+                if word_id > 0:
+                    to_avoid.add((i, word_id))
+
+        x_list = []  # type: List[int]
```
tdomhan: this could be simplified to ```list(zip(*to_avoid))```, right? (generating a list of tuples)

sockeye/inference.py
```diff
@@ -1421,6 +1456,11 @@ def _beam_search(self,
         constraints = constrained.init_batch(raw_constraint_list, self.beam_size, self.start_id,
                                              self.vocab_target[C.EOS_SYMBOL])
 
+        if self.global_avoid_trie or any(raw_avoid_list):
```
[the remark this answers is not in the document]
mjpost: No, we need to maintain negative constraint if either (a) there are global constraints or (b) constraints for any sentence in the batch.

sockeye/lexical_constraints.py
mjpost: Ah, yes, thank you! Hated this bit of code.

sockeye/lexical_constraints.py
mjpost: Ok.

sockeye/inference.py
```diff
@@ -691,12 +692,22 @@ def make_input_from_json_string(sentence_id: int, json_string: str) -> Translato
...
         constraints = jobj.get(C.JSON_CONSTRAINTS_KEY)
-        if isinstance(constraints, list) and len(constraints) > 0:
+        if constraints:
             constraints = [list(data_io.get_tokens(constraint)) for constraint in constraints]
         else:
             constraints = None
-        return TranslatorInput(sentence_id=sentence_id, tokens=tokens, factors=factors, constraints=constraints)
+
+        # List of phrases to prevent from occuring in the output
+        avoid_list = jobj.get(C.JSON_AVOID_KEY)
```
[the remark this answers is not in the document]
mjpost: I added a check that it's a list.

mjpost:
Okay, I've pulled everything in.

PR #435, opened 2018-06-13
Preserving max output length for each sentence to avoid non-deterministic behavior using batching.

abdelrahmanbadawy:
(description of the change)

until you can check this box.
- [ ] Updated CHANGELOG.md


tdomhan:
there seems to be a python 3.4 incompatability:
```
E     File "/home/travis/build/awslabs/sockeye/sockeye/inference.py", line 1139
E       batch_translations = self._translate_nd(*self._get_inference_input(batch), max_output_lengths)
E                                                                                             ^
E   SyntaxError: only named arguments may follow *expression
```

sockeye/inference.py
```diff
@@ -1132,7 +1133,10 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
             if rest > 0:
                 logger.debug("Extending the last batch to the full batch size (%d)", self.batch_size)
                 batch = batch + [batch[0]] * rest
-            batch_translations = self._translate_nd(*self._get_inference_input(batch))
+            max_output_lengths = []
+            for i in range(len(batch)):
+                max_output_lengths.append(self.models[0].get_max_output_length(len(batch[i])))
+            batch_translations = self._translate_nd(*self._get_inference_input(batch), max_output_lengths)
```
tdomhan: how about adding the creation of `max_output_lengths` directly to `_get_inference_input`? Also I'd directly create the NDArray and pass that instead of the list to _beam_search.

tdomhan:
otherwise, looks good!

tdomhan:
One more thing: please update the version number and add a small section to the CHANGELOG.

sockeye/image_captioning/inference.py
```diff
@@ -141,7 +141,7 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
 
         return results
 
-    def _get_inference_input(self, image_paths: List[str]) -> Tuple[mx.nd.NDArray, int, List[Optional[constrained.RawConstraintList]]]:
+    def _get_inference_input(self, image_paths: List[str]) -> Tuple[mx.nd.NDArray, int, List[Optional[constrained.RawConstraintList]], mx.nd.NDArray]:
```
mjpost: Can you update the doc string to describe the additional return value?

sockeye/image_captioning/inference.py
```diff
@@ -150,8 +150,13 @@ def _get_inference_input(self, image_paths: List[str]) -> Tuple[mx.nd.NDArray, i
         """
         ## TODO(bazzanil): support constraints
         raw_constraints = [None for x in range(self.batch_size)]  # type: List[Optional[constrained.RawConstraintList]]
+        max_output_lengths = []
```
mjpost: This should have a type declaration. Try:

    # type: List[int]

sockeye/inference.py
```diff
@@ -1159,7 +1159,8 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
     def _get_inference_input(self,
                              trans_inputs: List[TranslatorInput]) -> Tuple[mx.nd.NDArray,
                                                                            int,
-                                                                           List[Optional[constrained.RawConstraintList]]]:
+                                                                           List[Optional[constrained.RawConstraintList]],
+                                                                           mx.nd.NDArray]:
```
mjpost: Please update doc string.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.24]
+### Fixed
+- Preserving max output length for each sentence to avoid non-deterministic behavior using batching.
```
tdomhan: maybe phrase this as: ```...to allow for identical translation both with and without batching."```

sockeye/inference.py
```diff
@@ -1193,7 +1196,10 @@ def _get_inference_input(self,
                 raw_constraints[j] = [data_io.tokens2ids(phrase, self.vocab_target) for phrase in
                                       trans_input.constraints]
 
-        return source, bucket_key, raw_constraints
+        # max output lengths extended to shape (batch_size * beam_size,)
+        max_out_lengths_extended = mx.nd.repeat(mx.nd.array(max_output_lengths, ctx=self.context, dtype='int32'), self.beam_size)
```
tdomhan: maybe just return the NDArray and do the actual `mx.nd.repeat` in `_beam_search` because it is only needed due to the fact that we have a beam. If we were to add something else, say greedy search, we could use the same method to create the input, while internally beam search would need to repeat the lengths.

tdomhan:
thanks for iterating! :)

sockeye/image_captioning/inference.py
```diff
@@ -141,17 +141,20 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
+        :return NDArray of images paths, bucket key, a list of raw constraint lists, an NDArray of maximum output lengths.
         """
         ## TODO(bazzanil): support constraints
         raw_constraints = [None for x in range(self.batch_size)]  # type: List[Optional[constrained.RawConstraintList]]
+        max_output_lengths = [] # type: List[int]
+        for i in range(len(image_paths)):
+            max_output_lengths.append(self.models[0].get_max_output_length(len(image_paths[i])))
         images = self.data_loader(image_paths, self.source_image_size)
-        return mx.nd.array(images), 0, raw_constraints
+        return mx.nd.array(images), 0, raw_constraints, mx.nd.array(max_output_lengths, ctx=self.context, dtype='int32')
```
tdomhan: actually here previously we passed the `0` as source length to `get_max_output_length`. Therefore you should do something like:
```
max_input_len = 0
max_output_lengths = mx.nd.array([self.models[0].get_max_output_length(max_input_len)] * len(image_paths[i]), ...)
return mx.nd.array(images), max_input_len, raw_constraints, max_output_lens
```

The actual value is not actually important as it is anyway overwritten through the `forced_max_output_len` flag when captioning, but just to stay consistent with the rest of of the code

sockeye/inference.py
```diff
@@ -1535,6 +1544,7 @@ def _beam_search(self,
                 if mx.nd.sum(finished).asscalar() > 0:
                     break
             else:
+                finished = mx.nd.sign(finished + (mx.nd.cast(lengths.reshape(-1), 'int32') > max_output_lengths))
```
tdomhan: I think this check should actually be `lengths >= max_output_lengths`. For a single translation at the end of the for loop lengths will be exactly equal to max_output_length (because it starts at 1).

sockeye/image_captioning/inference.py
```diff
@@ -128,8 +128,8 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
                 if self.source_root is not None:
                     path = os.path.join(self.source_root, path)
                 paths.append(path)
-            batch_translations = self._translate_nd(*self._get_inference_input(paths))
             # truncate to remove filler translations
+            batch_translations = self._translate_nd(*self._get_inference_input(paths))
```
fhieber: seems like an unnecessary change

tdomhan: the reordering with the comment? I agree.

sockeye/inference.py
```diff
@@ -1363,6 +1369,9 @@ def _beam_search(self,
         lengths = mx.nd.ones((self.batch_size * self.beam_size, 1), ctx=self.context)
```
artemsok: can lengths be created with dtype='int32' not to be casted in L1373?

sockeye/inference.py
```diff
@@ -1535,6 +1544,7 @@ def _beam_search(self,
                 if mx.nd.sum(finished).asscalar() > 0:
                     break
             else:
+                finished = mx.nd.sign(finished + (mx.nd.cast(lengths.reshape(-1), 'int32') >= max_output_lengths))
```
artemsok: matter of taste probably, but ```mx.nd.logical_or(finished, lengths.reshape(-1) >= max_output_lengths))``` might be a bit more readable

sockeye/inference.py
```diff
@@ -1159,24 +1159,28 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
         """
 
         bucket_key = data_io.get_bucket(max(len(inp.tokens) for inp in trans_inputs), self.buckets_source)
         source = mx.nd.zeros((len(trans_inputs), bucket_key, self.num_source_factors), ctx=self.context)
         raw_constraints = [None for x in range(self.batch_size)]  # type: List[Optional[constrained.RawConstraintList]]
 
+        max_output_lengths = [] # type: List[int]
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)
+            max_output_lengths.append(self.models[0].get_max_output_length(num_tokens))
```
artemsok: just wondernig: could the self.models[i] return different max output lengths? if yes, should a max be taken over them?

sockeye/inference.py
```diff
@@ -1363,6 +1369,9 @@ def _beam_search(self,
         lengths = mx.nd.ones((self.batch_size * self.beam_size, 1), ctx=self.context)
         finished = mx.nd.zeros((self.batch_size * self.beam_size,), ctx=self.context, dtype='int32')
 
+        # Extending max_output_lengths to shape (batch_size * beam_size,)
+        max_output_lengths = mx.nd.repeat(max_output_lengths, self.beam_size)
```
artemsok: what's the relation to ```max_output_length``` (w/o trailing s) now? 
would it be correct to set ```max_output_length = nd.max(max_output_lengths)``` before the repeat?

sockeye/inference.py
tdomhan: load_models makes sure they don't. That said it is indeed not clear from the code and we should potentially refactor this.

sockeye/inference.py
tdomhan: makes sense. I wasn't aware of the logical_or operator.

sockeye/inference.py
tdomhan: yes, that's the case. Maybe we should indeed make this explicit the way you wrote it.

sockeye/inference.py
fhieber: Please check if logical ops are part of mxnet1.2 already

sockeye/inference.py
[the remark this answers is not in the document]
abdelrahmanbadawy: That would cause the following error:
mxnet.base.MXNetError: Error in operator normalizefinishedhypotheses13__div0: [13:00:12] src/operator/contrib/../elemwise_op_common.h:123: Check failed: assign(&dattr, (*vec)[i]) Incompatible attr in node normalizefinishedhypotheses13__div0 at 1-th input: expected float32, got int32

I believe L1725: scores_accumulated / self.length_penalty(lengths),

sockeye/inference.py
abdelrahmanbadawy: logical_or is not part of mxnet1.2 api installed through pip

fhieber: thanks for checking. we should switch to the logical ops only with mxnet 1.3 then. We have a few other places where these would be handy.

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: either way there has to be at least one casting because lengths is used as int type and as float type when computing the length penalty.

sockeye/inference.py
```diff
@@ -1756,4 +1765,4 @@ def hybrid_forward(self, F, scores, finished, inactive, scores_accumulated, inf_
         # pad_dist. Shape: (batch*beam, vocab_size)
         pad_dist = F.concat(pad_id_scores, pad_dist)
         scores = F.where(finished + inactive, pad_dist, scores)
-        return scores
```
fhieber: missing newline at EOF

sockeye/inference.py
```diff
@@ -1731,7 +1741,6 @@ def hybrid_forward(self, F, finished, inactive, lengths, best_word_indices):
         finished = ((best_word_indices == self.pad_id) + (best_word_indices == self.eos_id))
         return finished, finished_or_inactive, lengths
 
-
```
fhieber: pep8: 2 newlines between classes

sockeye/inference.py
```diff
@@ -1164,24 +1164,28 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
-                bucket key, list of raw constraint lists.
+                bucket key, list of raw constraint lists,
+                NDArray of maximum output lengths.
         """
 
         bucket_key = data_io.get_bucket(max(len(inp.tokens) for inp in trans_inputs), self.buckets_source)
         source = mx.nd.zeros((len(trans_inputs), bucket_key, self.num_source_factors), ctx=self.context)
         raw_constraints = [None for x in range(self.batch_size)]  # type: List[Optional[constrained.RawConstraintList]]
 
+        max_output_lengths = [] # type: List[int]
```
fhieber: pep8: two spaces

sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.18.28'
+__version__ = '1.18.29'
```
fhieber: missing newline

tdomhan:
Resolved the version conflict. Waiting for Travis, then merging.

PR #440, opened 2018-06-18
Top-K lexicon update: defaults, files for running fast_align in contrib.

mjdenkowski:
This expands our functionality for learning and using top-k lexicons to speed up CPU decoding.  The idea is that users should be able to take advantage of this functionality without manually installing and running software in addition to Sockeye.
- Add convenience files for running `fast_align` to contrib.  This is done via Docker as there are unfortunately no Python bindings for `fast_align`.  There is also a short convenience script for running `fast_align` to produce lex tables.
- Add reference to contrib/fast_align to the lexicon module help message.
- Update default size of top-k lexicon to 200 for better quality.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.24]
+### Added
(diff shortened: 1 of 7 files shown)


mjdenkowski:
In practice, it has be up to 1.5 BLEU.  We've found 200 to be a good default value since it's still significantly faster than running without vocabulary selection but delivers about the highest BLEU score short of going back to the full size (going up to 500 for example rarely improves further).

PR #447, opened 2018-06-19
Make MxNet requirement >= rather than == since master is already mark…

mkolod:
Related Github issue: #447.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/requirements.gpu-cu80.txt
+++ b/requirements.gpu-cu80.txt
@@ -1,4 +1,4 @@
 pyyaml
-mxnet-cu80==1.2.0
+mxnet-cu80>=1.2.0
 numpy>=1.12
 typing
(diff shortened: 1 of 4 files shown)


davvil:
This is dangerous as new releases of mxnet are not always backwards compatible. This is why we have the ==.

tdomhan:
For now we always manually updated after manually verifying everything works when upgrading and I'd still lean towards continuing that practice. If someone wants to use a newer version of MXNet with Sockeye they can do so by passing the `--no-deps` flag to pip and make that conscious choice.

fhieber:
Yes, lets stick with release version as dependencies

mkolod:
Sounds good.

PR #448, opened 2018-06-20
updated tutorial parameters to use RNN

mjpost:
Added `--encoder rnn --decoder rnn` since the default is now the transformer models. Also generate the data underneath `data/` (for easier cleanup) and line-fixed the README.


fhieber:
Default number of layers is 6 now as well. Probably not needed for seqcopy.

PR #454, opened 2018-06-22
Use version of NDArray split that always returns a list.

mjdenkowski:
This fixes an issue where inference was silently breaking when using vocabulary restriction, batch decoding, and a single source factor.  The source was the behavior of [`mxnet.ndarray.split`](https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.split) that returns a list when num_outputs is greater than 1, but the individual NDArray that would be element 0 when num_outputs is 1.  This was leading the code to pull element 0 of the NDarray instead of the NDArray itself.  This commit adds a wrapper for `split` that always returns a list for consistent behavior.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -158,7 +158,7 @@ class InferenceModel(model.SockeyeModel):
 
         def sym_gen(source_seq_len: int):
             source = mx.sym.Variable(C.SOURCE_NAME)
-            source_words = utils.split(source, num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
+            source_words = source.split(num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
             source_length = utils.compute_lengths(source_words)
 
             # source embedding
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -158,7 +158,7 @@ class InferenceModel(model.SockeyeModel):
 
         def sym_gen(source_seq_len: int):
             source = mx.sym.Variable(C.SOURCE_NAME)
-            source_words = source.split(num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
+            source_words = utils.split(source, num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
             source_length = utils.compute_lengths(source_words)
 
             # source embedding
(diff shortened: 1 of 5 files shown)


sockeye/inference.py
```diff
@@ -158,7 +158,7 @@ def _get_encoder_module(self) -> Tuple[mx.mod.BucketingModule, int]:
 
         def sym_gen(source_seq_len: int):
             source = mx.sym.Variable(C.SOURCE_NAME)
-            source_words = source.split(num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
+            source_words = utils.split(source, num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
```
fhieber: I think we don't need this change here, as in the symbolic API split seems to return always an 'indexable' Symbol/SliceChannel. We do these source-factor related splits also in other places of the code and it works there just fine. Also, the util function isn't typed for symbols and I am surprised `data.split` (aka using the fluent method) works for symbols).

```python
mx.sym.split(mx.sym.Variable('x'), num_outputs=1, axis=2)[0].eval(x=mx.nd.ones((2,2,2)))
```
This throws an error:
```python
mx.sym.split(mx.sym.Variable('x'), num_outputs=1, axis=2)[1].eval(x=mx.nd.ones((2,2,2)))
```

sockeye/utils.py
```diff
@@ -860,3 +860,28 @@ def uncast_conditionally(data: mx.sym.Symbol, dtype: str) -> mx.sym.Symbol:
...
+    :param data: The input.
+    :param num_outputs: Number of splits. Note that this should evenly divide
+                        the length of the axis.
+    :param axis: Axis along which to split.
+    :param squeeze_axis: If true, Removes the axis with length 1 from the shapes
+                         of the output arrays.
+    :return: List of NDArrays resulting from the split.
+    """
+    ndarray_or_list = data.split(num_outputs=num_outputs, axis=axis, squeeze_axis=squeeze_axis)
+    if num_outputs == 1:
```
fhieber: Would it make sense to avoid the split altogether when num_outputs is 1 anyway? If squeeze_axis==True, one only would need a reshape/[squeeze](https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.squeeze) which is essentially a no-op.

sockeye/inference.py
mjdenkowski: Good catch!

sockeye/utils.py
mjdenkowski: Another good point.  I think it's a toss-up between staying as close to the original as possible versus micro-optimizing the call.  Since we're now using this in just one place, called once per batch, I would lean toward keeping it this way for clarity.

fhieber: yes, probably not worth the additional complexity.

PR #481, opened 2018-07-18
Ignore pylint warning about bare except statement

fhieber:


tdomhan:
thanks for the fix. let's merge to unblock Travis and then check whether the try block can be removed.

Issue #478, opened 2018-07-16
pylint==2.0.0 complains about bare `except` statement


fhieber:
travis has updated pylint to verison 2.0.0 which [complains](https://travis-ci.org/awslabs/sockeye/builds/404257699) about the following:
```
************* Module sockeye.utils
sockeye/utils.py:632:16: E0705: The except handler raises immediately (try-except-raise)
pylint found errors in the sockeye package
```

The corresponding line of code in `acquire_gpus()` is::
```python
[...]
            if master_lock is not None and not any_failed:
                try:
                    yield acquired_gpus
                except:
                    raise
                return
[...]
```
@tdomhan do you know what kind of exceptions to expect here? Changing to `except Exception` is equally objectionable for pylint.

fhieber:
intermediate "solution" in #481

tdomhan:
I wonder if the raise is actually necessary. I remember running into some unintuitive behavior when exceptions were thrown in the generator.

fhieber:
no objections from my side to removing it :)

tdomhan:
ok, let's test if we can potentially remove it.

mjpost:
travis is now at pylint==2.0.1, which returns a number of new errors in the image_captioning module:

```bash
sockeye/image_captioning/train.py:268:20: E1101: Instance of 'Namespace' has no 'seed' member (no-member)
sockeye/image_captioning/train.py:271:36: E1101: Instance of 'Namespace' has no 'output' member (no-member)
sockeye/image_captioning/train.py:277:43: E1101: Instance of 'Namespace' has no 'quiet' member (no-member)
sockeye/image_captioning/train.py:282:45: E1101: Instance of 'Namespace' has no 'max_seq_len' member (no-member)
sockeye/image_captioning/train.py:290:53: E1101: Instance of 'Namespace' has no 'device_ids' member (no-member)
sockeye/image_captioning/train.py:291:50: E1101: Instance of 'Namespace' has no 'use_cpu' member (no-member)
sockeye/image_captioning/train.py:292:65: E1101: Instance of 'Namespace' has no 'disable_device_locking' member (no-member)
sockeye/image_captioning/train.py:293:51: E1101: Instance of 'Namespace' has no 'lock_dir' member (no-member)
sockeye/image_captioning/train.py:295:11: E1101: Instance of 'Namespace' has no 'batch_type' member (no-member)
[...]

PR #486, opened 2018-07-24
fixed bug with negative constraints, added integration test

mjpost:
Check for negative constraints didn't work when there are none. Added integration test that would have caught this.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1493,7 +1493,7 @@ class Translator:
             # Mark entries that should be blocked as having a score of np.inf
             if self.global_avoid_trie or any(raw_avoid_list):
                 block_indices = avoid_states.avoid()
-                if len(block_indices[0]) > 0:
+                if len(block_indices) > 0:
                     scores[block_indices] = np.inf
 
             # (3) Get beam_size winning hypotheses for each sentence block separately. Only look as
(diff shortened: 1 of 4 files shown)


mjpost:
The failing tests all have to do with `pylint==2.0.1` which is now on travis. Perhaps related to #478.

sockeye/inference.py
```diff
@@ -690,22 +690,29 @@ def make_input_from_json_string(sentence_id: int, json_string: str) -> Translato
...
-            constraints = None
 
-        # List of phrases to prevent from occuring in the output
-        avoid_list = jobj.get(C.JSON_AVOID_KEY)
+        # If there is overlap between positive and negative constraints, assume the user wanted
+        # the words, and so remove them from the avoid_list (negative constraints)
+        if constraints is not None and avoid_list is not None:
+            overlap = set(constraints).intersection(set(avoid_list))
+            if len(overlap) > 0:
+                new_avoid_list = []
```
tdomhan: shorter:
```
new_avoid_list = list(avoid_set.difference(overlap))
```

test/common.py
```diff
@@ -204,7 +205,11 @@ def tmp_digits_dataset(prefix: str,
                         start_pos = 0
                         end_pos = min(target_len, 3)
                         constraint = ' '.join(target_words[start_pos:end_pos])
-                        new_source['constraints'] = [constraint]
+                        # Alternate between positive and negative constraints
```
tdomhan: should we add the case of both here as well?

sockeye/inference.py
mjpost: Ah, much nicer.

test/common.py
mjpost: I think we should keep them separate, just because I don't want to always test both.

tdomhan: ok, sure. We could create a third branch add either constraints, avoid, or both.

tdomhan:
let's merge after travis.

PR #487, opened 2018-07-24
simplified and corrected length calculation (fixes #483)

### LINKED ISSUE #483: hypotheses lengths off by one
It seems that the current beam search implementation (`_beam_search`) does not compute hypotheses length correctly. This can be seen when printing `sequences` and `lengths` just before returning from `_beam_search()`: the length is always one short of the actual non-zero values in sequences:

```
> sequences[0]
[    2    92  7741   907  2710   295    28     7  7741   907  2710     4
     7   383 10174  7324  9735   286   671   564  4085    71    22    98
 10671     4    13   782   485     8  6323  8665     4     8    23     8
  2976  1037  2448  3577    10   294   575  4664    13    34  3898   608
    43     5     3     0]
<NDArray 52 @cpu(0)>

> lengths[0]
[50.]
<NDArray 1 @cpu(0)>
```
`3` is the EOS id and the length of this hypothesis is clearly 51.

Let me try to explain the problem below. In practice, it doesn't matter since `_assemble_translation()` uses the length value to cut off the EOS symbol; the following EOS stripping logic in `_make_result()` is then redundant. But of course it shouldn't be and I think `_beam_search()` should return consistent data structures.

Here's the issue; I think this was introduced with length normalization only done for finished hypotheses:
1. `lengths` is initialized with 1s to account for the initial data in sequences, namely the BOS symbol
2. In each beam search iteration: as soon as we run topk on hypotheses scores, we effectively have expanded an hypothesis and its length increases by 1.
3. However, we are not using the updated length to potentially normalize the hypothesis if its finished. We check the newly acquired best_word_indices for EOS/PAD to compute a current `finished` status, and if so, **we normalize with the OLD length**.
4. This doesn't yet explain why we end beam_search with a length off by 1. This must somehow happen because the `length` array is only updated for hypotheses that are not finished and active. Since `finished` was just recomputed in step 3, the problem must lie in the `inactive` logic or the fact that right after we update `lengths`, we recompute `finished` once again and determine breaking out of the beam search loop based on that.

I am not 100% sure whats wrong here, probably the order of operations isn't entirely correct, but I also don't fully understand the whole `inactive` logic introduced. @mjpost could you comment on this?

At a minimum, we should make sure to normalize with correct length values. At best, we should make sure to return a `lengths` array from `_beam_search` that is consistent with `sequences`. With that, `_assemble_translation` would not silently cut off the last token (EOS), which by design should be handled later by `_make_result()`.

mjpost:
Fixes and simplifies length calculation. If the code looks good I'll update CHANGELOG etc. Needs rebased on pylint fix.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -896,46 +896,39 @@ class LengthPenalty(mx.gluon.HybridBlock):
         return self.hybrid_forward(None, lengths)
 
 
-def _concat_translations(translations: List[Translation], start_id: int, stop_ids: Set[int],
+def _concat_translations(translations: List[Translation], stop_ids: Set[int],
                          length_penalty: LengthPenalty) -> Translation:
     """
     Combine translations through concatenation.
 
     :param translations: A list of translations (sequence starting with BOS symbol, attention_matrix), score and length.
-    :param start_id: The EOS symbol.
-    :param translations: The BOS symbols.
+    :param translations: The EOS symbols.
     :return: A concatenation if the translations with a score.
     """
     # Concatenation of all target ids without BOS and EOS
-    target_ids = [start_id]
+    target_ids = []
     attention_matrices = []
     beam_histories = []  # type: List[BeamHistory]
     for idx, translation in enumerate(translations):
-        assert translation.target_ids[0] == start_id
         if idx == len(translations) - 1:
-            target_ids.extend(translation.target_ids[1:])
-            attention_matrices.append(translation.attention_matrix[1:, :])
+            target_ids.extend(translation.target_ids)
+            attention_matrices.append(translation.attention_matrix)
         else:
             if translation.target_ids[-1] in stop_ids:
-                target_ids.extend(translation.target_ids[1:-1])
-                attention_matrices.append(translation.attention_matrix[1:-1, :])
+                target_ids.extend(translation.target_ids[:-1])
+                attention_matrices.append(translation.attention_matrix[:-1, :])
             else:
-                target_ids.extend(translation.target_ids[1:])
-                attention_matrices.append(translation.attention_matrix[1:, :])
+                target_ids.extend(translation.target_ids)
+                attention_matrices.append(translation.attention_matrix)
         beam_histories.extend(translation.beam_histories)
 
     # Combine attention matrices:
     attention_shapes = [attention_matrix.shape for attention_matrix in attention_matrices]
-    # Adding another row for the empty BOS alignment vector
-    bos_align_shape = np.asarray([1, 0])
-    attention_matrix_combined = np.zeros(np.sum(np.asarray(attention_shapes), axis=0) + bos_align_shape)
-
-    # We start at position 1 as position 0 is for the BOS, which is kept zero
-    pos_t, pos_s = 1, 0
+    attention_matrix_combined = np.zeros(np.sum(np.asarray(attention_shapes), axis=0))
+    pos_t, pos_s = 0, 0
     for attention_matrix, (len_t, len_s) in zip(attention_matrices, attention_shapes):
         attention_matrix_combined[pos_t:pos_t + len_t, pos_s:pos_s + len_s] = attention_matrix
-        pos_t += len_t
-        pos_s += len_s
+        pos_s, pos_t = pos_s + len_s, pos_t + len_t
 
     # Unnormalize + sum and renormalize the score:
     score = sum(translation.score * length_penalty.get(len(translation.target_ids))
@@ -1245,9 +1238,8 @@ class Translator:
         :param translation: The translation + attention and score.
         :return: TranslatorOutput.
         """
-        # remove special sentence start symbol (<s>) from the output:
-        target_ids = translation.target_ids[1:]
-        attention_matrix = translation.attention_matrix[1:, :]
+        target_ids = translation.target_ids
+        attention_matrix = translation.attention_matrix
 
         target_tokens = [self.vocab_target_inv[target_id] for target_id in target_ids]
 
@@ -1269,7 +1261,7 @@ class Translator:
         :param translations: A list of translations (sequence, attention_matrix), score and length.
         :return: A concatenation if the translations with a score.
         """
-        return _concat_translations(translations, self.start_id, self.stop_ids, self.length_penalty)
+        return _concat_translations(translations, self.stop_ids, self.length_penalty)
 
     def _translate_nd(self,
                       source: mx.nd.NDArray,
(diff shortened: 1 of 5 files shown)


sockeye/inference.py
```diff
@@ -1033,21 +1026,16 @@ def __init__(self,
...
         self._sort_by_index.hybridize()
 
-        self._normalize_finished = NormalizeFinishedHypotheses(pad_id=C.PAD_ID,
-                                                               eos_id=self.vocab_target[C.EOS_SYMBOL],
-                                                               length_penalty_alpha=self.length_penalty.alpha,
-                                                               length_penalty_beta=self.length_penalty.beta)
-        self._normalize_finished.initialize(ctx=self.context)
-        self._normalize_finished.hybridize()
+        self._update_finished = NormalizeAndUpdateFinished(pad_id=C.PAD_ID,
+                                                              eos_id=self.vocab_target[C.EOS_SYMBOL],
```
tdomhan: is the indentation correct here?

sockeye/inference.py
```diff
@@ -1619,7 +1603,7 @@ def _beam_search(self,
         scores_accumulated = scores_accumulated.take(best_hyp_indices)
         constraints = [constraints[x] for x in best_hyp_indices.asnumpy()]
 
-        return sequences, attentions, scores_accumulated, lengths, constraints, beam_histories
+        return sequences[:,1:], attentions[:,1:,:], scores_accumulated, lengths, constraints, beam_histories
```
tdomhan: maybe add a comment that we're removing the sentence start symbol here

mjpost: Fixed. We could also just not add these columns when we initialize sequences and attentions, but it seems you can't create and then later extend an `ndarray` with a 0 dimension.

sockeye/inference.py
```diff
@@ -1568,6 +1551,7 @@ def _beam_search(self,
 
             # (7b) optionally save beam history
             if self.store_beam:
+                finished_or_inactive = mx.nd.clip(data=finished + inactive, a_min=0, a_max=1)
                 unnormalized_scores = mx.nd.where(finished_or_inactive,
                                                   scores_accumulated * self.length_penalty(lengths - 1),
```
tdomhan: you also need to update this as otherwise we now had infinity for output-type beam_store

sockeye/inference.py
```diff
@@ -896,46 +896,39 @@ def get(self, lengths: Union[mx.nd.NDArray, int, float]) -> Union[mx.nd.NDArray,
...
-
-    # We start at position 1 as position 0 is for the BOS, which is kept zero
-    pos_t, pos_s = 1, 0
+    attention_matrix_combined = np.zeros(np.sum(np.asarray(attention_shapes), axis=0))
+    pos_t, pos_s = 0, 0
     for attention_matrix, (len_t, len_s) in zip(attention_matrices, attention_shapes):
         attention_matrix_combined[pos_t:pos_t + len_t, pos_s:pos_s + len_s] = attention_matrix
-        pos_t += len_t
-        pos_s += len_s
+        pos_s, pos_t = pos_s + len_s, pos_t + len_t
```
fhieber: i liked the previous version more :)

sockeye/inference.py
```diff
@@ -1250,9 +1238,8 @@ def _make_result(self,
         :param translation: The translation + attention and score.
         :return: TranslatorOutput.
         """
-        # remove special sentence start symbol (<s>) from the output:
```
fhieber: the docstring of this class method needs to be updated

sockeye/inference.py
```diff
@@ -1619,7 +1603,8 @@ def _beam_search(self,
         scores_accumulated = scores_accumulated.take(best_hyp_indices)
         constraints = [constraints[x] for x in best_hyp_indices.asnumpy()]
 
-        return sequences, attentions, scores_accumulated, lengths, constraints, beam_histories
+        # Remove the <s> column from both sequences and attentions
+        return sequences[:,1:], attentions[:,1:,:], scores_accumulated, lengths, constraints, beam_histories
```
fhieber: pep8 spacing

sockeye/inference.py
```diff
@@ -1400,16 +1387,15 @@ def _beam_search(self,
 
         best_word_indices = mx.nd.full((self.batch_size * self.beam_size,), val=self.start_id, ctx=self.context,
                                        dtype='int32')
-        # growing sequences: (batch_size * beam_size, 1), pre-filled with <s> symbols on index 0
-        sequences = mx.nd.full((self.batch_size * self.beam_size, 1), val=self.start_id, ctx=self.context,
-                               dtype='int32')
+        # The growing sequences representing the complete history for each hypothesis: (batch_size * beam_size, 1)
+        sequences = mx.nd.zeros((self.batch_size * self.beam_size, 1), ctx=self.context, dtype='int32')
```
fhieber: I think this obfuscates the fact that the 0th column represents `<s>`. Could you add a comment that the 0th column is stripped/not used in the end anyway?

fhieber:
Another question: do we need to update the logic for any output handlers / attention matrix ploting functions etc?

tdomhan:
note: I did test the align_plot, the beam_store and the  the standard handler, which all worked fine.

tdomhan:
I noticed that the align_plot currently misses the `EOS` from the source side we introduced recently. However, this was the case before and after the length check.

sockeye/inference.py
mjpost: I think it's already correct?

mjpost:
Okay I addressed the latest round of comments, and also snuck in a small documentation fix.

sockeye/inference.py
fhieber: I mean line 1236

sockeye/inference.py
```diff
@@ -1532,11 +1520,14 @@ def _beam_search(self,
...
             # (5) Normalize the scores of newly finished hypotheses. Note that after this until the
-            # next call to topk(), hypotheses may not be in sorted order.
-            finished, scores_accumulated = self._normalize_finished.forward(best_word_indices,
-                                                                            finished,
-                                                                            scores_accumulated,
-                                                                            lengths)
+            # next call to topk(), hypotheses may not be in sorted order. Don't update `finished` yet;
+            # we need to include newly-finished (on this time-step) items for normalization and pruning,
+            # but we need the old values of `finished` to update lengths first, using the old values
+            finished, scores_accumulated, lengths = self._update_finished.forward(best_word_indices,
```
fhieber: I don't get the comment about not updating `finished` yet. From the perspective in `_beam_search`, `finished` is returned from the block and is in fact different than the `finished` passed in as an argument. Moreover, the block is called `NormalizeAndUpdateFinished`.

mjpost: This comment applied when we were going to update the length by one, to include `<s>`. Since we decided not to, this comment no longer applies. Deleting.

sockeye/inference.py
mjpost: Still don't understand. The comment now reads:

```
        Returns a translator result from generated target-side word ids, attention matrix, and score.
        Strips stop ids from translation string.

        :param trans_input: Translator input.
        :param translation: The translation + attention and score.
        :return: TranslatorOutput.
```

which is correct.

fhieber: True, your are correct. I thought the stop id stripping was removed.

mjpost: Okay, cool. I had been willing to be 2:1 that you were right :)

fhieber: I should have insisted on it just a little bit more :D

PR #495, opened 2018-07-31
avoid_list now supports <s> in negative constraints

mjpost:
Simple fix that lets you add `<s>` as the beginning of a sequence to avoid.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.43]
+### Fixed
+- `<s>` now supported in negative constraints (as the start of a multi-token sequence)
+
 ## [1.18.42]
 ### Changed
 - Simplified gluon blocks for length calculation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
 ## [1.18.43]
-### Fixed
+### Added
 - `<s>` now supported as the first token in a multi-word negative constraint
   (e.g., '<s> I think' to prevent a sentence from starting with 'I think')
+### Fixed
 - Bugfix in resetting the state of a multiple-word negative constraint
 
 ## [1.18.42]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 
 ## [1.18.43]
 ### Fixed
-- `<s>` now supported in negative constraints (as the start of a multi-token sequence)
+- `<s>` now supported as the first token in a multi-word negative constraint
+  (e.g., '<s> I think' to prevent a sentence from starting with 'I think')
+- Bugfix in resetting the state of a multiple-word negative constraint
 
 ## [1.18.42]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ## [1.18.43]
 ### Added
 - `<s>` now supported as the first token in a multi-word negative constraint
-  (e.g., '<s> I think' to prevent a sentence from starting with 'I think')
+  (e.g., `<s> I think` to prevent a sentence from starting with `I think`)
 ### Fixed
 - Bugfix in resetting the state of a multiple-word negative constraint
 
--- a/contrib/sacrebleu/sacrebleu.py
+++ b/contrib/sacrebleu/sacrebleu.py
@@ -37,7 +37,7 @@ from typing import List, Iterable, Tuple
 import math
 import unicodedata
 
-VERSION = '1.3.0'
+VERSION = '1.2.10'
 
 try:
     # SIGPIPE is not available on Windows machines, throwing an exception.
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1461,6 +1461,7 @@ class Translator:
             avoid_states = constrained.AvoidBatch(self.batch_size, self.beam_size,
                                                   avoid_list=raw_avoid_list,
                                                   global_avoid_trie=self.global_avoid_trie)
+            avoid_states.consume(best_word_indices)
 
         # Records items in the beam that are inactive. At the beginning (t==1), there is only one valid or active
         # item on the beam for each sentence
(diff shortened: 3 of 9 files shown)


contrib/sacrebleu/sacrebleu.py
```diff
@@ -37,7 +37,7 @@
 import math
 import unicodedata
 
-VERSION = '1.3.0'
+VERSION = '1.2.10'
```
fhieber: not sure if sneaking this in is the right way. Why not make a small separate PR and explain the downgrade?

sockeye/inference.py
```diff
@@ -1461,6 +1461,7 @@ def _beam_search(self,
             avoid_states = constrained.AvoidBatch(self.batch_size, self.beam_size,
                                                   avoid_list=raw_avoid_list,
                                                   global_avoid_trie=self.global_avoid_trie)
+            avoid_states.consume(best_word_indices)
```
fhieber: any chance for a test? :)

tdomhan:
what does it mean to allow to avoid `<s>`?

tdomhan:
is the idea to have an avoid phrase to only apply at the beginning of the sentence?

fhieber:
I would think that this enables users to say that the translation should not start with "I think", when specifying an avoid constraint as such: "<s> I think".

sockeye/inference.py
mjpost: I can't sneak anything by you! Okay, I'll add one.

mjpost:
Yes, exactly. This allows you to avoid a prefix, e.g., "<s> I think". Trying to specify "<s>" has no effect.

tdomhan:
Thanks! Could you add that explanation to the CHANGELOG?

mjpost:
Okay, I added test cases and refactored them in the process. Adding the test cases uncovered a bug.

This had to be an integration test since it is testing the output of `_beam_search()`. But the sequence copy integration test was already getting unwieldy, and was never ideal for testing constraints. The refactored tests (a) remove testing constraints from `test_seq_copy_int.py` and move it into `test_constraints_int.py`. There, we first decode in an unconstrained setting, then add positive constraints, and negative constraints, and ensure that the proper thing happens in the output.

This uncovered a bug. Say you have a negative constraint of "6 5 5". This wouldn't prevent the output "6 6 5 5" from being generated, because the state wasn't getting properly reset when generating the second "6". The code was resetting the state to the beginning of the trie, but actually, it needed to be taking the first step from the root of the trie.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.43]
+### Fixed
+- `<s>` now supported as the first token in a multi-word negative constraint
```
fhieber: [nit] could make this a bullet point and the item in 1.18.42 as well?

sockeye/lexical_constraints.py
```diff
@@ -116,10 +116,21 @@ def consume(self, word_id: int) -> 'AvoidState':
         """
         Consumes a word, and updates the state based on it. Returns new objects on a state change.
 
+        The next state for a word can be tricky. Here are the cases:
+        (1) If the word is found in our set of outgoing child arcs, we take that transition.
+        (2) If the word is not found, and we are not in the root state, we need to reset.
+            This means we pretend we were in the root state, and see if we can take a step
+        (3) Otherwise, if we are not already the root state (i.e., we were partially through
```
fhieber: "if we are not already IN the root state"

sockeye/lexical_constraints.py
```diff
@@ -116,10 +116,21 @@ def consume(self, word_id: int) -> 'AvoidState':
...
+            the trie), we need to create a new object to track the state, in the root state
```
fhieber: "create a new object in root state" maybe?

test/common.py
```diff
@@ -375,6 +347,61 @@ def run_train_translate(train_params: str,
...
+                                                        _TRANSLATE_PARAMS_COMMON.format(model=model_path,
+                                                                                        input=new_test_source_path,
+                                                                                        output=out_path_constrained,
+                                                                                        quiet=quiet_arg),
+                                                        translate_params)
+
+                with patch.object(sys, "argv", params.split()):
+                    sockeye.translate.main()
+
+                print('JSON_INPUT', open(new_test_source_path).readlines(),
```
fhieber: could we change this to a logging statement? with pytest, logging statements are easier to track than stdout output.

test/common.py
```diff
@@ -375,6 +347,61 @@ def run_train_translate(train_params: str,
...
+                      'CONSTRAINED', open(out_path_constrained).readlines(),
+                      'UNCONSTRAINED', open(out_path).readlines(), sep='\n')
+                for json_input, constrained_out, unconstrained_out in zip(open(new_test_source_path), open(out_path_constrained), open(out_path)):
+                    jobj = json.loads(json_input)
+                    print('JSON_INPUT', json_input, 'CONSTRAINED', constrained_out, 'UNCONSTRAINED', unconstrained_out, sep='\n')
```
fhieber: same here

test/common.py
```diff
@@ -375,6 +347,61 @@ def run_train_translate(train_params: str,
...
+                    if jobj.get(constraint_type, None) == None:
+                        assert constrained_out == unconstrained_out
```
fhieber: would it be helpful to add error messages to these `assert` statements?

test/common.py
mjpost: Oh, this should be removed entirely.

test/common.py
mjpost: I can add comments, but none of the other asserts have comments, and I'm not sure exactly what to put...

CHANGELOG.md
mjpost: Sorry, I'm not sure I understand this one. This improvement doesn't apply to 1.18.42, right?

test/common.py
fhieber: lets leave as is, was just wondering about whether the if cases are clear, and/or what we are asserting. but after a second look its clear to me.

CHANGELOG.md
fhieber: I should be more clear: the description is missing a dash "-", so does the previous entry in the change log (length calculation).

mjpost:
Running system tests, shouldn't be affected, will wait to merge till they finish.

CHANGELOG.md
mjpost: Where is the dash missing? This is what I see right now:

```
## [1.18.43]
### Added
- `<s>` now supported as the first token in a multi-word negative constraint
  (e.g., '<s> I think' to prevent a sentence from starting with 'I think')
### Fixed
- Bugfix in resetting the state of a multiple-word negative constraint

## [1.18.42]
### Changed
- Simplified gluon blocks for length calculation
```

mjpost:
System tests pass.

Issue #479, opened 2018-07-18
How to measure the BLEU of training/translation


xinyu-intel:
Hi, 
I just trained a 8 layer rnn model and got the following result:
```
python -m sockeye.train -s corpus.tc.BPE.de \
                        -t corpus.tc.BPE.en \
                        -vs newstest2016.tc.BPE.de \
                        -vt newstest2016.tc.BPE.en \
                        --num-embed 512 \
                        --rnn-num-hidden 512 \
                        --rnn-attention-type dot \
                        --embed-dropout=0.2 \
                        --rnn-decoder-hidden-dropout=0.2 \
                        --max-seq-len 50 \
                        --decode-and-evaluate 500 \
                        --batch-size 128 \
			--batch-type sentence \
                        -o gnmt_model \
			--optimized-metric bleu \
                        --initial-learning-rate=0.0001 \
                        --learning-rate-reduce-num-not-improved=8 \
                        --learning-rate-reduce-factor=0.7 \
                        --weight-init xavier --weight-init-scale 3.0 \
                        --weight-init-xavier-factor-type avg \
			--lock-dir ~/.temp/ \
                        --num-layers 8:8 \
                        --device-ids 0 1
```
```
[2018-07-18:00:56:34:INFO:sockeye.training:fit] Training finished. Best checkpoint: 60. Best validation bleu: 0.746082
[2018-07-18:00:56:34:INFO:sockeye.utils:__exit__] Releasing GPU 1.
[2018-07-18:00:56:34:INFO:sockeye.utils:__exit__] Releasing GPU 0.
```
Evaluate translation:
```
[INFO:__main__] bleu	(s_opt)	chrf	(s_opt)
0.171	(-)	0.484	(-)
```
Is this bleu value correct? It is far away from the bleu value(20+) from the [paper](https://arxiv.org/abs/1712.05690).

By the way, how to use sockeye to build a GNMT model to align with the tensorflow's [config](https://github.com/tensorflow/nmt/blob/master/nmt/standard_hparams/wmt16_gnmt_8_layer.json)?

tdomhan:
The exact training parameters from the models trained for the paper can be found here: 
https://github.com/awslabs/sockeye/tree/arxiv_1217/arxiv/code/rnn/sockeye 

How did you run the evaluation exactly? Ideally you will need to detokenize, merge BPE-based sub-words and then run `sacrebleu` on the output.

[...]

PR #499, opened 2018-08-02
Add GNMT Config to Autopilot

xinyu-intel:
Hi, we trained a de-en gnmt model with 8 layers LSTM encoder & decoder using sockeye. The validation bleu on `newstests2016.de` has reached up to 26.91 both on CPU and GPU. Also, the translate performance is very good on CPU (around 24 sentences/sec on Skylake 8180 one socket) after some optimizations. We are going to add the model config to autopilot to enable out-of-box GNMT task.
@pengzhao-intel @TaoLv
until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/contrib/autopilot/autopilot.py
+++ b/contrib/autopilot/autopilot.py
@@ -43,7 +43,7 @@ from contrib.autopilot.tasks import TEXT_UTF8_RAW_BITEXT_REVERSE, TEXT_REQUIRES_
 from contrib.autopilot.tasks import TEXT_UTF8_TOKENIZED
 from contrib.autopilot.tasks import RAW_FILES
 from contrib.autopilot.tasks import Task, TASKS
-from contrib.autopilot.models import MODELS, MODEL_NONE, MODEL_TEST_ARGS
+from contrib.autopilot.models import MODELS, MODEL_NONE, MODEL_GNMT, MODEL_TEST_ARGS
 from contrib.autopilot.models import DECODE_ARGS, DECODE_STANDARD, DECODE_GNMT
 from contrib.autopilot import third_party
 
@@ -721,8 +721,8 @@ def run_steps(args: argparse.Namespace):
 
     logging.info("=== Train translation model ===")
     logging.info("Model: %s", args.model)
-    if args.model == "gnmt_like":
-        logging.info("WARNING: This is an 8 layer LSTMs model which can resembles the 'GNMT' architecture.")
+    if args.model == MODEL_GNMT:
+        logging.info("NOTE: This is an 8 layer LSTM model similar (but not exactly identical) to the 'GNMT' architecture.")
     step_dir_model = os.path.join(dir_task, DIR_PREFIX_MODEL + args.model)
     complete_fname = os.path.join(step_dir_model, FILE_COMPLETE)
     if os.path.exists(complete_fname):
--- a/contrib/autopilot/autopilot.py
+++ b/contrib/autopilot/autopilot.py
@@ -721,6 +721,8 @@ def run_steps(args: argparse.Namespace):
 
     logging.info("=== Train translation model ===")
     logging.info("Model: %s", args.model)
+    if args.model == "gnmt_like":
+        logging.info("WARNING: This is an 8 layer LSTMs model which can resembles the 'GNMT' architecture.")
     step_dir_model = os.path.join(dir_task, DIR_PREFIX_MODEL + args.model)
     complete_fname = os.path.join(step_dir_model, FILE_COMPLETE)
     if os.path.exists(complete_fname):
--- a/contrib/autopilot/autopilot.py
+++ b/contrib/autopilot/autopilot.py
@@ -43,8 +43,9 @@ from contrib.autopilot.tasks import TEXT_UTF8_RAW_BITEXT_REVERSE, TEXT_REQUIRES_
 from contrib.autopilot.tasks import TEXT_UTF8_TOKENIZED
 from contrib.autopilot.tasks import RAW_FILES
 from contrib.autopilot.tasks import Task, TASKS
-from contrib.autopilot.models import MODELS, MODEL_NONE, MODEL_TEST_ARGS
-from contrib.autopilot.models import DECODE_ARGS, DECODE_STANDARD
+from contrib.autopilot.models import MODELS, MODEL_NONE
+from contrib.autopilot.models import MODEL_TEST_ARGS, MODEL_TEST_ARGS_TRANSFORMER, MODEL_TEST_ARGS_GNMT
+from contrib.autopilot.models import DECODE_ARGS, DECODE_STANDARD, DECODE_GNMT
 from contrib.autopilot import third_party
 
 
@@ -863,6 +864,8 @@ def main():
                             help="Pre-defined data set for model training.")
     arg_parser.add_argument("--model", type=str, choices=sorted(MODELS.keys()),
                             help="Type of translation model to train.")
+    arg_parser.add_argument("--model-test-settings", type=str, choices=sorted(MODEL_TEST_ARGS.keys()), default=MODEL_TEST_ARGS_TRANSFORMER,
+                            help="Model test settings. Default: %(default)s.")
     arg_parser.add_argument("--decode-settings", type=str, choices=sorted(DECODE_ARGS.keys()), default=DECODE_STANDARD,
                             help="Decoding settings. Default: %(default)s.")
     arg_parser.add_argument("--custom-task", type=str, metavar="NAME",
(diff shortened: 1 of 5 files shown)


xinyu-intel:
The model in this PR is aligned with #479

contrib/autopilot/models.py
```diff
@@ -14,9 +14,15 @@
 # Names model types
 MODEL_NONE = "none"
 MODEL_TRANSFORMER = "transformer"
+MODEL_GNMT = "gnmt"
+
+# Names model test settings
+MODEL_TEST_ARGS_TRANSFORMER = "ttransformer"
+MODEL_TEST_ARGS_GNMT = "tgnmt"
```
mjdenkowski: We shouldn't need `MODEL_TEST_ARGS_TRANSFORMER ` or `MODEL_TEST_ARGS_GNMT` if we already have `MODEL_TRANSFORMER` and `MODEL_GNMT`.

contrib/autopilot/models.py
```diff
@@ -14,9 +14,15 @@
...
 
 # Named decoding settings
 DECODE_STANDARD = "standard"
+DECODE_GNMT = "dgnmt"
```
mjdenkowski: `dgnmt` -> `gnmt`

contrib/autopilot/models.py
```diff
@@ -59,22 +65,62 @@
...
-    "--transformer-feed-forward-num-hidden=16",
-    "--num-embed=16:16",
-    "--num-words=16:16",
-    "--batch-type=sentence",
-    "--batch-size=1",
-    "--max-updates=4",
-    "--checkpoint-frequency=2",
-]
+MODEL_TEST_ARGS = {
+    MODEL_TEST_ARGS_TRANSFORMER: [
```
mjdenkowski: `MODEL_TEST_ARGS_TRANSFORMER` -> `MODEL_TRANSFORMER`

contrib/autopilot/models.py
```diff
@@ -59,22 +65,62 @@
...
+        "--transformer-feed-forward-num-hidden=16",
+        "--num-embed=16:16",
+        "--num-words=16:16",
+        "--batch-type=sentence",
+        "--batch-size=1",
+        "--max-updates=4",
+        "--checkpoint-frequency=2",
+    ],
+
+    MODEL_TEST_ARGS_GNMT: [
```
mjdenkowski: `MODEL_TEST_ARGS_GNMT` -> `MODEL_GNMT`

contrib/autopilot/autopilot.py
```diff
@@ -43,8 +43,9 @@
 from contrib.autopilot.tasks import TEXT_UTF8_TOKENIZED
 from contrib.autopilot.tasks import RAW_FILES
 from contrib.autopilot.tasks import Task, TASKS
-from contrib.autopilot.models import MODELS, MODEL_NONE, MODEL_TEST_ARGS
-from contrib.autopilot.models import DECODE_ARGS, DECODE_STANDARD
+from contrib.autopilot.models import MODELS, MODEL_NONE
+from contrib.autopilot.models import MODEL_TEST_ARGS, MODEL_TEST_ARGS_TRANSFORMER, MODEL_TEST_ARGS_GNMT
```
mjdenkowski: We shouldn't need `MODEL_TEST_ARGS_TRANSFORMER` or `MODEL_TEST_ARGS_GNMT`.

contrib/autopilot/autopilot.py
```diff
@@ -863,6 +864,8 @@ def main():
                             help="Pre-defined data set for model training.")
     arg_parser.add_argument("--model", type=str, choices=sorted(MODELS.keys()),
                             help="Type of translation model to train.")
+    arg_parser.add_argument("--model-test-settings", type=str, choices=sorted(MODEL_TEST_ARGS.keys()), default=MODEL_TEST_ARGS_TRANSFORMER,
+                            help="Model test settings. Default: %(default)s.")
```
mjdenkowski: We shouldn't need to specify this as we can use the model type as the index into `MODEL_TEST_ARGS`.

fhieber:
I am not sure we should call this model 'gnmt'. I assume you are referring to the models in https://arxiv.org/pdf/1609.08144.pdf. Do they also use an embedding size of 256 for example?
I would prefer renaming this configuration to something like '8-layer rnn' to highlight the different architecture to the transformer. We can add a comment or description that this configuration resembles the 'GNMT' architecture described in https://arxiv.org/pdf/1609.08144.pdf.

xinyu-intel:
@fhieber Thanks for advice. I think this config is not exactly the same as GNMT described in paper but is very similar with it and has very good performance and accuracy. However, GNMT in paper also has more than one combinations. It's main idea is to use LSTMs as encoder and decoder and  I think it is free for users to choose 256 or 512 as embedding size and any other configs like encoder/decoder layer numbers, hidden layer numbers, right?

mjdenkowski:
Do the settings not exactly match the GNMT paper?  I like the idea of having presets that correspond to particular published models, but I also agree that we shouldn't call something "GNMT" when it's not.  Perhaps something like `gnmt_like` to indicate this is as close as we can get in Sockeye and see comparable BLEU scores?

xinyu-intel:
@mjdenkowski Thanks for your advice, I've changed the mdoel name to `gnmt_like` and add a comment when using this config `This is an 8 layer LSTMs model which can resembles the 'GNMT' architecture.` And all tests have passed:)

TaoLv:
I think implementing more famous published models can really help improve the community of framework and toolkits like sockeye. GNMT is one of this kind of famous models. So we are trying to implement and train it on the top of mxnet and sockeye. I know GluonNLP is also trying to do something similar.
Here in this PR, we want to present a GNMT model as close as possible to the original paper, and also want to make it comparable to the TF implementation for both accuracy and performance. As @xinyu-intel mentioned, we have got 26.91 for BLEU on both CPU and GPU. If you are interested in the performance of this model, I think @xinyu-intel can also share some initial data for it. Anyway, I think it would be great if we can figure out that mxnet and sockeye have the same GNMT model which has better performance compared with other frameworks, right?
So if you find anything in this PR does not align with the original GNMT paper or TF implementation, please kindly point it out. Or if you have a more general implementation for GNMT, please kindly share it with us if possible. We are pleased and looking forward to working with the sockeye community to make this happen.

fhieber:
Totally agree with you @TaoLv and I think its great to have a well-working RNN-type model readily available in Sockeye for others to use! My point was to ask how sure you (or others) are that this configuration is as close as possible to the GNMT model (given that there was a discussion around the correct parameters in #479). I haven't looked into replicating GNMT models in Sockeye, but did for the transformer and there, many details (down to embedding initialization, scaling etc.) are crucial to reproduce published results.
If the config in this PR does equally or better than published results in the GNMT paper, this is awesome! But we should try to be extremely clear on what number in the paper we are comparing against, how we compute it, how - to the best of our knowledge - the setup differs from the published paper, and what language pair/dataset we are reporting this on.
Again, more presets/configs for Sockeye autopilot are super welcome and helpful, but in my experience replicating published models across different toolkits and frameworks in a fair way can be quite complex, and before we make any claim that GNMT models in MXNet/Sockeye do better than previously published results, we need to be absolutely sure about it.

That said, calling the current config `gnmt_like` is absolutely fine with me and you can always iterate on it to make it more precise (maybe to eventually call it `gnmt`).

TaoLv:
You're right @fhieber. It's not an easy task to implement a publised model and reproduce the accuracy of it. That's why we want to provide model zoo, pre-trained models and fixed configurations to end users. I'm quite OK if you want to merge this PR as a `gnmt_like` configuration for now. And I still think it's a good start point for us to push this effort forward to create a more general, comparable and efficient GNMT model based mxnet and sockeye.

contrib/autopilot/autopilot.py
```diff
@@ -721,6 +721,8 @@ def report_data(file_sets):
 
     logging.info("=== Train translation model ===")
     logging.info("Model: %s", args.model)
+    if args.model == "gnmt_like":
```
mjdenkowski: `"gnmt_like"` -> `MODEL_GNMT`

contrib/autopilot/autopilot.py
```diff
@@ -721,6 +721,8 @@ def report_data(file_sets):
...
+        logging.info("WARNING: This is an 8 layer LSTMs model which can resembles the 'GNMT' architecture.")
```
mjdenkowski: This could be worded: `"NOTE: This is an 8 layer LSTM model similar (but not exactly identical) to the 'GNMT' architecture."`

xinyu-intel: @mjdenkowski Thanks, changed:)

mjdenkowski:
Can you update the minor version and changelog?

xinyu-intel:
@mjdenkowski added

TaoLv:
Glad to see this PR was merged. Now, sockeye autopilot supports both `transformer` and `gnmt_like`. So do you think we need to adjust the README of autopilot to promote that and tell users how to run `gnmt_like` and reproduce the accuracy. @xinyu-intel @fhieber

PR #500, opened 2018-08-03
Removed iterative sorting of `sequences` array from beam search in favor of backpointers

fhieber:
Removed iterative sorting of `sequences` array from beam search in favor of backpointers after beam search in order to extract the kth-best hypothesis.

This way we get rid of take & concat ops during beam search. Some speed testing from my side on a p3 yielded a very small speedup at slightly less memory usage.
```
batch size  1: +0.03 sent/sec
batch_size  8: +0.20 sent/sec
batch_size 16: +0.33 sent/sec
```

until you can check this box.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1360,6 +1360,7 @@ class Translator:
                                                                  mx.nd.NDArray,
                                                                  mx.nd.NDArray,
                                                                  mx.nd.NDArray,
+                                                                 mx.nd.NDArray,
                                                                  List[Optional[constrained.ConstrainedHypothesis]],
                                                                  Optional[List[BeamHistory]]]:
         """
@@ -1371,8 +1372,9 @@ class Translator:
                that must appear in each output.
         :param raw_avoid_list: A list of optional lists containing phrases (as lists of target word IDs)
                that must NOT appear in each output.
-        :return List of lists of word ids, list of attentions, array of accumulated length-normalized
-                negative log-probs.
+        :return List of best hypotheses indices, list of best word indices, list of attentions,
+                array of accumulated length-normalized negative log-probs, hypotheses lengths, constraints (if any),
+                beam histories (if any).
         """
 
         # Length of encoded sequence (may differ from initial input length)
@@ -1388,9 +1390,10 @@ class Translator:
 
         best_word_indices = mx.nd.full((self.batch_size * self.beam_size,), val=self.start_id, ctx=self.context,
                                        dtype='int32')
-        # The growing sequences representing the complete history for each hypothesis: (batch_size * beam_size, 1)
-        # The first column represents <s>, and is later removed when _beam_search() returns.
-        sequences = mx.nd.zeros((self.batch_size * self.beam_size, 1), ctx=self.context, dtype='int32')
+
+        # Best word and hypotheses indices across beam search steps from topk operation.
+        best_hyp_indices_list = []  # type: List[mx.nd.NDArray]
+        best_word_indices_list = []  # type: List[mx.nd.NDArray]
 
         # Beam history
         beam_histories = None  # type: Optional[List[BeamHistory]]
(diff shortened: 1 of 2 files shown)


sockeye/inference.py
```diff
@@ -1357,6 +1357,7 @@ def _beam_search(self,
                      raw_constraint_list: List[Optional[constrained.RawConstraintList]],
                      raw_avoid_list: List[Optional[constrained.RawConstraintList]],
                      max_output_lengths: mx.nd.NDArray) -> Tuple[mx.nd.NDArray,
+                                                                 mx.nd.NDArray,
```
tdomhan: you're returning numpy arrays now, righy? How about moving the numpy conversion to where you actually need numpy arrays?

test/unit/test_inference.py
```diff
@@ -399,3 +399,15 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
     assert all(mx_hyp == np_hyp)
     assert all(mx_word == np_word)
     assert all(mx_values == np_values)
+
+
+def test_get_best_word_indeces_for_kth_hypotheses():
+    np.random.seed(13)
```
tdomhan: how about running the test on the minimal example you had in order to test for the actual outputs?

sockeye/inference.py
fhieber: yes, I forgot to change the annotation for the return values. I don't have an opinion on whether to convert to numpy. Performance-wise it shouldn't matter at all. I can move the conversion to _get_best_from_beam() method if you like.

test/unit/test_inference.py
fhieber: yeah, probably makes sense. This way we can at least assert on the correctness of the outputs. The way the test is right now its not really helpful :)

sockeye/inference.py
fhieber: if I 'delay' the numpy conversion `_get_best_from_beam()` starts to look like this:
```python
    def _get_best_from_beam(self,
                            best_hyp_indices: mx.nd.NDArray,
                            best_word_indices: mx.nd.NDArray,
                            attentions: mx.nd.NDArray,
                            seq_scores: mx.nd.NDArray,
                            lengths: mx.nd.NDArray,
                            constraints: List[Optional[constrained.ConstrainedHypothesis]],
                            beam_histories: Optional[List[BeamHistory]] = None) -> List[Translation]:
        """
        Return the best (aka top) entry from the n-best list.

        :param best_hyp_indices: Array of best hypotheses indices ids. Shape: (batch * beam, num_beam_search_steps + 1).
        :param best_word_indices: Array of best hypotheses indices ids. Shape: (batch * beam, num_beam_search_steps).
        :param attentions: Array of attentions over source words.
                           Shape: (batch * beam, num_beam_search_steps, encoded_source_length).
        :param seq_scores: Array of length-normalized negative log-probs. Shape: (batch * beam, 1)
        :param lengths: The lengths of all items in the beam. Shape: (batch * beam). Dtype: int32.
        :param constraints: The constraints for all items in the beam. Shape: (batch * beam).
        :param beam_histories: The beam histories for each sentence in the batch.
        :return: List of Translation objects containing all relevant information.
        """
        # Initialize the best_ids to the first item in each batch
        best_ids = np.arange(0, self.batch_size * self.beam_size, self.beam_size, dtype='int32')
        seq_scores = seq_scores.asnumpy()
        if any(constraints):
            # For constrained decoding, select from items that have met all constraints (might not be finished)
            unmet = np.array([c.num_needed() if c is not None else 0 for c in constraints])
            filtered = np.where(unmet == 0, seq_scores.flatten(), np.inf)
            filtered = filtered.reshape((self.batch_size, self.beam_size))
            best_ids += np.argmin(filtered, axis=1).astype('int32')

        # Obtain translation data for all best hypotheses in the batch
        indices = self._get_best_word_indeces_for_kth_hypotheses(best_ids, best_hyp_indices.asnumpy())
        sequences = best_word_indices.asnumpy()[indices, np.arange(indices.shape[1])]
        lengths = lengths.asnumpy().astype('int32')[best_ids]
        attentions = attentions.asnumpy()[best_ids]
        seq_scores = seq_scores[best_ids]
        histories = beam_histories if beam_histories is not None else [None] * self.batch_size  # type: List

        return [self._assemble_translation(*x) for x in zip(sequences, lengths, attentions, seq_scores, histories)]
```
which I honestly find less pretty. Especially because we need to convert `seq_scores` for the constraints block and everything else only later.
I'd prefer the way it currently is in this PR (subject to updating the typing annotation for the return values). What do you think?

tdomhan: sure, I think either way is fine. This is all internal to the Translator, so it comes down to what we think is cleaner. If there is not other place where we need the mxnet ndarrays I think this is fine.

tdomhan: we should still change the annotation though.

fhieber:
this should be ready for merge. 
Additional speed testing on a p2 showed the same consistent but small speedup as reported above. And, this now one data structure less that is changing shape across beam search iterations :)

PR #508, opened 2018-08-17
No longer relying on external translation id.

tdomhan:
We used to rely on the external id for sorting the translations correctly, which is not necessary and can cause some issues. With this change we still carry through the externally provided id to the user, but no longer rely on it internally.

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -1184,12 +1184,13 @@ class Translator:
         # Concatenate results
         results = []  # type: List[TranslatorOutput]
         chunks_by_input_idx = itertools.groupby(translated_chunks, key=lambda translation: translation.input_idx)
-        for trans_input, (input_idx, chunks) in zip(trans_inputs, chunks_by_input_idx):
-            chunks = list(chunks)  # type: ignore
-            if len(chunks) == 1:  # type: ignore
-                translation = chunks[0].translation  # type: ignore
+        for trans_input, (input_idx, translations_for_input_idx) in zip(trans_inputs, chunks_by_input_idx):
+            translations_for_input_idx = list(translations_for_input_idx)  # type: ignore
+            if len(translations_for_input_idx) == 1:  # type: ignore
+                translation = translations_for_input_idx[0].translation  # type: ignore
             else:
-                translations_to_concat = [translated_chunk.translation for translated_chunk in chunks]
+                translations_to_concat = [translated_chunk.translation
+                                          for translated_chunk in translations_for_input_idx]
                 translation = self._concat_translations(translations_to_concat)
 
             results.append(self._make_result(trans_input, translation))
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -564,13 +564,14 @@ def get_max_input_output_length(supported_max_seq_len_source: Optional[int],
 
 BeamHistory = Dict[str, List]
 Tokens = List[str]
+SentenceId = Union[int, str]
 
 
 class TranslatorInput:
     """
     Object required by Translator.translate().
 
-    :param sentence_id: Sentence id.
+    :param sentence_id: SentenceId.
     :param tokens: List of input tokens.
     :param factors: Optional list of additional factor sequences.
     :param constraints: Optional list of target-side constraints.
@@ -579,7 +580,7 @@ class TranslatorInput:
     __slots__ = ('sentence_id', 'tokens', 'factors', 'constraints', 'avoid_list')
 
     def __init__(self,
-                 sentence_id: int,
+                 sentence_id: SentenceId,
                  tokens: Tokens,
                  factors: Optional[List[Tokens]] = None,
                  constraints: Optional[List[Tokens]] = None,
@@ -591,7 +592,7 @@ class TranslatorInput:
         self.avoid_list = avoid_list
 
     def __str__(self):
-        return 'TranslatorInput(%d, %s, factors=%s, constraints=%s, avoid=%s)' \
+        return 'TranslatorInput(%s, %s, factors=%s, constraints=%s, avoid=%s)' \
             % (self.sentence_id, self.tokens, self.factors, self.constraints, self.avoid_list)
 
     def __len__(self):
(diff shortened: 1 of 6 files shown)


tdomhan:
Missing:
* tests
* mypy

sockeye/inference.py
```diff
@@ -577,25 +577,23 @@ class TranslatorInput:
     :param chunk_id: Chunk id. Defaults to -1.
```
fhieber: can be removed

fhieber:
I am running into an issue with this fix when using `--batch-size > 1` and `--max-seq-len`:
```
> head -50 input | python -m sockeye.translate -m model --use-cpu --batch-size 2 --max-input-len 10

[INFO:sockeye.inference] Translator (1 model(s) beam_size=5 beam_prune=off beam_search_stop=all ensemble_mode=None batch_size=2 buckets_source=[10] avoiding=0)
[INFO:__main__] Translating...
[ERROR:__main__] Uncaught exception
Traceback (most recent call last):
  File "/Users/fhieber/miniconda3/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/fhieber/miniconda3/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/translate.py", line 221, in <module>
    main()
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/translate.py", line 40, in main
    run_translate(args)
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/translate.py", line 111, in run_translate
    input_is_json=args.json_input)
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/translate.py", line 188, in read_and_translate
    chunk_time = translate(output_handler, chunk, translator)
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/translate.py", line 212, in translate
    trans_outputs = translator.translate(trans_inputs)
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/inference.py", line 1174, in translate
    batch_translations = self._translate_nd(*self._get_inference_input(translator_inputs))
  File "/Users/fhieber/workspace/nmt/src/sockeye-github/sockeye/inference.py", line 1221, in _get_inference_input
    source = mx.nd.zeros((len(trans_inputs), bucket_key, self.num_source_factors), ctx=self.context)
  File "/Users/fhieber/miniconda3/lib/python3.6/site-packages/mxnet/ndarray/utils.py", line 67, in zeros
    return _zeros_ndarray(shape, ctx, dtype, **kwargs)
  File "/Users/fhieber/miniconda3/lib/python3.6/site-packages/mxnet/ndarray/ndarray.py", line 3750, in zeros
    return _internal._zeros(shape=shape, ctx=ctx, dtype=dtype, **kwargs)
  File "<string>", line 34, in _zeros
  File "/Users/fhieber/miniconda3/lib/python3.6/site-packages/mxnet/_ctypes/ndarray.py", line 92, in _imperative_invoke
    ctypes.byref(out_stypes)))
  File "/Users/fhieber/miniconda3/lib/python3.6/site-packages/mxnet/base.py", line 255, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))
mxnet.base.MXNetError: Invalid Parameter format for shape expect Shape(tuple) but value='(2, None, 1)', in operator _zeros(name="", dtype="float32", ctx="cpu(0)", shape="(2, None, 1)")

Seems like bucket key is `None` causing `zeros` to fail.
`bucket_key` is assigned with this statement:
```
```

After some quick debugging it turns out that the first `trans_input` has 11 input tokens (the others 10). With `--max-input-len 10` this yields `None` as `bucket_key`. 

This doesn't happen on current master.

**Comment by fhieber** [comment]
I may have found the problem: my toy model does not use `source_with_eos` yet so it goes into the `else` block when constructing TranslatorInputs.
```
[...]
                else:
                    if len(trans_input.tokens) > self.max_input_length:
                        # oversized input
                        logger.debug(
                            "Input %d has length (%d) that exceeds max input length (%d). "
                            "Splitting into chunks of size %d.",
                            trans_input.sentence_id, len(trans_input.tokens),
                            self.buckets_source[-1], self.max_input_length)
                        chunks = [trans_input_chunk
                                  for trans_input_chunk in
                                  trans_input.chunks(self.max_input_length)]
                        input_chunks.extend([IndexedTranslatorInput(trans_input_idx, chunk_idx, chunk_input)
                                             for chunk_idx, chunk_input in enumerate(chunks)])
                    else:
                        # regular input
                        input_chunks.append(IndexedTranslatorInput(trans_input_idx,
                                                                   chunk_idx=0,
                                                                   translator_input=trans_input.with_eos()))
```

When removing `.with_eos()` from the last line it works. Seems like a copy and paste error.

I will push this update to the branch.

File: sockeye/inference.py [hunk 2]
```diff
@@ -1115,26 +1126,33 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
+                                  for trans_input_chunk in
+                                  trans_input.chunks(self.max_input_length)]
+                        input_chunks.extend([IndexedTranslatorInput(trans_input_idx, chunk_idx, chunk_input)
+                                             for chunk_idx, chunk_input in enumerate(chunks)])
                     else:
-                        input_chunks.append(trans_input)
+                        # regular input
+                        input_chunks.append(IndexedTranslatorInput(trans_input_idx,
+                                                                   chunk_idx=0,
+                                                                   translator_input=trans_input.with_eos()))
```
fhieber: this shouldn't use `.with_eos()`

sockeye/inference.py
```diff
@@ -564,38 +564,36 @@ def get_max_output_length(input_length: int):
...
 Tokens = List[str]
+SentenceId = Union[int, str]
 
 
 class TranslatorInput:
     """
     Object required by Translator.translate().
 
-    :param sentence_id: Sentence id.
+    :param sentence_id: SentenceId.
```
davvil: I actually prefer the version with space.

tdomhan:
ohh, good catch! was indeed a copy/paste error!

PR #512, opened 2018-08-23
Remove dependence on nvidia-smi.

tdomhan:
Removes the dependence on nvidia-smi, by trying to allocate on each device id, as suggested in https://github.com/apache/incubator-mxnet/issues/2910. 

Note: With MXNet 1.3 we can simplify this to `mx.context.num_gpus()`.

### DIFF
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -433,19 +433,22 @@ def average_arrays(arrays: List[mx.nd.NDArray]) -> mx.nd.NDArray:
     return new_array
 
 
+MAX_NUM_GPU_TO_TRY = 100
+
+
 def get_num_gpus() -> int:
     """
-    Gets the number of GPUs available on the host (depends on nvidia-smi).
+    Gets the number of GPUs available on the host.
 
     :return: The number of GPUs on the system.
     """
-    if shutil.which("nvidia-smi") is None:
-        logger.warning("Couldn't find nvidia-smi, therefore we assume no GPUs are available.")
-        return 0
-    sp = subprocess.Popen(['nvidia-smi', '-L'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    out_str = sp.communicate()[0].decode("utf-8")
-    num_gpus = len(out_str.rstrip("\n").split("\n"))
-    return num_gpus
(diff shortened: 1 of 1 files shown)


sockeye/utils.py
```diff
@@ -433,19 +433,22 @@ def average_arrays(arrays: List[mx.nd.NDArray]) -> mx.nd.NDArray:
...
-    out_str = sp.communicate()[0].decode("utf-8")
-    num_gpus = len(out_str.rstrip("\n").split("\n"))
-    return num_gpus
+    # TODO (domhant): Switch to mx.context.num_gpus() with mxnet version 1.3
+    for device_id in range(MAX_NUM_GPU_TO_TRY):
+        try:
+            mx.nd.zeros((1,), ctx=mx.gpu(device_id))
+        except:
+            return device_id
+    return MAX_NUM_GPU_TO_TRY
```
davvil: Not that it really matters in practice, but as the number of GPUs is finite the allocation will fail at some point. It would be more elegant having an unlimited while loop couning GPUs (and then we support machines with >100 GPUs).

sockeye/utils.py
```diff
@@ -433,19 +433,22 @@ def average_arrays(arrays: List[mx.nd.NDArray]) -> mx.nd.NDArray:
...
-        return 0
-    sp = subprocess.Popen(['nvidia-smi', '-L'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    out_str = sp.communicate()[0].decode("utf-8")
-    num_gpus = len(out_str.rstrip("\n").split("\n"))
-    return num_gpus
+    # TODO (domhant): Switch to mx.context.num_gpus() with mxnet version 1.3
+    for device_id in range(MAX_NUM_GPU_TO_TRY):
+        try:
+            mx.nd.zeros((1,), ctx=mx.gpu(device_id))
+        except:
```
fhieber: wouldn't mypy/pylint complain again about this bare except statement?
Any chance we can use a more specific catch?

sockeye/utils.py
tdomhan: sounds reasonable. will do

PR #516, opened 2018-08-29
[WIP] Dynamic batch_size during inference

mr-lee:
Added option for --use-dynamic-batch-size, to allow Sockeye to support dynamic batch sizes during inference (on STDIN with json input only).

### Notes
This change makes the the "batch_size" argument dynamic, such that a user could specify a different batch_size per invocation. The default behavior of the pipeline is unchanged if you do not specify this option, and I have verified below that this does not have an impact on latency. I have also verified that there is no diff in the translation results whatsoever in the standard use-case.

This of course, may not be the case if we are to batch sentences together. If needed, we can do more extensive testing on that.

### Testing
Added new unit test cases, and everything passes.

For the dataset below:
```
[samples  202] [mean 18.30 tok] [+/- 12.61 tok] [max 123.00] [min 5.00]
```

#### Before Change
```
[samples  202] [mean 88.446089 tok/s] [+/- 22.044562 tok/s] [max 164.731684] [min 25.449224]
```

#### After Change
```
[samples  202] [mean 88.533140 tok/s] [+/- 22.097902 tok/s] [max 167.662940] [min 25.452589]
```

Note that this confirms that there should be no impact to latency.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -1139,6 +1139,14 @@ def add_inference_args(params):
                                type=int,
                                default=None,
                                help='Maximum input sequence length. Default: value from model(s).')
+    decode_params.add_argument('--use-dynamic-batch-size',
+                               action='store_true',
+                               default=False,
+                               help='Turning on this option allows inference to be done with dynamic batch-sizes, up to the'
+                                    'specified value in the --batch-size argument.  NOTE: This option can only be used when'
+                                    'reading from STDIN for translation using the built-in interface - otherwise the option'
+                                    'will simply be disregarded. Default: False')
+
     decode_params.add_argument('--softmax-temperature',
                                type=float,
                                default=None,
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -21,7 +21,7 @@ import os
 import time
 from collections import defaultdict
 from functools import lru_cache, partial
-from typing import Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union, Set
+from typing import Any, Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union, Set
 
 import mxnet as mx
 import numpy as np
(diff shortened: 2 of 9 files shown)


sockeye/translate.py
```diff
@@ -165,29 +177,47 @@ def read_and_translate(translator: inference.Translator,
...
-    for chunk in grouper(make_inputs(input_file, translator, input_is_json, input_factors), size=chunk_size):
-        chunk_time = translate(output_handler, chunk, translator)
-        total_lines += len(chunk)
-        total_time += chunk_time
+    # We allow for dynamic batch calls if calling from STDIN with json inputs
+    if dynamic_batch_mode_enabled and input_file is None and input_is_json:
+        logger.info("Dynamic batch mode enabled, translating in batches as delivered...")
+        for translation_in in make_input_lists():
+            # if the input goes beyond the max batch_size, split into batch_size chunks
+            max_batches = [translation_in[i:i + batch_size] for i in xrange(0, len(translation_in), batch_size)]
```
tdomhan: Should be `range.`

sockeye/arguments.py
```diff
@@ -1139,6 +1139,14 @@ def add_inference_args(params):
                                type=int,
                                default=None,
                                help='Maximum input sequence length. Default: value from model(s).')
+    decode_params.add_argument('--use-dynamic-batch-size',
```
tdomhan: How about instead of having a separate parameter enabling a dynamic batch size by default. Namely, so that `Translator.translate` would internally use a reduced batch_size for each `List[TranslatorInput]` that is smaller than the `batch_size`.  Then the user wouldn't need to specify any option and the `Translator` object would automatically do what's most efficient.

For input coming from stdin it is less clear why one would want dynamic batching, as in the current stage it is mainly aimed at batch translation where large chunks are read, sorted and translated. Then again, if we changed the Translator as described above, the last batch in a chunk (which might be smaller than `batch_size` could make use of dynamic batching. If we consider the realm time translation use case from stdin, it might make sense to define a json list of inputs interface, for which then in turn could use dyanmic batching.

sockeye/inference.py
```diff
@@ -1112,6 +1146,33 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
...
+        """
+        Batch-translates a list of TranslatorInputs, returns a list of TranslatorOutputs.
+        Allows the specification of dynamic batch size during inference, up to a maximum
+        of the originally specified batch size.
+        Splits oversized sentences to sentence chunks of size less than max_input_length.
+
+        :param trans_inputs: List of TranslatorInputs as returned by make_input().
+        :return: List of translation results.
+        """
+        return self._translate(trans_inputs, len(trans_inputs))
```
tdomhan: should we check that the batch_size is not larger than the default batch size?

fhieber:
I am closing this in favor of #641. Thanks for getting this started!

PR #519, opened 2018-08-31
Skip softmax and topk

bricksdont:
Skip softmax by default for greedy decoding, and adding option to skip topk:
- this change only affects inference
- only affects greedy decoding

Skipping softmax is significantly faster, skipping topk is significantly slower, but it could be faster if the MXnet implementation of argmin changes.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -55,6 +55,8 @@ class InferenceModel(model.SockeyeModel):
     :param decoder_return_logit_inputs: Decoder returns inputs to logit computation instead of softmax over target
                                         vocabulary.  Used when logits/softmax are handled separately.
     :param cache_output_layer_w_b: Cache weights and biases for logit computation.
+    :param skip_softmax: If True, does not compute softmax for greedy decoding.
+    :param skip_topk: If True, uses argmax instead of topk for greedy decoding.
     """
 
     def __init__(self,
@@ -67,7 +69,9 @@ class InferenceModel(model.SockeyeModel):
                  max_output_length_num_stds: int = C.DEFAULT_NUM_STD_MAX_OUTPUT_LENGTH,
                  decoder_return_logit_inputs: bool = False,
                  cache_output_layer_w_b: bool = False,
-                 forced_max_output_len: Optional[int] = None) -> None:
+                 forced_max_output_len: Optional[int] = None,
+                 skip_softmax: bool = False,
+                 skip_topk: bool = False) -> None:
         super().__init__(config)
         self.params_fname = params_fname
         self.context = context
@@ -79,6 +83,8 @@ class InferenceModel(model.SockeyeModel):
         self.max_input_length, self.get_max_output_length = models_max_input_output_length([self],
                                                                                            max_output_length_num_stds,
                                                                                            forced_max_output_len=forced_max_output_len)
+        self.skip_softmax = skip_softmax
+        self.skip_topk = skip_topk
 
         self.encoder_module = None  # type: Optional[mx.mod.BucketingModule]
         self.encoder_default_bucket_key = None  # type: Optional[int]
@@ -236,7 +242,11 @@ class InferenceModel(model.SockeyeModel):
                 logits = self.output_layer(target_decoded)
                 if self.softmax_temperature is not None:
                     logits = logits / self.softmax_temperature
-                outputs = mx.sym.softmax(data=logits, name=C.SOFTMAX_NAME)
+                if self.beam_size == 1 and self.skip_softmax:
+                    # skip softmax for greedy decoding
+                    outputs = logits
+                else:
+                    outputs = mx.sym.softmax(data=logits, name=C.SOFTMAX_NAME)
 
             data_names = [C.TARGET_NAME] + state_names
             label_names = []  # type: List[str]
@@ -362,9 +372,11 @@ def load_models(context: mx.context.Context,
                 decoder_return_logit_inputs: bool = False,
                 cache_output_layer_w_b: bool = False,
                 forced_max_output_len: Optional[int] = None,
-                override_dtype: Optional[str] = None) -> Tuple[List[InferenceModel],
-                                                               List[vocab.Vocab],
-                                                               vocab.Vocab]:
+                override_dtype: Optional[str] = None,
+                skip_softmax: bool = False,
+                skip_topk: bool = False) -> Tuple[List[InferenceModel],
+                                                  List[vocab.Vocab],
+                                                  vocab.Vocab]:
     """
     Loads a list of models for inference.
 
@@ -383,6 +395,8 @@ def load_models(context: mx.context.Context,
                                    restrict lexicon).
     :param forced_max_output_len: An optional overwrite of the maximum output length.
     :param override_dtype: Overrides dtype of encoder and decoder defined at training time to a different one.
+    :param skip_softmax: If True, does not compute softmax for greedy decoding.
+    :param skip_topk: If True, uses argmax instead of topk for greedy decoding.
     :return: List of models, source vocabulary, target vocabulary, source factor vocabularies.
     """
     logger.info("Loading %d model(s) from %s ...", len(model_folders), model_folders)
@@ -420,7 +434,9 @@ def load_models(context: mx.context.Context,
                                          batch_size=batch_size,
                                          softmax_temperature=softmax_temperature,
                                          decoder_return_logit_inputs=decoder_return_logit_inputs,
-                                         cache_output_layer_w_b=cache_output_layer_w_b)
+                                         cache_output_layer_w_b=cache_output_layer_w_b,
+                                         skip_softmax=skip_softmax,
+                                         skip_topk=skip_topk)
         utils.check_condition(inference_model.num_source_factors == len(model_source_vocabs),
                               "Number of loaded source vocabularies (%d) does not match "
                               "number of source factors for model '%s' (%d)" % (len(model_source_vocabs), model_folder,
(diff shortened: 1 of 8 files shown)


sockeye/utils.py
```diff
@@ -251,6 +251,28 @@ def variance(self) -> float:
             return self._M2 / self._count
 
 
+def top1(scores: mx.nd.NDArray,
+         offset: mx.nd.NDArray) -> Tuple[mx.nd.NDArray, mx.nd.NDArray, mx.nd.NDArray]:
+    """
+    Get the single lowest element per sentence from a `scores` matrix. Expects that
+    beam size is 1, for greedy decoding.
+
+    NOTE(mathmu): The current implementation of argmin in MXNet much slower than topk with k=1.
```
fhieber: In that case I think we should use topk then and add a TODO to change to argmin once its sped up.

test/integration/test_seq_copy_int.py
```diff
@@ -33,6 +33,22 @@
      " --decode-and-evaluate 0",
      "--beam-size 2 --softmax-temperature 0.01",
      True, False, False),
+    # "Vanilla" LSTM encoder-decoder with attention, greedy and skip softmax
+    ("--encoder rnn --decoder rnn --num-layers 1 --rnn-cell-type lstm --rnn-num-hidden 8 --num-embed 4 "
+     " --rnn-attention-type mlp"
+     " --rnn-attention-num-hidden 8 --batch-size 2 --loss cross-entropy --optimized-metric perplexity --max-updates 2"
+     " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --batch-type sentence "
+     " --decode-and-evaluate 0",
+     "--beam-size 2 --softmax-temperature 0.01",
```
fhieber: I don't see the `--skip-softmax` here; also, using --softmax-temperature in this test is quite confusing then :)

sockeye/inference.py
```diff
@@ -1377,10 +1412,13 @@ def _combine_predictions(self,
 
         # combine model predictions and convert to neg log probs
         if len(self.models) == 1:
-            neg_logprobs = -mx.nd.log(probs[0])  # pylint: disable=invalid-unary-operand-type
+            if self.beam_size == 1 and self.skip_softmax:
+                neg_probs = -probs[0]
+            else:
+                neg_probs = -mx.nd.log(probs[0])  # pylint: disable=invalid-unary-operand-type
```
fhieber: these are still negative log probs, no?

test/integration/test_seq_copy_int.py
bricksdont: `--skip-softmax` is not an option anymore, but the default behaviour. Sorry about the confusion; it is something I changed with the third commit.

sockeye/inference.py
bricksdont: If I understand correctly, if log is not applied they will not be in log space.

sockeye/utils.py
bricksdont: The current default behaviour is to use topk for beam size 1, using top1 is a CLI option: `--skip-topk`.

sockeye/inference.py
fhieber: sure, so neither name can be entirely correct :)

sockeye/inference.py
```diff
@@ -79,6 +83,8 @@ def __init__(self,
         self.max_input_length, self.get_max_output_length = models_max_input_output_length([self],
                                                                                            max_output_length_num_stds,
                                                                                            forced_max_output_len=forced_max_output_len)
+        self.skip_softmax = skip_softmax
```
tdomhan: you should probably add a `check_condition` on the `beam_size`

sockeye/inference.py
```diff
@@ -394,6 +406,13 @@ def load_models(context: mx.context.Context,
     if checkpoints is None:
         checkpoints = [None] * len(model_folders)
 
+    # skip softmax for a single model,
+    if len(model_folders) == 1:
+        skip_softmax = True
```
tdomhan: probably makes sense to only do this for `beam_size == 1` explicitly

sockeye/inference.py
bricksdont: Oh I see - yes you are right. Want me to change it?

sockeye/inference.py
```diff
@@ -67,7 +69,9 @@ def __init__(self,
                  max_output_length_num_stds: int = C.DEFAULT_NUM_STD_MAX_OUTPUT_LENGTH,
                  decoder_return_logit_inputs: bool = False,
                  cache_output_layer_w_b: bool = False,
-                 forced_max_output_len: Optional[int] = None) -> None:
+                 forced_max_output_len: Optional[int] = None,
+                 skip_softmax: bool = False,
+                 skip_topk: bool = False) -> None:
```
tdomhan: `skip_topk` is only used in Translator but not in InfereceModel. Therefore it should not be a member of InferenceModel, but rather be directly passed to Translator.

test/integration/test_seq_copy_int.py
tdomhan: you should update the comment then. Also: It's `--beam-size 2` and therefore not greedy ;)

sockeye/inference.py
```diff
@@ -67,18 +68,22 @@ def __init__(self,
...
+                 forced_max_output_len: Optional[int] = None,
+                 skip_softmax: bool = False) -> None:
         super().__init__(config)
         self.params_fname = params_fname
         self.context = context
         self.beam_size = beam_size
         utils.check_condition(beam_size < self.config.vocab_target_size,
                               'The beam size must be smaller than the target vocabulary size.')
+        if skip_softmax:
+            utils.check_condition(beam_size == 1, 'Skipping softmax does not have any effect for beam size > 1')
```
tdomhan: as this will raise a RuntimeError, maybe: "Softmax skipping can only be enabled with beam size of 1."

sockeye/inference.py
```diff
@@ -394,6 +403,13 @@ def load_models(context: mx.context.Context,
     if checkpoints is None:
         checkpoints = [None] * len(model_folders)
 
+    # skip softmax for a single model,
+    if len(model_folders) == 1 and beam_size == 1:
+        skip_softmax = True
+    else:
+        # but not for an ensemble
```
tdomhan: you should probably warn the user with a logging message in this case.

sockeye/inference.py
tdomhan: also, actually now skip_softmax is no longer a user facing parameter. So I in that case I'd say make it an assertion.

sockeye/inference.py
```diff
@@ -981,7 +1001,8 @@ def __init__(self,
                  restrict_lexicon: Optional[lexicon.TopKLexicon] = None,
                  avoid_list: Optional[str] = None,
                  store_beam: bool = False,
-                 strip_unknown_words: bool = False) -> None:
+                 strip_unknown_words: bool = False,
+                 skip_topk: bool = False,) -> None:
```
tdomhan: you probably want to remove that comma

sockeye/inference.py
```diff
@@ -1005,6 +1026,9 @@ def __init__(self,
         self.interpolation_func = self._get_interpolation_func(ensemble_mode)
         self.beam_size = self.models[0].beam_size
         self.batch_size = self.models[0].batch_size
+        # skip softmax for a single model, but not for an ensemble
+        self.skip_softmax = True if len(models) == 1 else False
```
tdomhan: let's decide on `skip_softmax` in just a single location, probably in `load_models` and reuse it after.

sockeye/inference.py
```diff
@@ -1352,9 +1380,14 @@ def _decode_step(self,
             # Compute logits and softmax with restricted vocabulary
             if self.restrict_lexicon:
                 logits = model.output_layer(decoder_outputs, out_w, out_b)
-                probs = mx.nd.softmax(logits)
+                if self.beam_size == 1 and self.skip_softmax:
```
tdomhan: wouldn't it be enough to use `skip_softmax`, as it should only be true with beam_size 1?

sockeye/inference.py
```diff
@@ -1005,6 +1028,9 @@ def __init__(self,
         self.interpolation_func = self._get_interpolation_func(ensemble_mode)
         self.beam_size = self.models[0].beam_size
         self.batch_size = self.models[0].batch_size
+        # skip softmax for a single model, but not for an ensemble
+        self.skip_softmax = self.models[0].skip_softmax
+        self.skip_topk = skip_topk
```
tdomhan: maybe a final check_condition to cover the case of someone creating a `Translator` without `load_models`.

```
if skip_softmax:
  check_condition(len(self.models) == 0 and self.beam_size ==1, "...")
```

sockeye/inference.py
```diff
@@ -1377,10 +1412,13 @@ def _combine_predictions(self,
 
         # combine model predictions and convert to neg log probs
         if len(self.models) == 1:
-            neg_logprobs = -mx.nd.log(probs[0])  # pylint: disable=invalid-unary-operand-type
+            if self.beam_size == 1 and self.skip_softmax:
```
tdomhan: is this check necessary?

sockeye/inference.py
```diff
@@ -394,6 +403,15 @@ def load_models(context: mx.context.Context,
...
         checkpoints = [None] * len(model_folders)
 
+    # skip softmax for a single model,
+    if len(model_folders) == 1 and beam_size == 1:
+        skip_softmax = True
+        logger.info("Enabled skipping softmax for a single model and greedy decoding.")
+    else:
+        # but not for an ensemble or beam search
+        skip_softmax = False
+        logger.info("Disabled skipping softmax for several models or beam size larger than 1.")
```
fhieber: I think this warning can be confusing if a user does NOT set skip_softmax and runs with the default settings (beam_size=5). Maybe only print this warning when skip_softmax==True? Or remove it entirely

sockeye/utils.py
fhieber: but aren't you saying that this is slower than using mx.nd.topk? What's the point of this option if its slower?

Issue #527, opened 2018-09-05
Using already available source feature vocabularies when training a new model


franckbrl:
Hello,

I have a `params` file, as well as vocabulary files, and I want to us them to continue the training on new data (for fine-tuning).

With `sockeye.train`, I can do:

```
--source-vocab vocab.src.0.json
--target-vocab vocab.trg.0.json
```

But I also have a `vocab.src.1.json` that I can't pass as argument, and therefore needs to be recomputed on the data. This is risky, since the vocabulary frequencies can be different given different data, which could lead to different indexes than the ones in my initial `vocab.src.1.json`.

I guess what I'm looking for is something like:

```
--source-vocabs vocab.src.0.json vocab.src.1.json
--target-vocab vocab.trg.0.json
```

I have seen the use of `sockeye.prepare_data` [here](https://github.com/awslabs/sockeye/tree/master/tutorials/adapt), but this module also takes one single source vocabulary file.

Is there a way to do this right?

tdomhan:
You are right, this is currently not supported.

What would be needed to change is to make `source_vocabs` a variable length argument and modifying these two locations:
https://github.com/awslabs/sockeye/blob/master/sockeye/train.py#L304
https://github.com/awslabs/sockeye/blob/master/sockeye/prepare_data.py#L49

franckbrl:
Thank you for your answer!

I could work on it and make a PR, if you want.

tdomhan:
Perfect, thanks a lot!

PR #530, opened 2018-09-06
Provide multiple source vocabularies as argument

franckbrl:
Following issue #527, `--source-vocab` can now take multiple files for additional source factor vocabularies.

We may want to consider changing a few variable/parameter names. For instance in `train.py`, now that `args.source_vocab` is a list, we may rename it to `args.source_vocabs` (parameter `--source-vocabs`), but it would probably not go well with the variable `source_vocabs` (produced by `create_data_iters_and_vocabs()`). This would also lead to a backwards-incompatible change.

Unit tests output 11 failures on the current master branch version. Since the content of this PR added no failures, I'm assuming the tests are passed.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
 ## [1.18.53]
-### Changed
-- `--source-vocab` can take multiple files for additional source factor vocabularies.
+### Added
+- `--source-factor-vocabs` can be set for source factor vocabularies.
 
 ## [1.18.52]
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,14 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.54]
+### Added
+- `--source-factor-vocabs` can be set to provide source factor vocabularies.
+
 ## [1.18.53]
 ### Added
-- `--source-factor-vocabs` can be set for source factor vocabularies.
+- Always skipping softmax for greedy decoding by default, only for single models.
+- Added option `--skip-topk` for greedy decoding.
 
 ## [1.18.52]
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.52]
+### Fixed
+- Fixed bug in constrained decoding to make sure best hypothesis satifies all constraints.
+
 ## [1.18.51]
 ### Added
 - Added a CLI for reranking of an nbest list of translations.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.53]
+### Changed
+- `--source-vocab` can take multiple files for additional source factor vocabularies.
+
 ## [1.18.52]
 ### Fixed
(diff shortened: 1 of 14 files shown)


fhieber:
Thanks for adding this! Could you elaborate on what fails for you in unit tests? Current master runs tests just fine for me (and for Travis) so this must be something on your side.

Regarding the breaking change. If we would rename the `--source-vocab` argument for `sockeye.train` and `sockeye.prepare_data`, it would only affect training CLIs. While breaking any user scripts that call these, I would not view this as something that requires a major version bump. So i'd be fine with renaming these args and bumping only the minor version. Thoughts, @tdomhan ?

franckbrl:
Ok, I'll change `--source-vocab` into `--source-vocabs` as soon as I get your confirmation.

The unit tests fails I get are all related to image captioning: see [unit_test.txt](https://github.com/awslabs/sockeye/files/2360231/unit_test.txt).

fhieber:
Do you have python-pillow installed or something? Maybe its a version that is not compatible but it should skip these tests if not installed at all.
@lorisbaz can you advise?

franckbrl:
I do have Pillow installed: 

```
>>> PIL.VERSION
'1.1.7'
```

lorisbaz:
It seems that the problem is:`TypeError: integer argument expected, got float`.
It might be related to the PIL version. I tested it with `5.2.0`.

mjpost:
Note that there is a hidden backward-incompatibility in the 1.18 range with the `lhuc` parameter (if you trained a model with 1.18.early, it won't run with 1.18.now). It's an easy fix (just add the keyword to the config file) but technically should have required a version bump. So together with this, it might be a reason to bump the major version.

franckbrl:
Thanks @lorisbaz, upgrading Pillow to version `5.2.0` removed a few failures, but I still get 6: [unit_test.txt](https://github.com/awslabs/sockeye/files/2361078/unit_test.txt).

```
>>> PIL.__version__
'5.2.0'
```

lorisbaz:
It seems related to multiple sources, since image captioning supports only a single-source input encoders.

fhieber:
@lorisbaz but that doesn't explain the failure of the tests, or does it? On Travis, are these tests being run? I think they are because travis installs the dev requirements, including Pillow.

franckbrl:
I get the same 6 failures on the master branch version. There's certainly an issue with my environment.

fhieber:
Seems like it. Could you rebase onto latest master?

franckbrl:
Sure, so we stay on a minor change? If so, should I increment the Sockeye version to `1.18.54`, or just add a line in `CHANGELOG.md` under version `1.18.53`? Otherwise, we switch to version `1.19.0`.

tdomhan:
I'm not sure this warrants a major version change, as there is no backwards compatibility in the sense that a trained model won't work at inference time anymore. The fact that LHUC apparently introduced a non backwards-compatible changes means that we should fix that separately (namely, being robust against `lhuc` missing).

franckbrl:
The code is rebased on the latest master branch. We now have version `1.18.54`.

Are there any further thoughts on changing `--source-vocab` to `--source-vocabs`?

mjpost:
To keep this backwards compatible, why not add `--source-vocabs` but keep `--source-vocab` (that takes just one argument). Then, when 1.19 comes out, you could remove the latter.

I agree about not incrementing for this relatively small change, but it seems like bad form to actually break everyone's scripts without incrementing.

franckbrl:
I like the idea, although it would require two distinct ways of importing the vocabularies, according to whether `--source-vocab` is set or not. And we would probably need to raise an error when `source_vocabs[0] != source_vocab` and both are set.

Another solution would be to leave `--source-vocab` and create `--source-factor-vocabs` specifically for source factor vocabularies (not tokens). This would be consistent with the way source token and factor data files are provided (`--source` and `--source-factors`). And this would be fully backwards compatible.

mjpost:
The sanity check don't seem like that much of a burden. But I do also like the `--source-factor-vocabs` approach for 1.18, then removing it in 1.19 in favor of `--source-factors`. But this will all be up to the Amazon folks.

tdomhan:
Either of those two solutions sound good. Not breaking people's training scripts is definitely something we should try to do, unless it's unavoidable.

The solution with `--source-factor-vocabs` might be a bit cleaner. Then we could open a issue for version 1.19 in order to not forget reconsidering using a single `--source-vocabs` argument.

franckbrl:
`--source-vocab` remains unchanged and `--source-factor-vocabs` was just added.

fhieber:
yes, thanks for the change!

PR #532, opened 2018-09-07
add nbest translation

bricksdont:
Adding nbest translation. Details:
- `Translation` and `TranslatorOutput` objects now have additional attributes for nbest translations
- nbest translation is exposed as `--nbest-size` in the translate CLI
- if there are reference constraints, we can only guarantee that the _first_ (=best) translation in an nbest list satisfies all constraints. Also, with reference constraints there is a possibility that two hypotheses in the nbest list are identical.
- if sentences are too long and need to be split up for translation and concatenated again, the same happens for their nbest translations. I consider this an edge case.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/checkpoint_decoder.py
```diff
@@ -139,6 +141,7 @@ def decode_and_evaluate(self,
                                           length_penalty=inference.LengthPenalty(self.length_penalty_alpha, self.length_penalty_beta),
                                           beam_prune=0.0,
                                           beam_search_stop='all',
+                                          nbest_size=1,
```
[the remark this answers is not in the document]
bricksdont: Yes, actually using `self.nbest_size` is even better, isn't it?

sockeye/inference.py
```diff
@@ -1297,35 +1396,48 @@ def _make_result(self,
...
+        target_tokens = ids_to_tokens(translation.target_ids)
+        target_tokens_list = [ids_to_tokens(ids) for ids in nbest_target_ids ]
 
-        target_string = C.TOKEN_SEPARATOR.join(
-            tok for target_id, tok in zip(target_ids, target_tokens) if target_id not in self.strip_ids)
+        def ids_tokens_to_string(target_ids, target_tokens):
+            return C.TOKEN_SEPARATOR.join(
+                tok for target_id, tok in zip(target_ids, target_tokens) if target_id not in self.strip_ids)
+
+        target_string = ids_tokens_to_string(target_ids, target_tokens)
```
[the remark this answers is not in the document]
bricksdont: Yes, I like this idea. `NBestTranslation` could have a subset of the attributes of `Translation`. I'll give it a try.

sockeye/inference.py
```diff
@@ -1297,35 +1396,48 @@ def _make_result(self,
...
         :param trans_input: Translator input.
         :param translation: The translation + attention and score.
         :return: TranslatorOutput.
         """
         target_ids = translation.target_ids
+        nbest_target_ids = translation.nbest_target_ids
         attention_matrix = translation.attention_matrix
 
-        target_tokens = [self.vocab_target_inv[target_id] for target_id in target_ids]
+        def ids_to_tokens(ids):
```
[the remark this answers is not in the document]
bricksdont: No, there is no function like that in utils - adding it to utils.

sockeye/inference.py
```diff
@@ -1706,32 +1818,43 @@ def _get_best_from_beam(self,
...
+            best_ids = np.arange(n, self.batch_size * self.beam_size, self.beam_size, dtype='int32')
 
-        histories = beam_histories if beam_histories is not None else [None] * self.batch_size  # type: List
-        return [self._assemble_translation(*x) for x in zip(best_word_indices[indices, np.arange(indices.shape[1])],
-                                                            lengths[best_ids],
-                                                            attentions[best_ids],
-                                                            seq_scores[best_ids],
-                                                            histories)]
+            # only check for constraints for 1-best translation for each sequence in batch
+            if n == 0 and any(constraints):
```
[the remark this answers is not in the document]
bricksdont: Why do we have to? Alternatively, we could clearly state that the behaviour for nbest translation with reference constraints is that only the first hypothesis will satisfy all constraints.

sockeye/constants.py
```diff
@@ -322,14 +323,24 @@
...
                    OUTPUT_HANDLER_ALIGN_TEXT,
-                   OUTPUT_HANDLER_BEAM_STORE]
+                   OUTPUT_HANDLER_BEAM_STORE,
+                   OUTPUT_HANDLER_NBEST,
+                   OUTPUT_HANDLER_NBEST_WITH_SCORE,
+                   OUTPUT_HANDLER_NBEST_WITH_ALIGNMENTS]
+
+OUTPUT_HANDLERS_NBEST = [OUTPUT_HANDLER_NBEST,
+                         OUTPUT_HANDLER_NBEST_WITH_ALIGNMENTS,
+                         OUTPUT_HANDLER_NBEST_WITH_SCORE]
```
mjpost: Instead of expanding the set of output handlers and exposing them as independent choices to the command-line CLI, wouldn't it make sense to just apply the existing output handlers to every item on the n-best list when `--nbest-size > 1`? This seems redundant here.

sockeye/translate.py
```diff
@@ -56,6 +56,13 @@ def run_translate(args: argparse.Namespace):
         check_condition(args.beam_size == 1, "--skip-topk has no effect if beam size is larger than 1")
         check_condition(len(args.models) == 1, "--skip-topk has no effect for decoding with more than 1 model")
 
+    if args.nbest_size > 1:
+        check_condition(args.beam_size >= args.nbest_size,
+                        "Size of nbest list (--nbest-size) must be smaller or equal to beam size (--beam-size).")
+        if args.output_type not in C.OUTPUT_HANDLERS_NBEST:
+            logger.warning("For nbest translation, output handler must be one of %s, overriding option --output-type to '%s'.",
+                       C.OUTPUT_HANDLERS_NBEST, C.OUTPUT_HANDLER_NBEST)
+
```
mjpost: See comment above about just applying the existing output handler to every item in an n-best list.

sockeye/constants.py
bricksdont: Thanks Matt! The question is what behaviour is expected by users. So, you would for instance use `StringOutputHandler` for nbest translation, let's say with `--nbest-size 5` and silently output 5 times the amount of lines?

mjpost: I think that makes more sense to me: the chosen output format gets applied to all items. Then you could add a 'json' output handler that would do different things. It seems convoluted to me to have different output handlers that are all outputting different fields in a JSON object; instead, just return a JSON object with everything, and users can ignore the fields they don't want.

bricksdont: Ok, I could live with that, but still think you could, by the same logic, say that an output handler for 1-best translations that outputs the strings without the scores is redundant. @fhieber @tdomhan Your take?

mjpost: I see your point, but I also see a difference between outputting (a) plain text (which is what the existing output handlers moderate) and (b) a structured object like JSON. For the latter, you might as well just include everything that is free to compute, since you have to parse the object anyway.

fhieber: I would keep the basic contract for the translate CLI that n lines in mean n lines out. This clearly means that nbest_size > 1 should return some structured object that holds multiple hypotheses.
Regarding the contents of that object: the additional data (alignments & scores) are computed anyway so there is no additional cost of adding them to the JSON so the 3 different nbest output handlers seem redundant to me.

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: Right, we don't have to raise but giving a warnings makes sense (people rarely read argument descriptions I think). I'd add a log.warning somewhere in the Translator constructor or where it fits best.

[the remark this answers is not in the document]
bricksdont: Finally getting back to this; in the `Translator` constructor we don't know about constraints yet. Also, it's not a CLI option, just implicitly as an optional field for `--json-input`. We could warn for every batch, in `get_best_from_beam` - but what do you think is best?

sockeye/constants.py
bricksdont: Ok, will remove all handlers except one, which always outputs all fields.

bricksdont:
Just bumping this; I would like to have it in master soon if possible. Is there anything else I should change? Thanks!

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: hm, warning for every request is not a good idea. Maybe its ok as is.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.54]
+### Added
+- Added nbest translation, exposed as `--nbest-size`.
```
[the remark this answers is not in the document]
fhieber: *bump*

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: Any update on this thought?

sockeye/inference.py
```diff
@@ -1102,12 +1211,22 @@ def __init__(self,
                     logger.warning("Global avoid phrase '%s' contains an %s; this may indicate improper preprocessing.", ' '.join(phrase), C.UNK_SYMBOL)
                 self.global_avoid_trie.add_phrase(phrase_ids)
 
+        if self.nbest_size > 1:
```
fhieber: could this be shortened?
```python
self._concat_translations = partial(_concat_nbest_translations if self.nbest > 1 else _concat_translations,
 stop_ids=self.stop_ids,
 length_penalty=self.length_penalty)
```

sockeye/inference.py
```diff
@@ -1309,37 +1428,48 @@ def _make_result(self,
...
+        Returns a translator result from generated target-side word ids, attention matrices and scores.
         Strips stop ids from translation string.
 
         :param trans_input: Translator input.
         :param translation: The translation + attention and score.
         :return: TranslatorOutput.
         """
         target_ids = translation.target_ids
+        target_tokens = utils.ids_to_tokens(translation.target_ids, self.vocab_target_inv)
+        target_string = utils.tokens_to_string(target_ids, target_tokens, self.strip_ids)
```
fhieber: this PR is probably going on top of the scoring PR from @mjpost . He also added a function like this. Please make sure that this is properly merged/unified.

sockeye/constants.py
```diff
@@ -322,14 +323,18 @@
 OUTPUT_HANDLER_ALIGN_PLOT = "align_plot"
 OUTPUT_HANDLER_ALIGN_TEXT = "align_text"
 OUTPUT_HANDLER_BEAM_STORE = "beam_store"
+OUTPUT_HANDLER_NBEST = "nbest_translation"
+OUTPUT_HANDLER_NBEST_WITH_SCORE = "nbest_translation_with_score"
```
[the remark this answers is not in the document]
bricksdont: Yes, absolutely - removing.

sockeye/inference.py
bricksdont: Sure. Then I should probably wait until the scoring PR is merged.

sockeye/data_io.py
```diff
@@ -1042,7 +1042,7 @@ def ids2tokens(token_ids: Iterable[int],
 """
 
     tokens = [vocab_inv[token] for token in token_ids]
-    return (tok for token_id, tok in zip(token_ids, tokens) if token_id not in exclude_set)
+    return [tok for token_id, tok in zip(token_ids, tokens) if token_id not in exclude_set]
```
fhieber: please leave this as a generator as discussed in the scoring PR

bricksdont: Sorry about that - missed the discussion.

davvil:
Continued in #558.

PR #538, opened 2018-09-18
[WIP] Scoring

mjpost:
This implements scoring by fully reusing the training computation graph, per @bricksdont's original suggestion. It replaces PR #413.

It's nearly done, but I thought I'd submit it as WIP since it makes some changes required to generalize many of the procedures. These include:

- Making validation data optional when creating data iterators
- Adding a new data iterator `fill_up` policy that pads an uneven batch with zeros and does not randomly permute batches

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/modules.rst
+++ b/docs/modules.rst
@@ -176,6 +176,13 @@ sockeye.rnn_attention module
     :members:
     :show-inheritance:
 
+sockeye.score module
+----------------------------
+
+.. automodule:: sockeye.score
+    :members:
+    :show-inheritance:
+
 sockeye.train module
 --------------------
 
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -1092,6 +1092,17 @@ def add_score_cli_args(params):
     params.add_argument("--model", "-m", required=True,
                         help="Model directory containing trained model.")
 
+    params.add_argument('--length-penalty-alpha',
+                        default=1.0,
+                        type=float,
+                        help='Alpha factor for the length penalty used in beam search: '
+                        '(beta + len(Y))**alpha/(beta + 1)**alpha. A value of 0.0 will therefore turn off '
+                        'length normalization. Default: %(default)s')
+    params.add_argument('--length-penalty-beta',
+                        default=0.0,
+                        type=float,
+                        help='Beta factor for the length penalty used in beam search: '
+                        '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
 
 def add_max_output_cli_args(params):
     params.add_argument('--max-output-length',
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -782,7 +782,7 @@ def add_scoring_args(params):
 def add_training_args(params):
     train_params = params.add_argument_group("Training parameters")
 
-    add_batch_args(training_params)
+    add_batch_args(train_params)
 
     train_params.add_argument('--decoder-only',
                                action='store_true',
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -754,33 +754,43 @@ def add_model_parameters(params):
                                    "(and all convolutional weight matrices for CNN decoders). Default: %(default)s.")
 
 
+def add_batch_args(params):
+    params.add_argument('--batch-size', '-b',
+                        type=int_greater_or_equal(1),
+                        default=4096,
+                        help='Mini-batch size. Note that depending on the batch-type this either refers to '
+                             'words or sentences.'
+                             'Sentence: each batch contains X sentences, number of words varies. '
+                             'Word: each batch contains (approximately) X words, number of sentences varies. '
+                             'Default: %(default)s.')
+    params.add_argument("--batch-type",
+                        type=str,
+                        default=C.BATCH_TYPE_WORD,
+                        choices=[C.BATCH_TYPE_SENTENCE, C.BATCH_TYPE_WORD],
+                        help="Sentence: each batch contains X sentences, number of words varies."
+                             "Word: each batch contains (approximately) X target words, "
+                             "number of sentences varies. Default: %(default)s.")
+
+
+
+def add_scoring_args(params):
+    scoring_params = params.add_argument_group("Scoring parameters")
+
+    add_batch_args(scoring_params)
+
+
 def add_training_args(params):
     train_params = params.add_argument_group("Training parameters")
+
+    add_batch_args(training_params)
 
     train_params.add_argument('--decoder-only',
                                action='store_true',
                                help='Pre-train a decoder. This is currently for RNN decoders only. '
                                     'Default: %(default)s.')
-
-    train_params.add_argument('--batch-size', '-b',
-                              type=int_greater_or_equal(1),
-                              default=4096,
-                              help='Mini-batch size. Note that depending on the batch-type this either refers to '
-                                   'words or sentences.'
-                                   'Sentence: each batch contains X sentences, number of words varies. '
-                                   'Word: each batch contains (approximately) X words, number of sentences varies. '
-                                   'Default: %(default)s.')
-    train_params.add_argument("--batch-type",
-                              type=str,
-                              default=C.BATCH_TYPE_WORD,
-                              choices=[C.BATCH_TYPE_SENTENCE, C.BATCH_TYPE_WORD],
-                              help="Sentence: each batch contains X sentences, number of words varies."
-                                   "Word: each batch contains (approximately) X target words, "
-                                   "number of sentences varies. Default: %(default)s.")
-
     train_params.add_argument('--fill-up',
                               type=str,
-                              default='replicate',
+                              default=C.DEFAULT_FILL_UP_STRATEGY,
                               help=argparse.SUPPRESS)
 
     train_params.add_argument('--loss',
@@ -1067,6 +1077,20 @@ def add_translate_cli_args(params):
     add_inference_args(params)
     add_device_args(params)
     add_logging_args(params)
+
+
+def add_score_cli_args(params):
+    add_training_data_args(params, required=False)
+    add_prepared_data_args(params)
+    add_vocab_args(params)
+    add_bucketing_args(params)
+    add_scoring_args(params)
+    add_device_args(params)
+    add_logging_args(params)
+
+    params = params.add_argument_group("Scoring parameters")
+    params.add_argument("--model", "-m", required=True,
+                        help="Model directory containing trained model.")
 
 
 def add_max_output_cli_args(params):
--- a/sockeye/constants.py
+++ b/sockeye/constants.py
@@ -393,6 +393,7 @@ DATA_INFO = "data.info"
 DATA_CONFIG = "data.config"
 PREPARED_DATA_VERSION_FILE = "data.version"
 PREPARED_DATA_VERSION = 2
+DEFAULT_FILL_UP_STRATEGY = 'replicate'
 
 # reranking
 RERANK_BLEU = "bleu"
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -751,8 +751,8 @@ def get_prepared_data_iters(prepared_data_dir: str,
 
 def get_training_data_iters(sources: List[str],
                             target: str,
-                            validation_sources: List[str],
-                            validation_target: str,
+                            validation_sources: Optional[List[str]],
+                            validation_target: Optional[str],
                             source_vocabs: List[vocab.Vocab],
                             target_vocab: vocab.Vocab,
                             source_vocab_paths: List[Optional[str]],
@@ -842,17 +842,19 @@ def get_training_data_iters(sources: List[str],
                                     bucket_batch_sizes=bucket_batch_sizes,
                                     num_factors=len(sources))
 
-    validation_iter = get_validation_data_iter(data_loader=data_loader,
-                                               validation_sources=validation_sources,
-                                               validation_target=validation_target,
-                                               buckets=buckets,
-                                               bucket_batch_sizes=bucket_batch_sizes,
-                                               source_vocabs=source_vocabs,
-                                               target_vocab=target_vocab,
-                                               max_seq_len_source=max_seq_len_source,
-                                               max_seq_len_target=max_seq_len_target,
-                                               batch_size=batch_size,
-                                               fill_up=fill_up)
+    validation_iter = None
+    if validation_sources is not None and validation_target is not None:
+        validation_iter = get_validation_data_iter(data_loader=data_loader,
+                                                   validation_sources=validation_sources,
+                                                   validation_target=validation_target,
+                                                   buckets=buckets,
+                                                   bucket_batch_sizes=bucket_batch_sizes,
+                                                   source_vocabs=source_vocabs,
+                                                   target_vocab=target_vocab,
+                                                   max_seq_len_source=max_seq_len_source,
+                                                   max_seq_len_target=max_seq_len_target,
+                                                   batch_size=batch_size,
+                                                   fill_up=fill_up)
 
     return train_iter, validation_iter, config_data, data_info
 
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -672,9 +672,10 @@ def get_prepared_data_iters(prepared_data_dir: str,
                             batch_size: int,
                             batch_by_words: bool,
                             batch_num_devices: int,
-                            fill_up: str) -> Tuple['BaseParallelSampleIter',
-                                                   'BaseParallelSampleIter',
-                                                   'DataConfig', List[vocab.Vocab], vocab.Vocab]:
+                            fill_up: str,
+                            no_permute: bool = False) -> Tuple['BaseParallelSampleIter',
+                                                               'BaseParallelSampleIter',
+                                                               'DataConfig', List[vocab.Vocab], vocab.Vocab]:
     logger.info("===============================")
     logger.info("Creating training data iterator")
     logger.info("===============================")
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -728,7 +728,8 @@ def get_prepared_data_iters(prepared_data_dir: str,
                                            batch_size,
                                            bucket_batch_sizes,
                                            fill_up,
-                                           num_factors=len(data_info.sources))
+                                           num_factors=len(data_info.sources),
+                                           no_permute=no_permute)
 
     data_loader = RawParallelDatasetLoader(buckets=buckets,
                                            eos_id=target_vocab[C.EOS_SYMBOL],
@@ -765,9 +766,10 @@ def get_training_data_iters(sources: List[str],
                             max_seq_len_source: int,
                             max_seq_len_target: int,
                             bucketing: bool,
-                            bucket_width: int) -> Tuple['BaseParallelSampleIter',
-                                                        'BaseParallelSampleIter',
-                                                        'DataConfig', 'DataInfo']:
+                            bucket_width: int,
+                            no_permute: bool = False) -> Tuple['BaseParallelSampleIter',
+                                                               'BaseParallelSampleIter',
+                                                               'DataConfig', 'DataInfo']:
     """
     Returns data iterators for training and validation data.
 
@@ -840,7 +842,8 @@ def get_training_data_iters(sources: List[str],
                                     buckets=buckets,
                                     batch_size=batch_size,
                                     bucket_batch_sizes=bucket_batch_sizes,
-                                    num_factors=len(sources))
+                                    num_factors=len(sources),
+                                    no_permute=no_permute)
 
     validation_iter = None
     if validation_sources is not None and validation_target is not None:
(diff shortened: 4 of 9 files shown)


ciyongch:
From the user's perspective, when should we use `ScoringModel`, any scenarios? Thanks

mjpost:
The most straightforward application is to use it to discover the model scores for translating one sentence to another, via the CLI. People have used this, for example, [for corpus-level filtering](https://arxiv.org/pdf/1809.00196.pdf) (§4). But there are many other things one could use it for.

bricksdont:
@ciyongch Another prominent application is [contrastive evaluation](http://aclweb.org/anthology/E17-2060.pdf) of NMT models.

bricksdont:
Why is does the output have the source sentence but not the target? I think that a good default is to just output the model scores, one per input sentence pair. Otherwise, we need output handlers for scoring.

mjpost:
That was just an arbitrary choice that I used for debugging (the sentences weren't processed in order until I figured out how to turn off random permutations in the data iterator). I agree it should just output the score, with command-line parameters for optionally outputting other information. I'll add those.

ciyongch:
Got it, thanks @mjpost @bricksdont .

sockeye/scoring.py
```diff
@@ -0,0 +1,270 @@
...
+            # target_decoded: (batch-size, target_len, decoder_depth)
+            target_decoded = self.decoder.decode_sequence(source_encoded, source_encoded_length, source_encoded_seq_len,
+                                                          target_embed, target_embed_length, target_embed_seq_len)
+
+            # target_decoded: (batch_size * target_seq_len, decoder_depth)
+            target_decoded = mx.sym.reshape(data=target_decoded, shape=(-3, 0))
+
+            # output layer
+            # logits: (batch_size * target_seq_len, target_vocab_size)
+            logits = self.output_layer(target_decoded)
```
fhieber: does output_layer also accept `(batch, seq_len, depth)` tensors? If yes we can get rid of the reshape above, run the softmax over axis=1 and return a 3d tensor. Then you don't need to reshape the outputs in the `Scorer`.

sockeye/scoring.py
```diff
@@ -0,0 +1,270 @@
...
+            outputs = self.model.get_outputs()
+
+            batch_size, target_seq_len, _ = batch.provide_data[0][1]
+            outputs = mx.nd.reshape(data=outputs[0], shape=(-4, batch_size, target_seq_len, -2))
+
+            probs = mx.nd.pick(outputs, labels)
+            ones = mx.nd.ones_like(probs, ctx=self.model.context)
+            lengths = mx.nd.sum(labels != 0, axis=1) - 1
+
+            scores = mx.nd.log(mx.nd.where(labels != 0, probs, ones, ctx=self.model.context), ctx=self.model.context)
```
fhieber: Everything from getting the outputs up to here could be part of the computation graph, no need to do this imperatively.

sockeye/scoring.py
```diff
@@ -0,0 +1,270 @@
...
+            if score_type == C.SCORING_TYPE_NEGLOGPROB:
+                scores = -1 * scores
+            sums = mx.nd.sum(scores, axis=1) / self.length_penalty(lengths)
+            sums = sums.asnumpy().tolist()
+            for source, target, score in zip(batch.data[0], batch.data[1], sums):
```
fhieber: as said above, output should be handled by either some OutputHandler, or maybe refactor this into a separate function that takes tensors of source, target, score as input and yields output lines.

sockeye/score.py
```diff
@@ -0,0 +1,141 @@
...
+from .config import Config
+from .log import setup_main_logger
+from .optimizers import OptimizerConfig
+from .utils import check_condition, log_basic_info
+
+# Temporary logger, the real one (logging to a file probably, will be created in the main function)
+logger = setup_main_logger(__name__, file_logging=False, console=True)
+
+
+def check_arg_compatibility(args: argparse.Namespace):
```
fhieber: is this TODO? If not, let's remove it.

sockeye/data_io.py
```diff
@@ -1256,23 +1278,38 @@ def fill_up(self,
...
 
+                if isinstance(source[bucket_idx], np.ndarray):
+                    source[bucket_idx] = np.concatenate((bucket_source, bucket_source.take(desired_indices_np)), axis=0)
+                else:
+                    source[bucket_idx] = mx.nd.concat(bucket_source, bucket_source.take(desired_indices), dim=0)
+                target[bucket_idx] = mx.nd.concat(bucket_target, bucket_target.take(desired_indices), dim=0)
+                label[bucket_idx] = mx.nd.concat(bucket_label, bucket_label.take(desired_indices), dim=0)
+
+                if fill_up == 'zeros':
+                    source[bucket_idx][num_samples:,:,:] = 0
```
fhieber: spacing :)

fhieber: might make sense to fill these with C.PAD_ID (which of course is 0).

But could you make sure that this is actually necessary? I might miss some context, but for example `RawParallelDatasetLoader.load()` creates source/target/label arrays that are filled with C.PAD_ID (0) already.

sockeye/data_io.py
```diff
@@ -1019,6 +1025,22 @@ def ids2strids(ids: Iterable[int]) -> str:
     return " ".join(map(str, ids))
 
 
+def ids2tokens(token_ids: Iterable[int],
```
fhieber: would that be better placed in utils.py where other token2id and vice versa methods are defined?

sockeye/scoring.py
```diff
@@ -0,0 +1,239 @@
...
+        self.exclude_list = [source_vocabs[0][C.BOS_SYMBOL], target_vocab[C.EOS_SYMBOL], C.PAD_ID]
+
+    def score(self,
+              score_iter):
+
+        for i, batch in enumerate(score_iter):
+            # data_io generates labels, too, which we don't need
+            label, batch.provide_label = batch.provide_label, None
+            labels = batch.label[0]
+            self.model.prepare_batch(batch)
```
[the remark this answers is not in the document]
mjpost: I think I am preparing the current one. The problem is that the batch includes labels but the computation graph doesn't, so I have to remove them. However, I still need them for the `pick` call later on. I could just use the labels in the `ScoringModel` CG, but I want the model to be able to return the whole distributions (we have a use case for this in mind).

Or am I misunderstanding something from your comment? (Very likely).

sockeye/utils.py
```diff
@@ -953,3 +953,18 @@ def split(data: mx.nd.NDArray,
     if num_outputs == 1:
         return [ndarray_or_list]
     return ndarray_or_list
+
+
+def inflect(noun: str,
```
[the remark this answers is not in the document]
mjpost: No, but isn't it nice?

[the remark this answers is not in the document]
fhieber: if you feel strongly about it :)

sockeye/data_io.py
```diff
@@ -1256,23 +1278,38 @@ def fill_up(self,
             bucket_label = self.label[bucket_idx]
             num_samples = bucket_source.shape[0]
 
+            # Fill up the last batch according to the fill up policy.
+            # 'replicate' randomly samples to populate the remaining spots.
+            # 'repeate_last' instead repeats the last element.
```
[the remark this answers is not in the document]
mjpost: That's a type — I originally used that name instead of "zeros".

sockeye/scoring.py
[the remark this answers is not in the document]
fhieber: Similar to my other comment: why don't you put the labels into the computation graph. You can always return a grouped symbol that contains both, output of pick and the distributions.

sockeye/score.py
```diff
@@ -0,0 +1,141 @@
...
+from . import initializer
+from . import loss
+from . import lr_scheduler
+from . import model
+from . import rnn
+from . import rnn_attention
+from . import scoring
+from . import train
+from . import training
+from . import transformer
```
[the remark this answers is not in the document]
mjpost: Ah, yes, leftover from copy-paste. Will fix.

sockeye/constants.py
```diff
@@ -393,8 +393,22 @@
...
 RERANK_METRICS = [RERANK_BLEU, RERANK_CHRF]
+
+# scoring
+SCORING_OUTPUT_ID = 'id'
+SCORING_OUTPUT_SCORE = 'score'
+SCORING_OUTPUT_SOURCE = 'source'
+SCORING_OUTPUT_TARGET = 'target'
+SCORING_OUTPUT_UNKNOWN = '[UNKNOWN_TYPE]'
+SCORING_OUTPUT_DEFAULT = SCORING_OUTPUT_SCORE
+SCORING_OUTPUT_CHOICES = [SCORING_OUTPUT_SCORE, SCORING_OUTPUT_ID, SCORING_OUTPUT_SOURCE, SCORING_OUTPUT_TARGET]
```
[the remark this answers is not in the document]
mjpost: It would be good to combine them. One option here is that I like the flexibility of having `outputs` take multiple arguments, rather than predefining subsets of all possible outputs (i.e., `--outputs score translation` instead of `--output-format translation_with_score`). That way the user can specify whatever she likes. What do you think?

sockeye/scoring.py
```diff
@@ -0,0 +1,270 @@
...
+                                        context=self.context)
+
+        self.module.bind(data_shapes=provide_data,
+                         label_shapes=None,
+                         for_training=False,
+                         force_rebind=False,
+                         grad_req=None)
+
+
+    def prepare_batch(self, batch: mx.io.DataBatch):
```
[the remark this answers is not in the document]
mjpost: What do you mean? Do you mean I should just call self.module.prepare(batch) directly?

sockeye/scoring.py
mjpost: This is true, but having the `ScoringModel` return distributions could be useful (and in fact, is useful) for other applications.

A cleaner way to do this might be to parameterize `ScoringModel` with the desired output layer, i.e., one that returns either distributions or the desired scores. Do you prefer that?

sockeye/constants.py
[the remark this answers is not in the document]
fhieber: yeah, that'd be nice but its kind of orthogonal to the scoring feature in general. In essence you would only have a single OutputHandler with different configuration depending on the commandline specification of the user. Might make sense to add that in the future, but in a separate PR.

sockeye/scoring.py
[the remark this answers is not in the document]
fhieber: There is no need to 'prepare a batch' in general. What it does internally is switching the current module in the BucketingModule to the correct one for the NEXT batch. It does this switching anyway however when you fall forward with a batch. So unless you are not using the next batch but the current one there is really no point in doing this.

sockeye/scoring.py
fhieber: why not both by default? Graphs can have multiple outputs (grouped symbol).

sockeye/data_io.py
mjpost: It does seem necessary in that removing this code results in an underfilled batch that is then ignored, but I don't quite understand why. I'll look into it.

sockeye/data_io.py
mjpost: Actually those are all defined in data_io; I put this right next to them.

sockeye/scoring.py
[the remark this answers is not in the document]
mjpost: Ah, returning a group is a much better idea. I hadn't thought of it.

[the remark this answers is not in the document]
mjpost: Okay this is much nicer.

sockeye/constants.py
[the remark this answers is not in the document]
mjpost: Done.

sockeye/train.py
```diff
@@ -219,10 +219,16 @@ def create_data_iters_and_vocabs(args: argparse.Namespace,
                                  max_seq_len_target: int,
                                  shared_vocab: bool,
                                  resume_training: bool,
-                                 output_folder: str) -> Tuple['data_io.BaseParallelSampleIter',
-                                                              'data_io.BaseParallelSampleIter',
-                                                              'data_io.DataConfig',
-                                                              List[vocab.Vocab], vocab.Vocab]:
+                                 validation_sources: Optional[List[str]] = None,
```
fhieber: I probably need to look better into this, but why is this suddenly getting all these new arguments?

mjpost: I had to pull out arguments from `args` that are used in training and not in scoring.

sockeye/score.py
```diff
@@ -0,0 +1,114 @@
...
+            max_seq_len_source, max_seq_len_target = args.max_seq_len
+
+        score_iter, _, config_data, source_vocabs, target_vocab, data_info = train.create_data_iters_and_vocabs(
+            args=args,
+            max_seq_len_source=max_seq_len_source,
+            max_seq_len_target=max_seq_len_target,
+            shared_vocab=args.shared_vocab,
+            resume_training=True,
+            output_folder=args.model,
+            bucketing=False,
```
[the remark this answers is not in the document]
mjpost: Bucketing destroys source ordering, so this is a necessary evil.

sockeye/arguments.py
```diff
@@ -1108,7 +1153,7 @@ def add_inference_args(params):
                                     "Optionally, a list of factors can be provided: "
                                     "{'text': 'some input string', 'factors': ['C C C', 'X X X']}.")
 
-    decode_params.add_argument(C.INFERENCE_ARG_OUTPUT_LONG, C.INFERENCE_ARG_OUTPUT_SHORT,
+    decode_params.add_argument('--output', '-o',
```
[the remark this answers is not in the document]
mjpost: Just an annoyance when searching in code. I thought others shared this (based on a conv). Can revert if you like. Makes sense in other spots but less so here maybe.

sockeye/data_io.py
```diff
@@ -1256,23 +1278,38 @@ def fill_up(self,
...
-                                rest, num_samples, bucket, bucket_batch_size)
-                    random_indices_np = rs.randint(num_samples, size=rest)
-                    random_indices = mx.nd.array(random_indices_np)
-                    if isinstance(source[bucket_idx], np.ndarray):
-                        source[bucket_idx] = np.concatenate((bucket_source, bucket_source.take(random_indices_np)), axis=0)
-                    else:
-                        source[bucket_idx] = mx.nd.concat(bucket_source, bucket_source.take(random_indices), dim=0)
-                    target[bucket_idx] = mx.nd.concat(bucket_target, bucket_target.take(random_indices), dim=0)
-                    label[bucket_idx] = mx.nd.concat(bucket_label, bucket_label.take(random_indices), dim=0)
+                    logger.info("Filling bucket %s from size %d to %d by sampling with replacement",
```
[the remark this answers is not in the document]
mjpost: I was a bit confused by exactly what it was doing and thought this version was clearer. I’ll get rid of `inflect` though.

sockeye/data_io.py
```diff
@@ -1256,23 +1278,38 @@ def fill_up(self,
...
+                                bucket, num_samples, bucket_batch_size)
+                    desired_indices_np = rs.randint(num_samples, size=rest)
+                    desired_indices = mx.nd.array(desired_indices_np)
+
+                elif fill_up == 'zeros':
+                    logger.info("Filling bucket %s from size %d to %d by repeating the last element %d %s",
```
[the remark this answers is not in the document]
mjpost: It repeats the last one and then sets them all to zeros. But I think it’s clearer to have it just fill the batch randomly always and then set to zeros if requested. Will add soon.

sockeye/score.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            max_seq_len_source, max_seq_len_target = args.max_seq_len
+
+        # This call has a number of different parameters compared to training which reflect our need to get scores
+        # one-for-one and in order with the input data.
+        # Bucketing and permuting need to be turned off in order to preserve the ordering of sentences.
+        # The 'zeros' fill_up strategy fills underfilled buckets with zeros which can then be used to find the last item.
+        # Finally, 'resume_training' needs to be set to True because it causes the model to be loaded instead of initialized.
+        score_iter, _, config_data, source_vocabs, target_vocab, data_info = train.create_data_iters_and_vocabs(
+            args=args,
```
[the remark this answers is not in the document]
fhieber: Ok, got your point now, thanks for the explanation. I agree that passing args is somehow lazy, but one could say that your approach of reusing `create_data_iters_and_vocabs()` for scoring also is: you are going quite some way to overload the semantics of certain arguments, such that this function fits your needs (i.e. by enabling resume_training for something that is definitely not training), which effectively renders half of the (admittedly huge function) useless.
To me it looks like there are two paths forward to this: we either stay lazy, but then changing the args namespace would work equally well, would entail fewer changes in existing code, and is local to score.py; or we do a proper refactor of this by splitting up `create_data_iters_and_vocabs()` so that pieces that are needed by train.py and score.py can be reused.
I'd be fine either way :)

sockeye/data_io.py
mjpost: Okay, this is clear now. `RawParallelDatasetLoader.load()` creates a list of ndarrays of shape `(num_samples, len)` for the buckets. `num_samples` is different for each bucket, obviously, and is not guaranteed to be a multiple of the batch size. That is fixed when `ParallelDataset.fill_up()` is called.

Put more succinctly, `load()` pads columns, `fill_up()` pads rows.

sockeye/score.py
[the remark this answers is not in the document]
mjpost: This makes sense; I hadn't thought about the semantics of `resume_training`, for some reason. I agree with you now that we should either use `args` or a separate function. I'll look at this tomorrow and see which of your two options makes more sense.

mjpost:
Okay, I made a number of changes: 

- `score` no longer reuses code in `train`. I mostly reverted everything in `train.py`.
- scoring no longer supports prepared data
- source factors works now
- I added test cases

It should be all set apart from nitpicks (nitpicks welcome).

docs/modules.rst
```diff
@@ -176,6 +176,13 @@ sockeye.rnn_attention module
     :members:
     :show-inheritance:
 
+sockeye.score module
+----------------------------
```
fhieber: [nitpick] sphinx will complain that these dashes are longer than the line above (I know its silly)

sockeye/arguments.py
```diff
@@ -1075,6 +1079,56 @@ def add_translate_cli_args(params):
     add_logging_args(params)
 
 
+def add_score_cli_args(params):
+    add_training_data_args(params, required=False)
+    add_vocab_args(params)
+    add_device_args(params)
```
fhieber: just out of curiosity: would scoring work with multiple gpus where each gpu gets a chunk of batch size?

sockeye/data_io.py
```diff
@@ -1235,13 +1258,13 @@ def load(fname: str) -> 'ParallelDataSet':
 
     def fill_up(self,
                 bucket_batch_sizes: List[BucketBatchSize],
-                fill_up: str,
+                policy: str,
                 seed: int = 42) -> 'ParallelDataSet':
         """
         Returns a new dataset with buckets filled up using the specified fill-up strategy.
```
fhieber: [nit] strategy -> policy

sockeye/data_io.py
```diff
@@ -1022,6 +1029,22 @@ def ids2strids(ids: Iterable[int]) -> str:
     return " ".join(map(str, ids))
 
 
+def ids2tokens(token_ids: Iterable[int],
+               vocab_inv: Dict[int, str],
+               exclude_set: Set[int] = set()) -> List[str]:
```
fhieber: The typical use case of this function is to join the resulting list on C.TOKEN_SEPARATOR. In that case its better to return a generator (`return (tok for token_id ....)`).

sockeye/scoring.py
```diff
@@ -0,0 +1,266 @@
...
+                                        logger=logger,
+                                        context=self.context)
+
+        self.module.bind(data_shapes=provide_data,
+                         label_shapes=provide_label,
+                         for_training=False,
+                         force_rebind=False,
+                         grad_req=None)
+
+    def run(self, batch: mx.io.DataBatch) -> mx.sym.Symbol:
```
fhieber: return type annotation is wrong. I think its `List[mx.nd.NDArray]` or something.

sockeye/scoring.py
```diff
@@ -0,0 +1,266 @@
...
+                                        data_names=data_names,
+                                        label_names=label_names,
+                                        logger=logger,
+                                        context=self.context)
+
+        self.module.bind(data_shapes=provide_data,
+                         label_shapes=provide_label,
+                         for_training=False,
+                         force_rebind=False,
+                         grad_req=None)
```
fhieber: I think `grad_req` should be set to 'null', not None. This could save you memory :)
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/executor_group.py#L182

test/common.py
```diff
@@ -331,22 +336,39 @@ def run_train_translate(train_params: str,
         cp_metrics = cp_decoder.decode_and_evaluate()
         logger.info("Checkpoint decoder metrics: %s", cp_metrics)
 
+        # import shutil
+        # shutil.copytree(model_path, "/Users/post/code/sockeye/t/model")
```
fhieber: please remove :)

test/common.py
```diff
@@ -331,22 +336,39 @@ def run_train_translate(train_params: str,
...
+                                                                        translate_params)
 
         if test_source_factor_paths is not None:
             params += _TRANSLATE_WITH_FACTORS_COMMON.format(input_factors=" ".join(test_source_factor_paths))
 
         with patch.object(sys, "argv", params.split()):
             sockeye.translate.main()
 
+        # Break out translation and score
+        outputs = open(out_path).readlines()
```
fhieber: please use with statement for this

test/common.py
```diff
@@ -331,22 +336,39 @@ def run_train_translate(train_params: str,
...
+        with open(out_path, 'w') as out_translate, open(translate_score_path, 'w') as out_scores:
+            for output in outputs:
+                output = output.strip()
+                try:
+                    score, translation = output.split('\t')
+                except:
```
fhieber: lets avoid bare except statements. This should be a ValueError, but why would we expect this to fail? The test is in control of the output type. Lets remove the try-except alltogether

test/common.py
```diff
@@ -403,6 +425,44 @@ def run_train_translate(train_params: str,
...
+
+        # Test scoring. We make sure that we can score the (input, translation output) and get the same
+        # model score.
+        if not use_prepared_data:
+            ## Score
+            # We use the translation parameters, but have to remove irrelevant arguments from it.
+            # Currently, the only relevant flag passed is the --softmax-temperature flag.
+            score_params = ''
+            if 'softmax-temperature' in translate_params:
+                tokens = translate_params.split(C.TOKEN_SEPARATOR)
```
fhieber: tokens is a confusing name, why not `arg` or `param`?

test/common.py
```diff
@@ -403,6 +425,44 @@ def run_train_translate(train_params: str,
...
+
+            if test_source_factor_paths is not None:
+                params += _SCORE_WITH_FACTORS_COMMON.format(source_factors=" ".join(test_source_factor_paths))
+
+            with patch.object(sys, "argv", params.split()):
+                sockeye.score.main()
+
+            ## Compare scored output to original translation output. First remove -inf lines from the translate_score_path
+            ## file. These correspond to blank lines in translate, which are skipped in sockeye.score.
+            for translate_score, score_score in zip(filter(lambda x: x != '-inf\n', open(translate_score_path).readlines()),
```
fhieber: please add some ExitStack context or use with statements for opening files.

test/common.py
```diff
@@ -403,6 +425,44 @@ def run_train_translate(train_params: str,
...
+                                                    open(scores_output_file).readlines()):
+                translate_score = float(translate_score)
+                score_score = float(score_score)
+                print('SCORES', translate_score, score_score)
```
fhieber: remove

test/common.py
```diff
@@ -403,6 +425,44 @@ def run_train_translate(train_params: str,
...
+                assert abs(translate_score - score_score) < 0.002
```
fhieber: just 0.002 ? Do we except such large diffs?

mjpost:
Yeah, it runs fine finally. I'm not sure what the issue is but will figure it out.

test/common.py
mjpost: I find values often differ on the third point to the right of the decimal (e.g., 1.567 versus 1.568). Testing for `<= 0.001` doesn't cover this case.

test/common.py
mjpost: For blank lines (which are explicitly created so as to be tested), the output when using `translation_with_score` is just the score (so only a single field).

sockeye/arguments.py
mjpost: I haven't written any code to do this. Is it handled transparently by MXNet, or do you have to do it manually in train?

sockeye/data_io.py
mjpost: I did not know you could create a generator in this way!

mjpost:
It looks like the issue is with `--skip-topk`. This makes sense, but I don't understand why it's passing locally.

mjpost:
Okay, sometimes the outputs were empty due to `max_seq_len` being overrun, leading to the tests trivially passing.

@tdomhan, do you know what the intended semantics of `max_seq_len` (for training) is? Is it supposed to include the hidden <eos> symbol or no?

sockeye/arguments.py
mjpost: Should just work then.

mjpost:
Okay, I'm going to have to give up on this for the near future. Messing with travis tests has consumed nearly my entire day. It passes on OS X and in a Linux environment for me, but fails in travis. The travis output also shows that the translator is outputting scores of both "inf" and "-inf" (search for "TRANSLATE SCORES" in the travis output). This doesn't make any sense and also does not happen in my environments, using either newer or older versions of pytest et al.

mjpost:
This last pytest failure seems to be that some poorly-trained models result in translations that contain multiple <unk> and <s> tokens, which `sockeye.score` then does not score correctly.

mjpost:
I did run them on two systems and that helped me uncover some of these cases. Note that Travis runs them, too, right? 

matt (from my phone)

fhieber:
Travis runs only a small subset for commits and a larger set with the nightly cron.
Good to hear that they pass.

fhieber:
One more thing: can you add an entry point for the new cli in setup.py?
Thanks!

Issue #542, opened 2018-09-24
Warnings with latest pytest


bricksdont:
With current master and latest pytest, I get warnings of this kind:

    /Users/mathiasmuller/miniconda2/envs/py36/lib/python3.6/site-packages/_pytest/python.py:198:
    RemovedInPytest4Warning: Fixture "mock_translator" called directly.
    Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters.
    See https://docs.pytest.org/en/latest/fixture.html for more information.

It's a recent addition to pytest, for context:

https://github.com/pytest-dev/pytest/issues/3661

Perhaps the requirements should specify a slightly older version of pytest (I can open a PR for that), or we should add ignore statements in the code (if that is possible in this case, I have no idea)?

Pytest version:

```sh
$ pip freeze | grep pytest
pytest==3.8.1
pytest-cov==2.6.0
```

fhieber:
These style checkers get more and more annoying :)
We are already fixing a version of mypy because of similar issues. Please go ahead with a PR that locks pytest to a certain version.

mjdenkowski:
Do we like fixing the version of pytest versus adding an ignore statement to the code?  Either way, the warning is ignored and we might want positive features of future versions of pytest.

fhieber:
good point. If its easy enough to add an ignore statement, lets prefer that. (In the case of mypy this was pretty hard so we opted for the version fixture.)

fhieber:
For some reason I read this whole issue as related to pylint...
I was able to reproduce the warnings and will take a look.

fhieber:
In #543 I just remove the fixture annotation and everything still passes.

tdomhan:
well I guess the point of fixing to a certain version, at least for Travis, was to avoid the case where a version bump in a dependency could suddenly block our CI pipeline.  For a particular warning in this case it could be fine to add a ignore statement though.

bricksdont:
Fixed by 3e554adc30c0495c5939f94ccdf940c11d9d6421, closing.

PR #543, opened 2018-09-24
Remove pytest.fixtures to avoid pytest warnings

fhieber:
Removes warnings mentioned in #542 

until you can check this box.

### DIFF
--- a/test/unit/test_inference.py
+++ b/test/unit/test_inference.py
@@ -29,7 +29,6 @@ _BOS = 0
 _EOS = -1
 
 
-@pytest.fixture
 def mock_translator(batch_size: int = 1,
                     beam_size: int = 5,
                     beam_prune: float = 0,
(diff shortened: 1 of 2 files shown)


test/unit/test_inference.py
```diff
@@ -29,7 +29,6 @@
 _EOS = -1
 
 
-@pytest.fixture
```
tdomhan: wait, shouldn't it be enough to remove this fixture decorator? The warning was about calling the fixture directly, not about having fixtures in general.

test/unit/test_translate.py
```diff
@@ -27,16 +25,6 @@
             "Test file line 2\n"
 
 
-@pytest.fixture
```
tdomhan: why do we want to remove these?

fhieber: Why would it matter? The test pass as is and there are no fixtures needed. Imo, these just affected readability negatively and didn't really have a useful purpose.

tdomhan: just wondering. That's true, given the one liners, they did not add a ton of value.

Issue #540, opened 2018-09-21
Is there a website for sockeye?


eric-haibin-lin:
opennmt has a website to guide ppl through their APIs, pretrained models and tutorials. Does sockeye have that?

pengzhao-intel:
Hope the website online soon and we'd like to contribute to the new models and tutorials for the hardware-specific tips/flow, like INT8 :)

mjpost:
Okay, it's live: [https://awslabs.github.io/sockeye/](https://awslabs.github.io/sockeye/)

tdomhan:
closing.

PR #549, opened 2018-09-27
website

mjpost:
This change reorganizes the `docs/` directory in order to create a rudimentary website for Sockeye (addressing #540). Among other things, this PR:

- builds a website on the [Github Jekyll framework](https://help.github.com/articles/using-jekyll-as-a-static-site-generator-with-github-pages/).
- removes `tutorials` as a top level directory, folding all that content into the website itself
- moves some of the information in the top-level README into `docs/`
- (perhaps controversially) removes the sphinx files, on the idea that these can be generated locally or browsed through github

You can see a live demo [here](http://mjpost.github.io/sockeye-docs/). The website is quite simple and basic. The idea is to get something in place so that people can contribute, and improve it iteratively down the line.


mjdenkowski:
This is a great first step and something users have been asking about for awhile!  I support merging this version and iterating, including accepting pull requests from others who have expressed interest in adding to our documentation.

It looks like check-manifest couldn't find some files.  Otherwise the demo looks good.

fhieber:
what about the readthedocs build now?
Does `python setup.py docs` still run sphinx and is able to create the documentation hosted at https://sockeye.readthedocs.io/en/latest/ ?

mjpost:
No, I killed that. I thought we had discussed this being okay privately a while back, since people can always just browse through Github.

If you want Sphinx, too, that could work. The problem is they both require use of the `docs/` directory, but I think the only file overlap is `index.md`.

mjpost:
I haven't tried Sphinx at all since I don't think I have permissions to generate that and it was too time-consuming to worry about getting them both to work (I've already probably spent more time on this than I should have). My hope here is that we fold this in (I think the website is more important, actually), and then try to rebuild Sphinx on top of that, afterward. It shouldn't be too difficult, I wouldn't think.

docs/training.md
```diff
@@ -9,6 +9,39 @@ layout: default
 For easily training popular model types on known data sets, see the [Sockeye Autopilot documentation](https://github.com/awslabs/sockeye/tree/master/sockeye_contrib/autopilot).
 For manually training and running translation models on your data, read on.
 
+## Data preparation
+
+Sockeye can read the raw data at training time in two sentence-parallel files via the `--source` and `--target` command-line options.
+You can also prepare the data ahead of time and dump it to disk as MXNet NDArrays.
+This basically eliminates the data loading time when running training (since three passes over the raw data are required), and also reduces memory consumption, since prepared data is also placed into random shards (which have one million lines each, by default).
```
fhieber: I think its 10 million lines each per shard by default :)

mjpost: [Looks like 1m](https://github.com/awslabs/sockeye/blob/5a50d960b6e20527a006b176419b188d79a85a85/sockeye/arguments.py#L438). This value should probably be pulled out as a constant.

fhieber: you are right :)

mjpost:
Sounds good, thanks! Consider also, though, that there is nothing new here at all: this PR just pulls together documentation that was already there but in many different places. I think the only thing I added was the docs for source factors (an existing feature).

fhieber:
I think it be useful to have some kind of link on the website that points to the code/github.

tdomhan:
Thanks for putting together the website change! Regarding sphinx: we did indeed discuss to have just a single place for documentation, namely the website. In general it definitely makes total sense to avoid having the same information available redundantly in multiple locations. As the github page will no longer contain the docs generated from modules we could consider hosting module docs on readthedocs and everything else on the github page. This way we wouldn't have redundant information. That said, it's unclear how useful the module docs are. So maybe we just go ahead with the page change as is and revisit the sphinx documentation later.
It's definitely great that we will finally have a website :)

docs/setup.md
```diff
@@ -0,0 +1,90 @@
...
+
+If you want to run sockeye on a GPU you need to make sure your version of Apache MXNet
+Incubating contains the GPU bindings. Depending on your version of CUDA you can do this by
+running the following:
+
+```bash
+> pip install -r requirements/requirements.gpu-cu${CUDA_VERSION}.txt
+> pip install .
+```
+where `${CUDA_VERSION}` can be `75` (7.5), `80` (8.0), `90` (9.0), or `91` (9.1).
```
tdomhan: 9.2 seems to be missing here.

docs/tutorials/seqcopy.md
```diff
@@ -1,4 +1,5 @@
 # Sequence copy model
+
```
tdomhan: The sequence copy tutorial previously assumed you were in the tutorials folder where the script lies. Now that we're hosting the tutorial separately, this assumption no longer makes sense sense. Would it make sense to add something like
```
wget https://raw.githubusercontent.com/awslabs/sockeye/master/tutorials/seqcopy/genseqcopy.py
```

to the start of the tutorial itself?

docs/tutorials/seqcopy.md
```diff
@@ -1,4 +1,5 @@
...
 This tutorial will show you the basic usage of Sockeye on a on a simple task: copying a sequence.
 We will generate sequences consisting of digits of variable lengths.
 The task is then to train a model that copies the sequence from the source to the target.
```
tdomhan: There are two links below that I believe need to be adjusted to the website. Namely:
* `For this tutorial we assume that you have successfully [installed](../../README.md#installation) Sockeye.`
* `his means that even while the model is still training you can use the model folder for translation, as described in the [next section](#3-translation).`

docs/image_captioning.md
```diff
@@ -1,3 +1,6 @@
+---
+layout: default
+---
```
tdomhan: the link to lexical constraints needs to be adjusted to point to the Inference subpage:
* `[decoding with lexical constraints](../constraints/README.md).`

docs/_layouts/default.html
```diff
@@ -0,0 +1,64 @@
...
+          <ul>
+            <li><a href="{{ site.baseurl }}/index.html">Home</a>
+            <li><a href="{{ site.baseurl }}/setup.html">Installation</a>
+            <li><a href="{{ site.baseurl }}/training.html">Training</a>
+            <li><a href="{{ site.baseurl }}/inference.html">Inference</a>
+            <li><a href="{{ site.baseurl }}/scoring.html">Scoring</a>
+            <li><a href="{{ site.baseurl }}/development.html">Development</a>
+            <li><a href="{{ site.baseurl }}/tutorials.html">Tutorials</a>
+            <li><a href="{{ site.baseurl }}/image_captioning.html">Image Captioning</a>
+            <li><a href="{{ site.baseurl }}/faq.html">FAQ</a>
```
tdomhan: Would it make sense to remove the FAQ entirely? Initially I thought we'd fill it with questions/answers over time. As it is currently, it is not very useful to users. I think the error message in the FAQ is even out-dated and would not be thrown anymore.

docs/development.md
```diff
@@ -44,45 +40,52 @@ If you want to develop sockeye, please adhere to the following development guide
...
- * When importing other sockeye modules import the entire module instead of individual functions and classes using
+- When importing other sockeye modules import the entire module instead of individual functions and classes using
    relative imports:
     ```python
     from . import attention
     ```
 
 
 ## Building the Documentation
+
```
tdomhan: Let's remove this section, as you removed Sphinx.

docs/inference.md
```diff
@@ -0,0 +1,147 @@
...
+
+This will output tab-delimited pairs of (score, translation).
+As always, don't forget to apply source- and target-side preprocessing to your input and your constraint.
+
+## CPU process per core translation
+
+On multi-core computers, translation per core separately can speedup translation performance, due to some operation can't be handled parallel in one process.
+Using this method, translation on each core can be parallel.
+
+One [python script example](tutorials/cpu_process_per_core_translation.py] is given and you can run it as follows:
```
tdomhan: there's a typo in the link (final parentheses). Also: would it actually work to point to a file in the repo this way?

tdomhan:
Looks great. I think a couple of links wouldn't work anymore. Let's fix those and enable the website :)

docs/_layouts/default.html
mjpost: sure

docs/inference.md
mjpost: Fixed. Does this even work?

tdomhan: no, I think it was just displayed as plain text.

mjpost: I mean, does the multi-core CPU processing still work?

tdomhan: ahh :) probably, but not sure. I would have probably left this in the tutorials and not put it under the inference page, as it might create more confusion to users.

mjpost:
Excellent! Now we just need one of you to flip the docs switch, right? Then [the website](https://aws.github.io/sockeye) will be live and we can start advertising.

tdomhan:
perfect. I'll merge now and then pull the switch :)

mjpost:
I am page-reloading with bated breath.

docs/inference.md
```diff
@@ -0,0 +1,149 @@
...
+Given a set of constraints, the decoder will find the best output that includes the constraints.
+This file describes how to use lexical constraints; for more technical information, please see our paper:
+
+> Matt Post and David Vilar. 2018.
+> [Fast Lexically Constrained Decoding with Dynamic Beam Allocation for Neural Machine Translation](http://aclweb.org/anthology/N18-1119).
+> Proceedings of the 2018 Conference of the North American Chapter of the Association for Computational Linguistics: Human Language Technologies, Volume 1 (Long Papers).
+
+### Example
+
+You need a [trained model](../wmt/README.md).
```
tdomhan: actually here's one more link that's not working :P

docs/tutorials/wmt.md
```diff
@@ -142,38 +139,31 @@ In the next section we discuss how you can monitor the training progress.
...
 
 ```bash
 tensorboard --logdir .
 ```
 
-Once tensorboard is up and running you can check out the learning curves by
-opening [http://localhost:6006](http://localhost:6006).
+Once tensorboard is up and running you can check out the learning curves by opening [http://localhost:6006](http://localhost:6006).
 
 ![screenshot of tensorboard](tb_screenshot.png "Screenshot of tensorboard")
```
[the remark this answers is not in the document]
tdomhan: I think we need the `{{ site.baseurl }}` here.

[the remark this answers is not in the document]
tdomhan: something like:
```
![screenshot of tensorboard]({{ site.baseurl}}/tutorials/wmt/tb_screenshot.png "Screenshot of tensorboard")
```
(not sure this works)

docs/tutorials/wmt.md
```diff
@@ -208,16 +197,14 @@ This will create a file `align_1.png` that looks similar to this:
 
 ![Alignment plot](align.png "Alignment plot")
```
tdomhan: same here.

tdomhan:
I enabled github pages. Somehow it doesn't show up at https://awslabs.github.io/sockeye/ yet. Let's see if its just delayed.

mjdenkowski:
Looks like it's live now!

tdomhan:
great, thanks for putting this together, Matt!

Issue #546, opened 2018-09-26
Bug with beam-size=1?


mjpost:
In trying to get tests to passing with scoring (#538), I have turned up some weird behavior with scores output by Sockeye. Here are two commands using a transformer model built in the system tests. Notice:

- The invocations differ only in the beam size (1 or 2)
- `--skip-topk` is not enabled
- With beam size of 1, the scores output should be impossible, since Sockeye outputs negative logprobs.

Any ideas? 

CC: @bricksdont

```bash
$ python3 -m sockeye.translate -i src --output-type translation_with_score --use-cpu -m model --beam-size 1 2> /dev/null | head
-10.556	7 5 2 7 3 6 5 4 7 7
-10.727	9 2 4 1 6 7 8 6 8
-12.788	8 6 8 7
-10.413	0 5 0 7 5 9 0 6 3 1
-10.731	7 9 2 6 8 5 0 6 5
-12.490	5 6 3 2
-inf	
-11.242	3 9 1 3 8 7
-15.759	2 1
-10.506	8 8 8 2 4 4 5 5 2 5
$ python3 -m sockeye.translate -i src --output-type translation_with_score --use-cpu -m model --beam-size 2 2> /dev/null | head
0.003	7 5 2 7 3 6 5 4 7 7
0.001	9 2 4 1 6 7 8 6 8
0.000	8 6 8 7
0.002	0 5 0 7 5 9 0 6 3 1
0.001	7 9 2 6 8 5 0 6 5
0.001	5 6 3 2
-inf	
0.001	3 9 1 3 8 7
0.001	2 1
0.002	8 8 8 2 4 4 5 5 2 5
```

mjpost:
I will be digging into this, but thought I'd post an issue in case it looks familiar. Here are the training params (on a digits data set with 10,000 training points):

    train_params = '--encoder transformer --decoder transformer --max-updates 4000 --num-layers 2 --transformer-attention-heads 4 --trans...l-learning-rate 0.001 --decode-and-evaluate 0 --label-smoothing 0.0 --optimized-metric perplexity --loss cross-entropy', 
    translate_params = '--beam-size 1', 
    translate_params_equiv = None, 
    train_source_path = '/var/folders/mw/ks6b_m_x3916ybg0prddtg3r0000gn/T/test_seq_copy.9obml2kp/train.src'

bricksdont:
[...]

PR #550, opened 2018-09-28
Only skip softmax if scores are not output anyway

bricksdont:
Fixing the behaviour of single-model, greedy decoding - to work around the issue that those scores are not passed through softmax by default:

https://github.com/awslabs/sockeye/issues/546

until you can check this box.
- [] Updated major/minor version in `sockeye/__init__.py`. Major version bump if this is a backwards incompatible change.
- [ ] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -392,6 +393,7 @@ def load_models(context: mx.context.Context,
                                    restrict lexicon).
     :param forced_max_output_len: An optional overwrite of the maximum output length.
     :param override_dtype: Overrides dtype of encoder and decoder defined at training time to a different one.
+    :param output_scores: Whether the scores will be needed as outputs.
```
fhieber: can we make the parameter and its description more descriptive? I.e. what do we mean by scores exactly? Normalized negative log probabilities.

sockeye/translate.py
```diff
@@ -62,6 +62,8 @@ def run_translate(args: argparse.Namespace):
                                         args.output,
                                         args.sure_align_threshold)
 
+    output_scores = True if args.output_type in C.OUTPUT_HANDLERS_SCORING else False
```
fhieber: `args.output_type in C.OUTPUT_HANDLERS_SCORING` ?

fhieber:
I guess this change could simplify some of the involved test logic that @mjpost introduced with the scoring PR. Can we simplify it with this PR?

sockeye/constants.py
```diff
@@ -334,7 +334,9 @@
                    OUTPUT_HANDLER_ALIGN_TEXT,
                    OUTPUT_HANDLER_BEAM_STORE]
 OUTPUT_HANDLERS_SCORING = [OUTPUT_HANDLER_SCORE,
-                           OUTPUT_HANDLER_PAIR_WITH_SCORE]
+                           OUTPUT_HANDLER_PAIR_WITH_SCORE,
+                           OUTPUT_HANDLER_TRANSLATION_WITH_SCORE,
+                           OUTPUT_HANDLER_TRANSLATION_WITH_ALIGNMENT_MATRIX]
```
mjpost: This is incorrect—the purpose of `OUTPUT_HANDLERS_SCORING` is to list valid values of `--output-type` for `sockeye.score`.

sockeye/translate.py
mjpost: What about just checking if the string "score" is in the output type? And then add a comment in `constants.py` noting that softmax is disabled for anything containing that keyword. i.e., 

    output_scores = C.OUTPUT_SCORE_KEYWORD in args.output_type?

Then in `constants.py`:

    # If this token is a substring of the output type, it disables the disabling of softmax in inference.py
    OUTPUT_SCORE_KEYWORD = "score"

tdomhan: Checking for a string sounds a bit hacky. Keeping a separate list seems error prone for new output handlers that will be added in the future. How about making it explicit by extending the OutputHandler interface:

Something like:
```
class OutputHandler(ABC):
    """
    Abstract output handler interface
    """

    @abstractmethod
    def handle(self,
               t_input: inference.TranslatorInput,
               t_output: inference.TranslatorOutput,
               t_walltime: float = 0.):
        """
        :param t_input: Translator input.
        :param t_output: Translator output.
        :param t_walltime: Total wall-clock time for translation.
        """
        pass

    @abstractmethod
    def reports_score(self) -> bool:
        pass
```

mjpost: Yes this is much better.

fhieber:
I addressed the remaining comments in continuation PR #582.

PR #555, opened 2018-10-02
Correct source factor input files check.

tdomhan:

### DIFF
--- a/sockeye/translate.py
+++ b/sockeye/translate.py
@@ -144,9 +144,10 @@ def make_inputs(input_file: Optional[str],
     else:
         input_factors = [] if input_factors is None else input_factors
         inputs = [input_file] + input_factors
-        check_condition(translator.num_source_factors == len(inputs),
-                        "Model(s) require %d factors, but %d given (through --input and --input-factors)." % (
-                            translator.num_source_factors, len(inputs)))
+        if not input_is_json:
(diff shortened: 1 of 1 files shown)


sockeye/translate.py
```diff
@@ -144,9 +144,10 @@ def make_inputs(input_file: Optional[str],
     else:
         input_factors = [] if input_factors is None else input_factors
         inputs = [input_file] + input_factors
-        check_condition(translator.num_source_factors == len(inputs),
-                        "Model(s) require %d factors, but %d given (through --input and --input-factors)." % (
-                            translator.num_source_factors, len(inputs)))
+        if input_is_json:
```
fhieber: `if not input_is_json`, no?

fhieber: I would suggest making the `input_is_json` if statement the top-level block in this method:
```python
if input_is_json:
  # expect only a single stream, inputs is []
  yield inference.make_input_from_json_string(sentence_id=sentence_id, json_string=input)
else:
  # check for multiple streams
  yield inference.make_input_from_multiple_strings(sentence_id=sentence_id, strings=list(inputs))

tdomhan: ah yeah right. changed.

The check is only needed once, that's why I didn't move it in. Alternatively we would have to repeat the for loop.

PR #571, opened 2018-11-01
Ngram block

ballerlina:
Add another argument to the inference model specifying a value $n$ which is the minimum length n-gram to block from the output sentences. Hypotheses that generated repeating n-grams will be removed from the beam unless they are the only ones remaining, in which case they are left as-is (for now).


ballerlina:
Sorry! Wrong project, please ignore :)

PR #575, opened 2018-11-08
Fix the seqcopy tutorial URL

bittlingmayer:
Right now it's a 404.


tdomhan:
Thanks for noticing and fixing!

PR #581, opened 2018-11-26
Don't do state cleanup if training not converged and return trainer's state

artemsok:
By default the ```training_state``` folder is deleted even if the training did not converge, i.e. the number of checkpoints did not hit ```--max-num-not-improved```, but the training stopped because of reaching ```--max-samples```, ```--max-num-epochs```, or ```--max-updates```. This changes the behaviour to keeping the folder to allow later training continuation and not enforcing the above parameters to match with the old run. Additionally a full training state is returned (with metrics, number of epochs, convergence status etc.), allowing its processing in a calling code.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
 ## [1.18.59]
+### Added
 - A flag is returned alongside the best metric indicating if training converged.
 ### Changed
 - Training state cleanup will not be performed for training runs that did not converge.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 
 ## [1.18.59]
 ### Added
-- A flag is returned alongside the best metric indicating if training converged.
+- Full training state is now returned from EarlyStoppingTrainer's fit().
 ### Changed
 - Training state cleanup will not be performed for training runs that did not converge.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ### Added
 - Full training state is now returned from EarlyStoppingTrainer's fit().
 ### Changed
-- Training state cleanup will not be performed for training runs that did not converge.
+- Training state cleanup will not be performed for training runs that did not converge yet.
 
 ## [1.18.58]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.59]
+- A flag is returned alongside the best metric indicating if training converged.
+### Changed
+- Training state cleanup will not be performed for training runs that did not converge.
+
 ## [1.18.58]
 ### Added
 - Added nbest translation, exposed as `--nbest-size`. Nbest translation means to not only output the most probable translation according to a model, but the top n most probable hypotheses. If `--nbest-size > 1` and the option `--output-type` is not explicitly specified, the output type will be changed to one JSON list of nbest translations per line. `--nbest-size` can never be larger than `--beam-size`.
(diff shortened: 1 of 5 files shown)


sockeye/constants.py
```diff
@@ -238,7 +238,8 @@
 # Arguments that may differ and still resume training
 ARGS_MAY_DIFFER = ["overwrite_output", "use-tensorboard", "quiet",
                    "align_plot_prefix", "sure_align_threshold",
-                   "keep_last_params"]
+                   "keep_last_params", "seed", "max-updates",
+                   "max-num-epochs", "max-samples"]
```
fhieber: How about `--min-num-*`? In general, would it make sense to make a pass over current arguments to see whether we should add more to this list? Its a suboptimal solution as it becomes outdated quickly, but if you are at it anyway right now :)

sockeye/training.py
```diff
@@ -761,6 +767,9 @@ def _cleanup(self, lr_decay_opt_states_reset: str, process_manager: Optional['De
                 self.tflogger.log_metrics(decoder_metrics, decoded_checkpoint)
             utils.write_metrics_file(self.state.metrics, self.metrics_fname)
 
+        if keep_training_state:
```
fhieber: matter of taste, but I would prefer to move the following lines under an `if not keep_training_state` instead of the early return.

sockeye/training.py
```diff
@@ -523,6 +523,8 @@ def fit(self,
         speedometer = Speedometer(frequency=C.MEASURE_SPEED_EVERY, auto_reset=False)
         tic = time.time()
 
+        converged = False
```
fhieber: not really important, but would it make sense to move this boolean into `TrainState` just for clarity?
In that case, this function could even return the TrainState object, which holds all kinds of information from training (best metric, updates, epochs etc.)

sockeye/train.py
```diff
@@ -863,25 +863,25 @@ def train(args: argparse.Namespace):
...
-                    min_epochs=min_epochs,
-                    max_epochs=max_epochs,
-                    lr_decay_param_reset=args.learning_rate_decay_param_reset,
-                    lr_decay_opt_states_reset=args.learning_rate_decay_optimizer_states_reset,
-                    decoder=create_checkpoint_decoder(args, exit_stack, context),
-                    mxmonitor_pattern=args.monitor_pattern,
-                    mxmonitor_stat_func=args.monitor_stat_func,
-                    allow_missing_parameters=args.allow_missing_params or model_config.lhuc,
-                    existing_parameters=args.params)
+        return trainer.fit(train_iter=train_iter,
```
fhieber: [nit] I'd prefer a more explicit style here, i.e.
```python
best_metric, converged = trainer.fit(...)
return best_metric, converged
```

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.59]
+- A flag is returned alongside the best metric indicating if training converged.
```
fhieber: I'd put this under an `### Added section`.

artemsok: Oh, it was eaten up by a merge, returned.

sockeye/train.py
artemsok: Done

sockeye/training.py
artemsok: Makes sense, actually this may simplify calling training from some code.

sockeye/training.py
artemsok: fixed

sockeye/constants.py
artemsok: Agree, ```--min-*``` shouldn't matter either, added them.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.59]
+### Added
+- Full training state is now returned from EarlyStoppingTrainer's fit().
+### Changed
+- Training state cleanup will not be performed for training runs that did not converge.
```
tdomhan: Maybe add a 'yet': 'that did not converge **yet**'

sockeye/training.py
```diff
@@ -455,7 +456,7 @@ def fit(self,
             mxmonitor_pattern: Optional[str] = None,
             mxmonitor_stat_func: Optional[str] = None,
             allow_missing_parameters: bool = False,
-            existing_parameters: Optional[str] = None):
+            existing_parameters: Optional[str] = None) -> 'TrainState':
```
tdomhan: No need to forward reference as the TrainState is defined above, no?
```suggestion
            existing_parameters: Optional[str] = None) -> TrainState:
```

CHANGELOG.md
artemsok: fixed

sockeye/training.py
artemsok: thanks, indeed

PR #582, opened 2018-11-26
Skip softmax (continuation)

fhieber:
continuation of #550 as this was inactive for a while, but its an important feature to enable consistent scores for various use cases in Sockeye.

Like for the nbest translation feature, this is mostly work done by @bricksdont and merging should either not squash commits, or squash and preserve authorship of @bricksdont.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -78,12 +78,13 @@ class InferenceModel(model.SockeyeModel):
                               'The beam size must be smaller than the target vocabulary size.')
         if skip_softmax:
             assert beam_size == 1, 'Skipping softmax does not have any effect for beam size > 1'
+        self.skip_softmax = skip_softmax
+
         self.batch_size = batch_size
         self.softmax_temperature = softmax_temperature
         self.max_input_length, self.get_max_output_length = models_max_input_output_length([self],
                                                                                            max_output_length_num_stds,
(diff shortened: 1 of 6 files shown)


sockeye/inference.py
```diff
@@ -1144,10 +1147,10 @@ def __init__(self,
         utils.check_condition(self.beam_size >= nbest_size,
                               'Nbest size must be smaller or equal to beam size.')
         self.batch_size = self.models[0].batch_size
-        # skip softmax for a single model, but not for an ensemble
-        self.skip_softmax = self.models[0].skip_softmax
-        if self.skip_softmax:
-            utils.check_condition(len(self.models) == 1 and self.beam_size == 1, "Skipping softmax cannot be enabled for several models, or a beam size > 1.")
+
+        if self.models[0].skip_softmax:
```
tdomhan: That check doesn't seem to be complete. If there is more than one model we now would need to make sure that none of the models have `skip_softmax` set, right? (The check was already like this before the PR, but I'm just wondering).

fhieber: I think this requirement is implicitly given by the checks in `load_models()`, but I can extend the check here.

tdomhan:
looks good, thanks!

PR #586, opened 2018-11-27
sampling

mjpost:
This adds sampling to Sockeye (via `--sample [N]`), a collaborative effort with @edwdh. It will create `beam_size` samples (and return as many as `nbest_size`). If an integer is passed to `--sample N`, samples at each time step for each hypothesis will only be taken for the top `N` most-probable target-language tokens. The default is 0, which samples from the entire target vocabulary.

This commit also changes the way `inactive` works, removing its dual purpose of handling the special case at timestep 1 when we need to run `topk()` only over items from the first row in each batch. This is now reverted to the old behavior, where we have a special conditional for `t==1`. This simplifies some of the code and removes some confusing logic with `inactive`.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -1690,37 +1690,51 @@ def _beam_search(self,
...
+            if self.sample:
+                best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores, target_dists, finished)
+            else:
+                # On the first timestep, all hypotheses have identical histories, so force topk() to choose extensions
+                # of the first row only
+                if t == 1 and not self.skip_topk:
+                    batch_indices = mx.nd.array(np.arange(0, self.batch_size * self.beam_size, self.beam_size), dtype='int32', ctx=self.context)
+                    best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores[batch_indices,:])
+                else:
+                    best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores)
```
mjpost: This restores the t==1 special case. We pass only the first row of `scores` (for all items in the batch).

sockeye/constants.py
```diff
@@ -28,6 +29,9 @@
 # reserve extra space for the EOS or BOS symbol that is added to both source and target
 SPACE_FOR_XOS = 1
 
+# Defined once here so the value used can be retrieved elsewhere (e.g., for unit tests)
+DEFAULT_RANDOM_SEED = int(time.time())
```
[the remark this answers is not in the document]
tdomhan: mhh we should not start relying on a changing random seed for the tests. What's the behavior you want to achieve? I assume the goal is that every time you call `translate.py` you get a different set of samples, unless someone specifies a seed on the CLI. I'd suggest the following:
1. remove `DEFAULT_RANDOM_SEED `
2. Make the `args.seed` default to `None` for `translate.py`
3. Do the following in `translate.py`:
```
if args.seed is None:
    seed = int(time.time())
else:
    seed = args.seed
utils.seed_rngs(args.seed)
```
This way we don't need an explanation as to why we have something changing in the constants and we are explicit about the behavior, namely that we set the seed with the time if it is not set manually.

What do you think?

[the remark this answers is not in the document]
mjpost: Yes, your way is better, thanks.

sockeye/inference.py
fhieber: For brevity, could we do 
```python
if t == 1 and not self.skip_topk:
  scores = scores[self.batch_indices, :]
best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores)
```
, or does this interfere with line 1746 (passing unmodified `scores` into `constrained.topk`? If yes, can we change the constraint code to work with this simplification?

sockeye/inference.py
```diff
@@ -2026,16 +2036,57 @@ def hybrid_forward(self, F, scores, offset):
...
-        folded_scores = F.reshape(scores, shape=(self.batch_size, self.k * self.vocab_size))
+        folded_scores = F.reshape(scores, shape=(self.batch_size, -1))
+
         values, indices = F.topk(folded_scores, axis=1, k=self.k, ret_typ='both', is_ascend=True)
+
         indices = F.reshape(F.cast(indices, 'int32'), shape=(-1,))
-        unraveled = F.unravel_index(indices, shape=(self.batch_size * self.k, self.vocab_size))
+        unraveled = F.unravel_index(indices, shape=(-1, self.vocab_size))
+
         best_hyp_indices, best_word_indices = F.split(unraveled, axis=0, num_outputs=2, squeeze_axis=True)
```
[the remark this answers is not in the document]
fhieber: This is awesome, as we can remove `self.batch_size` and `self.vocab_size` again from this block right?
I wonder if we can even remove the use of `batch_size` for `self.offset` in the constructor somehow. Is the newly added `batch_indices` array in the calling code not equal to `self.offset`?

In general, having to pass in a constant shape parameter (batch size) into these Blocks is a nuisance that we have to live with so far. If we ever want to add dynamic batch sizes for translation, this would need to be changed anyway. Ideally, we would create these auxiliary masks/arrays based on the input to the block dynamically, but if that is too much overhead, passing them in as inputs and create them once outside the block is the better option.

sockeye/lexical_constraints.py
```diff
@@ -483,7 +483,8 @@ def __str__(self):
         return '({}, {}, {}, {})'.format(self.row, self.col, self.score, self.hypothesis.num_met())
 
 
-def topk(batch_size: int,
+def topk(timestep: int,
```
fhieber: missing docstring

sockeye/lexical_constraints.py
```diff
@@ -533,7 +535,8 @@ def topk(batch_size: int,
     return best_ids, best_word_ids, seq_scores, hypotheses, inactive
 
 
-def _topk(beam_size: int,
+def _topk(timestep,
```
fhieber: missing docstring and type annotation for `timestep`

sockeye/translate.py
```diff
@@ -87,6 +91,12 @@ def run_translate(args: argparse.Namespace):
                            'This feature may be removed or change its behaviour in future. '
                            'DO NOT USE IT IN PRODUCTION!')
 
+        # Skip softmax for a single model when not doing ensembling or beam search or sampling
+        skip_softmax = False
```
fhieber: please rebase onto latest master with skip_softmax changes added

sockeye/inference.py
```diff
@@ -2026,16 +2041,72 @@ def hybrid_forward(self, F, scores, offset):
...
         best_hyp_indices = best_hyp_indices + offset
         values = F.reshape(values, shape=(-1, 1))
         return best_hyp_indices, best_word_indices, values
 
 
+class SampleK(mx.gluon.HybridBlock):
```
fhieber: Could you add a unit test for this block in `test_inference.py`?

sockeye/lexical_constraints.py
```diff
@@ -575,7 +578,7 @@ def _topk(beam_size: int,
     # (3) the best item (constrained or not) in that row
     best_next = mx.nd.NDArray.argmin(scores, axis=1)
```
fhieber: btw, I think this can be simplified to `mx.nd.argmin(...)`

sockeye/translate.py
mjpost: Done. I changed the function to pass `skip_softmax` directly, since it now needs to check both `not args.sample and not output_handler.reports_score()` (and passing just one variable seems better).

sockeye/inference.py
```diff
@@ -1696,37 +1699,52 @@ def _beam_search(self,
...
+                    if self.sample is not None:
+                        target_dists[block_indices] = np.inf
+
+            # finished_on_prev_step = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]) + (best_word_indices == C.PAD_ID)
 
             # (3) Get beam_size winning hypotheses for each sentence block separately. Only look as
             # far as the active beam size for each sentence.
-            best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores)
+
+            # When sampling, manually set the chosen word to C.PAD_ID for finished hypotheses
```
tdomhan: what code does this comment refer to?

sockeye/inference.py
```diff
@@ -1696,37 +1699,52 @@ def _beam_search(self,
...
 
             # Mark entries that should be blocked as having a score of np.inf
             if self.global_avoid_trie or any(raw_avoid_list):
                 block_indices = avoid_states.avoid()
                 if len(block_indices) > 0:
                     scores[block_indices] = np.inf
+                    if self.sample is not None:
+                        target_dists[block_indices] = np.inf
+
+            # finished_on_prev_step = (best_word_indices == self.vocab_target[C.EOS_SYMBOL]) + (best_word_indices == C.PAD_ID)
```
tdomhan: dead code?

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: I guess you are right, I didn't see the use of `self.batch_size` when reshaping into `folded_scores`. Afaik, MXNet doesn't really have an op (yet?) that returns the shape of its input, which can then be used as an argument to another op (most shape arguments of operators are not symbols but int tuples).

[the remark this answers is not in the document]
mjpost: Yeah, I wish you could write something like:

    unraveled = F.unravel_index(indices, shape=scores)

sockeye/inference.py
mjpost: Stray comment.

mjpost:
FYI, werde ich in einer Woche darauf zurückkommen.

sockeye/inference.py
```diff
@@ -1177,26 +1176,31 @@ def __init__(self,
...
                 self._top = TopK(k=self.beam_size,
                                  batch_size=self.batch_size,
                                  vocab_size=len(self.vocab_target))  # type: mx.gluon.HybridBlock
-                self._top.initialize(ctx=self.context)
-                self._top.hybridize(static_alloc=True, static_shape=True)
+
+
+            self._top.initialize(ctx=self.context)
+            self._top.hybridize(static_alloc=True, static_shape=True)
+        else:
```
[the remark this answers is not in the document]
mjpost: Done.

test/unit/test_inference.py
```diff
@@ -476,6 +476,42 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
     assert all(mx_values == np_values)
 
 
+@pytest.mark.parametrize("batch_size, beam_size, target_vocab_size, top_n",
+                        [(1, 5, 200, 0),
+                         (5, 5, 200, 0),
+                         (1, 100, 200, 5),
+                         (5, 100, 200, 5)])
+def test_samplek_func(batch_size, beam_size, target_vocab_size, top_n):
```
fhieber: Thanks for adding this!

test/unit/test_inference.py
```diff
@@ -476,6 +476,42 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
...
+    target_dists = mx.nd.broadcast_div(scores, mx.nd.sum(scores, axis=1, keepdims=True))
+
+    samplek = sockeye.inference.SampleK(k=beam_size, n=top_n, batch_size=batch_size, context=mx.context.cpu())
+    samplek.initialize()
+
+    # 0..(batch_size * beam_size)-1
+    expected_hyps = mx.nd.array(range(batch_size * beam_size), dtype='int32')
+    finished = mx.nd.array(mx.nd.random.uniform(0, 1, (batch_size * beam_size)) > 0.5, dtype='int32')
+
+    for i in [1, 2]:
```
fhieber: These are independent iterations right? Note that when you hybridize a block i staya hybridized, so if you want to test hybridized block and non-hybridized then I would only hybridize the second iteration.

test/unit/test_inference.py
```diff
@@ -476,6 +476,42 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
...
+@pytest.mark.parametrize("batch_size, beam_size, target_vocab_size, top_n",
+                        [(1, 5, 200, 0),
+                         (5, 5, 200, 0),
+                         (1, 100, 200, 5),
+                         (5, 100, 200, 5)])
+def test_samplek_func(batch_size, beam_size, target_vocab_size, top_n):
+    # arrange scores increasing values from left to right, so the best item is always index 0, next-best 1, and so on
+    scores = mx.nd.array([list(range(1, target_vocab_size + 1)) for x in range(batch_size * beam_size)])
+    # normalize
+    target_dists = mx.nd.broadcast_div(scores, mx.nd.sum(scores, axis=1, keepdims=True))
```
fhieber: [nit] could use the fluent method for sum in this function for brevity: x.sum(axis=1) etc

test/unit/test_inference.py
```diff
@@ -476,6 +476,42 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
...
+
+    samplek = sockeye.inference.SampleK(k=beam_size, n=top_n, batch_size=batch_size, context=mx.context.cpu())
+    samplek.initialize()
+
+    # 0..(batch_size * beam_size)-1
+    expected_hyps = mx.nd.array(range(batch_size * beam_size), dtype='int32')
+    finished = mx.nd.array(mx.nd.random.uniform(0, 1, (batch_size * beam_size)) > 0.5, dtype='int32')
```
fhieber: Is mx.nd.array required here? If uniform accepts dtype I dont think so.

test/unit/test_inference.py
```diff
@@ -476,6 +476,42 @@ def test_topk_func(batch_size, beam_size, target_vocab_size):
...
+        hyps, words, values = samplek(scores, scores, finished)
+        assert hyps.shape[0] == batch_size * beam_size
+        # The indices should always be the integers from 0 to batch*beam-1
+        assert sum(hyps == expected_hyps).asscalar() == (batch_size * beam_size)
+        if top_n != 0:
+            # Scores are increasing left-to-right, so best items are all the lowest word IDs.
+            # No word id greater than the cap (top_n) should be selected
+            assert mx.nd.sum(words >= top_n)[0].asscalar() == 0
+        # word index should be zero for all finished hypotheses
+        print("finished", finished)
```
fhieber: Leftover prints

sockeye/utils.py
```diff
@@ -280,22 +280,26 @@ def topk(scores: mx.nd.NDArray,
...
     :param use_mxnet_topk: True to use the mxnet implementation or False to use the numpy one.
     :return: The row indices, column indices and values of the k smallest items in matrix.
     """
+    # Compute the batch size from the offsets. We don't know the batch size because it is
+    # either 1 (timestep 1) or k (timestep 2+).
     # (batch_size, beam_size * target_vocab_size)
-    folded_scores = scores.reshape((-1, k * scores.shape[-1]))
-    batch_size = folded_scores.shape[0]
+    batch_size = int(offset[-1].asscalar() / k) + 1
+    folded_scores = scores.reshape((batch_size, -1))
```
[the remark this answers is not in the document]
fhieber: I wonder if we can avoid the `asscalar()` conversion here, since we know that offset is always batch*beam:
```python
beam_size = int(offset.shape[0] / k)
scores = scores.reshape(-1, beam_size, 0) # maybe we need reshape(-4, -1, beam_size, 0). resulting shape: (batch, beam, target_vocab_size)
scores = scores.reshape(0, -3) # shape: (batch, beam*target_vocab_size)
```

test/unit/test_inference.py
mjpost: Argh—I had `range(2)` and then made the loop values explicit without updating the conditional. Thanks for the catch!

sockeye/inference.py
```diff
@@ -395,9 +395,8 @@ def load_models(context: mx.context.Context,
                                    restrict lexicon).
     :param forced_max_output_len: An optional overwrite of the maximum output length.
     :param override_dtype: Overrides dtype of encoder and decoder defined at training time to a different one.
-    :param output_scores: Whether the scores will be needed as outputs. If True, scores will be normalized, negative
-           log probabilities. If False, scores will be negative, raw logit activations if decoding with beam size 1
-           and a single model.
+    :param skip_softmax: If False, scores will be returned by the decoder step as normalized negative log probabilities.
```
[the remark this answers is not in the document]
mjpost: No, I had to update it. Softmax is now skipped if scores are requested or if sampling is turned on. Since there were two variables, it was more efficient to make the decisions outside the loop and then pass that in.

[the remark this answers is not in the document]
fhieber: Ah got it, thanks! However, if we think of `load_models()` as an API/library call it may be better to move the logic into there instead of relying on cmdline args. Sampling is kind of similar to the existing variables of this function (skip output layer for vocab restriction etc.) Not sure though, so your call.
In any case, I liked the name `output_scores` (and its docstring mentioning the conditions when this can be active) more than `skip_softmax`, which is easy to confuse with `skip_topk`.

sockeye/inference.py
```diff
@@ -1716,37 +1721,49 @@ def _beam_search(self,
...
             # far as the active beam size for each sentence.
-            best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores)
+
+            if self.sample is not None:
+                best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores, target_dists, finished)
+            else:
+                # On the first timestep, all hypotheses have identical histories, so force topk() to choose extensions
+                # of the first row only
+                if t == 1 and not self.skip_topk:
+                    scores = scores[self.batch_indices, :]
```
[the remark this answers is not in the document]
mjpost: seconds / sentence for a WMT18 test set with a German–English model, single run but on the same GPU, one after the other:

beam | batch | master | sampling
------ | ------|--------|----------
5 | 1 | .221 | .219
5 | 10 | .042 | 0.42
10 | 1 | .234 | .235
10 | 10 | .067 | .067
20 | 1 | .265 | .263

Looks fine.

sockeye/utils.py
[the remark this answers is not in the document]
mjpost: Oh, sorry, missed this one. I’ll take a look later.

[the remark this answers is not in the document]
mjpost: Actually, I don't understand this. Why do we need to avoid `asscalar()`? Also, your `beam_size` variable above is actually the batch size.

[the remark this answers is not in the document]
mjpost: Okay, I think I fixed this after swapping it back in.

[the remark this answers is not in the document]
fhieber: Thanks! You are right, I was confusing batch and beam here. The reason, I would like to avoid `asscalars` is the same as for `asnumpys`: they cause a synchronization between mxnets backend and python/numpy frontend and can potentially make things slower. The shape information of ndarrays is available instantaneously. Probably not too important here, but why take the 'risk'? :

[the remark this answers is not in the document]
mjpost: Yeah, even after noting that typo of yours, I still managed to fix what you wanted :)

PR #590, opened 2018-11-29
Informative error message when --max-seq-len is too small.

tdomhan:
Just adding a small check to fail with an informative error message if none of the sentence fits, for example because `--max-seq-len` was set incorrectly.

### DIFF
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -780,7 +780,8 @@ def get_training_data_iters(sources: List[str],
                             max_seq_len_target: int,
                             bucketing: bool,
                             bucket_width: int,
-                            permute: bool = True) -> Tuple['BaseParallelSampleIter',
+                            permute: bool = True,
+                            allow_empty: bool = False) -> Tuple['BaseParallelSampleIter',
                                                            Optional['BaseParallelSampleIter'],
                                                            'DataConfig', 'DataInfo']:
     """
@@ -803,6 +804,7 @@ def get_training_data_iters(sources: List[str],
     :param max_seq_len_target: Maximum target sequence length.
     :param bucketing: Whether to use bucketing.
     :param bucket_width: Size of buckets.
+    :param allow_empty: Unless True if no sentences are below or equal to the maximum length an exception is raised.
     :return: Tuple of (training data iterator, validation data iterator, data config).
     """
     logger.info("===============================")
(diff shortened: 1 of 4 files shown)


sockeye/data_io.py
```diff
@@ -533,6 +533,9 @@ def prepare_data(source_fnames: List[str],
     length_statistics = analyze_sequence_lengths(source_fnames, target_fname, source_vocabs, target_vocab,
                                                  max_seq_len_source, max_seq_len_target)
 
+    check_condition(length_statistics.num_sents > 0, "0 training sentence within maximum length, consider increasing "
```
fhieber: maybe plural "sentences"?

sockeye/data_io.py
```diff
@@ -644,6 +647,10 @@ def get_validation_data_iter(data_loader: RawParallelDatasetLoader,
     validation_length_statistics = analyze_sequence_lengths(validation_sources, validation_target,
                                                             source_vocabs, target_vocab,
                                                             max_seq_len_source, max_seq_len_target)
+
+    check_condition(validation_length_statistics.num_sents > 0, "0 validation sentence within maximum length, consider"
```
fhieber: maybe plural "sentences"?

sockeye/data_io.py
```diff
@@ -802,6 +809,10 @@ def get_training_data_iters(sources: List[str],
     # Pass 1: get target/source length ratios.
     length_statistics = analyze_sequence_lengths(sources, target, source_vocabs, target_vocab,
                                                  max_seq_len_source, max_seq_len_target)
+
+    check_condition(length_statistics.num_sents > 0, "0 training sentence within maximum length, consider increasing "
```
fhieber: maybe plural "sentences"?

fhieber:
Seems that this check is surfacing an issue with the CNN test in `test/integration/test_seq_copy_int.py`. Good to know :)

tdomhan:
Sure, I'll change the wording. I will maybe leave out the actual maximum length as at this part of the code we're already at `max_len + 1`, which would also be confusing for the user.

tdomhan:
As for the failed tests: The reason is that for the CNN we generate sequences that are longer than the maximum length which we set for the scorer so that all sentences are actually filtered. I'm not sure what the best fix is actually. I feel like `run_train_translate` has become quite hard to maintain and especially the scoring part seems quite brittle.

tdomhan:
I added a flag to allow for empty iterators for the scorer, which apparently is to be expected for the test. I feel like we should refactor the scoring test, but that seems outside of the scope of this PR.

fhieber:
> As for the failed tests: The reason is that for the CNN we generate sequences that are longer than the maximum length which we set for the scorer so that all sentences are actually filtered. I'm not sure what the best fix is actually. I feel like `run_train_translate` has become quite hard to maintain and especially the scoring part seems quite brittle.

+1, I actually started looking into this on the side. Will have a PR on it eventually.

PR #593, opened 2018-12-06
rewrote scoring to load data on demand and not skip sentences

mjpost:
The current scoring implementation (a) loads the entire training data into memory and (b) skips sentences that are too long so that the output is not parallel with the input. This PR fixes both of these issues. Data is loaded from disk one batch at a time via a new `BatchRawParallelSampleIter`, and sentences that are too long are truncated.

This code works, except for a few failing test cases. I wanted to see if you like the overall look of this before fixing those issues. In particular, I needed to change `parallel_iter()` so that it doesn't reset the passed-in iterators, but can continue from where it left off. If you don't like the look of this, I could factor it out into a separate function.

One application of this is [dual cross-entropy filtering](http://aclweb.org/anthology/W18-6478) of the training data.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/data_io.py
```diff
@@ -1114,17 +1168,32 @@ def create_sequence_readers(sources: List[str], target: str,
     return source_sequence_readers, target_sequence_reader
 
 
-def parallel_iter(source_iters: Sequence[Iterable[Optional[Any]]], target_iterable: Iterable[Optional[Any]]):
+def parallel_iter(source_iters: Sequence[Iterable[Optional[Any]]],
+                  target_iterable: Iterable[Optional[Any]],
+                  reiterate: bool = True):
```
mjpost: This is a deep change, perhaps. The function needs to be able to continue reading from an already-begun iterator in order to load the data from disk one batch at a time. The test cases pass, but are failing because a list gets returned in one place instead of a tuple. The code is also a bit uglier; I could refactor this into a `paraller_iter_continued()` function or something of that sort. I don't understand why the current code calls `iter()` at the top, since `zip()` will call this anyway.

sockeye/data_io.py
```diff
@@ -1477,10 +1546,109 @@ def load_state(self, fname: str):
...
+
+    def load_state(self, fname: str):
+        raise Exception('Not supported!')
+
+
 class ShardedParallelSampleIter(BaseParallelSampleIter):
     """
     Goes through the data one shard at a time. The memory consumption is limited by the memory consumption of the
-    largest shard. The order in which shards are traversed is changed with each reset.
+    largest shard. The order in whcirh shards are traversed is changed with each reset.
```
fhieber: is this a vim command gone wrong? :)

mjpost: When in doubt, blaming vim is a good idea.

sockeye/data_io.py
fhieber: I think its useful to have the functionality of reiterating and iterator, but imo the current version of your code mixes the concepts of `Iterable` and `Iterator` (https://stackoverflow.com/questions/9884132/what-exactly-are-iterator-iterable-and-iteration): the `parallel_iter()` function expects an `Iterable` (something that implements `__iter__` to return an `Iterator`). In your `reiterate=False` case, the function gets passed an `Iterator` (something that implements a `next()` method). Our implementation of `SequenceReader.__iter__()` is a `Generator`, which returns an `Iterator`, so the `iter()` call in `parallel_iter` creates that iterator. I am not sure why we call `iter()` explicitly before the `zip()` but I think there was a reason for it, @tdomhan ?

So in my opinion this new implementation is not properly reflected in the type annotation of `parallel_iter`, and I like the simplicity of the zip-for loop over the while True logic.
How about making `parallel_iter()` a wrapper function that calls `iter()` on the `Iterable` inputs, passing the result to `parallel_iterator()`, containing your current implementation minus the 1st part. Downside is that we loose the nice `zip` & for loop which I like a lot more than this `while True, break` stuff.

sockeye/data_io.py
```diff
@@ -1477,10 +1546,109 @@ def load_state(self, fname: str):
         pass
 
 
+class BatchedRawParallelSampleIter(BaseParallelSampleIter):
```
fhieber: Just an idea: the signature of the constructor of this class is not necessarily set in stone. Instead of creating the SequenceReaders for source/target outside, you could re-create them for every epoch in the implementation of this class, whenever `reset()` is called. Essentially, keep the sequence reader states (which are themselves streaming from disk) as states in the class below.

sockeye/data_io.py
```diff
@@ -848,25 +847,80 @@ def get_training_data_iters(sources: List[str],
...
+                                               target_vocab=target_vocab,
+                                               max_seq_len_source=max_seq_len_source,
+                                               max_seq_len_target=max_seq_len_target,
+                                               batch_size=batch_size,
+                                               fill_up=fill_up)
 
     return train_iter, validation_iter, config_data, data_info
 
 
+def get_scoring_data_iters(sources: List[str],
```
fhieber: As suggested in my previous comment: the main contents of this function before creating the `BatchedParallel...Iter` could be part of the constructor of this class itself.

mjpost: This is a good idea, though for posterity, I would like to record here [the Lord of the Rings allusion](http://tolkiengateway.net/wiki/Ring_Verse) I buried in the comments, which will be sadly lost to this refactoring:

```python
    # One bucket to hold them all,
    bucket = (max_seq_len_source, max_seq_len_target)

    # ...One iterator to raise them,
    sources_sentences, target_sentences = create_sequence_readers(sources, target, source_vocabs, target_vocab)

    # ...One loader to mold them all,
    data_loader = RawParallelDatasetLoader(buckets=[bucket],
                                           eos_id=target_vocab[C.EOS_SYMBOL],
                                           pad_id=C.PAD_ID,
                                           skip_blanks=False)

    # ...and with the model appraise them.
    scoring_iter = BatchedRawParallelSampleIter(data_loader=data_loader,
                                                sources_sentences=sources_sentences,
                                                target_sentences=target_sentences,
                                                bucket=bucket,
                                                batch_size=batch_size,
                                                bucket_batch_size=BucketBatchSize(bucket, batch_size, None),
                                                max_lens=(max_seq_len_source, max_seq_len_target),
                                                fill_up=fill_up,
                                                num_factors=len(sources))
```

mjpost:
Okay, I think this is complete. It now does not skip any sentences, and returns `-inf` for invalid pairs (such as a null source or target). 

I was also able to remove the `fill_up` policy that I had added earlier. I was using this to mark the last sentence in the last batch (since the last batch might be undersized), but that is better accomplished using the `pad` field of MXNet's `DataBatch`.

mjpost:
Getting these test cases to pass has been a hassle. They run fine on my laptop and local cluster, but fail when uploading. These are caused by a host of problems relating to rescoring translator output. I am going to propose that, unless and until these problems are fixed, that we drop the "fancy" scoring testing that ensures it gets the same scores as translation. Instead, I propose that integration tests will just test that scoring runs.

Here are the problems with trying to re-score translation output in this test setting with very small models:

- `sockeye.translate` is not guaranteed to include the cost of `</s>`. There is no way to know whether it has or not, since this information is stripped off.
- I tried to manually run the translator output through forced decoding, so as to force the inclusion of `</s>`, but this creates its own problems. Some of the sentences are too long and therefore get truncated. I was trying to extend the maximum output length, but apparently `--max-output-len` doesn't actually work for inference (argparse, somewhat confusing, opaquely maps this to `--max-output-len-num-std`, and has no effect).

In all my offline tests on real data, the scoring works perfectly. I think the changes here are important (not skipping source sentences is pretty important if you want to do filtering), so I'm hoping this will pass muster.

test/common.py
```diff
@@ -232,15 +232,26 @@ def check_train_translate(train_params: str,
...
-        test_scoring(data, translate_params, compare_translate_vs_scoring_scores)
+    # Test scoring. Inference does not guaranatee that </s> was generated, so we need to first
+    # force-decode against the output of the unconstrained system to get the real scores, that
+    # scoring will generate. Since it adds complexity to attach source factors to the input JSON
+    # object needed for constrained decoding, we only test scoring at the moment when not source
+    # factors are needed.
+    if 'test_source_factors' not in data:
+
+        # Run translate decoding against the reference
+        data = test_constrained_decoding_against_ref(data, translate_params)
```
fhieber: is this relevant for this PR? It seems that this is currently failing the system tests run with every commit on Travis.

test/common.py
```diff
@@ -440,18 +485,16 @@ def test_scoring(data: Dict[str, Any], translate_params: str, test_similar_score
     with open(out_path) as score_out:
         score_scores = [float(line.strip()) for line in score_out]
 
-    # Compare scored output to original translation output. There are a few tricks: for blank source sentences,
-    # inference will report a score of -inf, so skip these. Second, we don't know if the scores include the
-    # generation of </s> and have had length normalization applied. So, skip all sentences that are as long
-    # as the maximum length, in order to safely exclude them.
+    # Compare scored output to original translation output. Unfortunately, sockeye.translate doesn't enforce
+    # that </s> was generated, and we can't know outside the decoder. So we use scores from force-decoding
```
fhieber: with the recent PR that disabled scoring tests for integration tests (models which have only received 1 or 2 updates) scoring system tests as they were should pass just fine.

test/common.py
mjpost: Yes, with that commit, we can simplify this.

sockeye/scoring.py
```diff
@@ -249,9 +244,21 @@ def score(self,
                 target_ids = [int(x) for x in target.asnumpy().tolist()]
                 target_string = C.TOKEN_SEPARATOR.join(
                     data_io.ids2tokens(target_ids, self.target_vocab_inv, self.exclude_list))
-                score = score.asscalar()
+
+                # Report a score of 0 for invalid sentence pairs (empty source and/or target)
```
fhieber: 'Report a score of -inf' right?

fhieber:
System tests run as part of travis still seem to fail. Otherwise looks good. Once the travis build goes through, I will approve & merge.

mjpost:
I see what's happening. Will push a fix after testing locally.

mjpost:
***wipes brow***

PR #594, opened 2018-12-10
Fix BLEU score for decodings with restrict-lexicon in system tests

fhieber:
Relevant to the recent note in #590 about the ugliness of `run_train_translate()` for testing: `bleu` and `bleu_restrict` variables computed at the end for asserting on scores in system tests were always holding the same value, because we passed the regular hypotheses produced without lexicon-restriction into the bleu calculation, even for `bleu-restrict`. This fixes it. 
I can't easily trace the origin of this bug; seems like it was always there :)

until you can check this box.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -593,7 +593,9 @@ def run_train_translate(train_params: str,
 
         bleu_restrict = None
         if restrict_lexicon:
-            bleu_restrict = raw_corpus_bleu(hypotheses=hypotheses, references=references, offset=0.01)
+            with open(out_restrict_path) as out:
+                hypotheses_restrict = out.readlines()
+            bleu_restrict = raw_corpus_bleu(hypotheses=hypotheses_restrict, references=references, offset=0.01)
 
         # Run evaluate cli
         eval_params = "{} {} ".format(sockeye.evaluate.__file__,
--- a/test/system/test_seq_copy_sys.py
+++ b/test/system/test_seq_copy_sys.py
@@ -69,9 +69,9 @@ COMMON_TRAINING_PARAMS = " --checkpoint-frequency 1000 --optimizer adam --initia
      "--encoder rnn --decoder rnn --num-layers 1 --rnn-cell-type lstm --rnn-num-hidden 64 --num-embed 32 "
      " --rnn-attention-type mlp --rnn-attention-num-hidden 32 "
      " --batch-size 80 --batch-type word "
-     " --max-updates 5000  "
+     " --max-updates 5000 "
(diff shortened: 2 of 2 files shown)


fhieber:
This makes the system test 'Copy:word-based-batching:pruning' fail:
```python
>           assert bleu_restrict >= bleu_thresh
E           assert 0.9567168655138901 >= 0.99
test/system/test_seq_copy_sys.py:144: AssertionError
```

Should we adapt the threshold here or does this look like an actual error? (e.g. a combination of restrict-lexicon & pruning).

fhieber:
This is the only system test that runs beam pruning and non-default `beam-search-stop=first`.
Here are the decoder parameters for that test:
`--beam-size 5 --batch-size 2 --beam-prune 1 --beam-search-stop first`

- Changing `--beam-search-stop` from `first` to `all` makes the test pass with `bleu_restrict=1.0`.
- Removing `--beam-prune 1` and leaving `--beam-search-stop first` still fails with `bleu_restrict=0.96`.
- Changing `--batch-size` to 1 and leaving `--beam-prune 1 --beam-search-stop first` fails the test with `bleu_restrict=0.68` (!)

**This suggests that pruning isn't the issue, neither is batching per se, but `beam-search-stop first` in combination with vocabulary selection seems to behave strange across different batch sizes.**

In all cases, standard BLEU without vocabulary selection is 0.99 or 1.0.

@mjpost @tdomhan Any ideas

tdomhan:
mhh interesting. Thanks for fixing the test. If `bleu_restrict` wasn't tested before then the thresholds are probably not correct, no? Without digging deeper into what's going on at beam search its hard to tell though..

fhieber:
Yeah, but there was always jus one threshold for both bleu scores, with and without vocab selection.

fhieber:
I took a closer look at this and it seems that the lower bleu score is actually a search error. I analysed a single example input from the test data and both decodes with `--beam-search-stop first/all` behave exactly the same until the beam contains the first finished hypotheses. In that case `all` continues until max output length, eventually yielding a higher scoring (correct) output, whereas `first` obviously stops and returns a shorter (in that case wrong) output.

So as far as I can tell, this is working fine. It also makes sense to observe higher bleu_restrict with larger batch size, as we are forced to continue with beam search until all inputs in the inference batch have at least 1 finished hypotheses.

Given that this is an artifact of these small toy models, I would propose removing the `--beam-search-stop first` argument from the test so that this test just checks for a working prune feature.
We may have to think about another system test in the future with more robust data to test `--beam-search-stop first`.

PR #596, opened 2018-12-11
Added a check for vocabularies loaded from json.

tdomhan:
Adding a check that the vocabulary is valid to make sure user provided vocabularies meet all criteria.

### DIFF
--- a/sockeye/vocab.py
+++ b/sockeye/vocab.py
@@ -112,6 +112,36 @@ def vocab_to_json(vocab: Vocab, path: str):
         logger.info('Vocabulary saved to "%s"', path)
 
 
+def is_valid_vocab(vocab: Vocab) -> bool:
+    """
+    Checks if a vocabulary is valid. We define valid as:
+    1. All indices from 0 to num_words - 1 are present without duplicates.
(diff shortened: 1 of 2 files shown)


sockeye/vocab.py
```diff
@@ -112,6 +112,36 @@ def vocab_to_json(vocab: Vocab, path: str):
...
 
 
+def is_valid_vocab(vocab: Vocab) -> bool:
+    """
+    Checks if a vocabulary is valid. We define valid as:
+    1. All indices from 0 to num_words - 1 are present without duplicates.
+    2. All special symbols C.PAD_SYMBOL, C.UNK_SYMBOL, C.BOS_SYMBOL, C.EOS_SYMBOL are present.
+    3. PAD_ID has word id 0.
+    """
+    for symbol in [C.PAD_SYMBOL, C.UNK_SYMBOL, C.BOS_SYMBOL, C.EOS_SYMBOL]:
```
fhieber: Do we have a set/list of these defined in constants.py or somewhere?

tdomhan: We have `C.VOCAB_SYMBOLS` but I didn't want to use it, as if we extended it in the future we would make this check fail for old models even though they would still be compatible, potentially not noticing that we broke old models.

fhieber: Makes sense, thanks!

PR #601, opened 2018-12-16
Refactored run_train_translate function for integration and system testing

fhieber:
Rewrote end-to-end testing suite. Changes include:
- removed quiet arg
- expanded scoring tests to cover tests with prepared data
- reduced I/O in test code
- decode with vocab selection is always done for all integration and system tests
- secondary clis (other than train/translate) are now tested once with a tiny default model in a separate integration test -> system tests now focus on meeting the defined thresholds.

until you can check this box.


test/common.py
```diff
@@ -324,284 +315,321 @@ def run_train_translate(train_params: str,
...
-        eval_params = "{} {} ".format(sockeye.evaluate.__file__,
-                                      _EVAL_PARAMS_COMMON.format(hypotheses=out_path,
-                                                                 references=test_target_path,
-                                                                 metrics="bleu chrf rouge1",
-                                                                 quiet=quiet_arg), )
-        with patch.object(sys, "argv", eval_params.split()):
-            sockeye.evaluate.main()
+        bleu = raw_corpus_bleu(hypotheses=translate_outputs, references=test_references, offset=0.01)
+        chrf = raw_corpus_chrf(hypotheses=translate_outputs, references=test_references)
+        # TODO translate_outputs is wrong
```
fhieber: this is fixed in #594

test/common.py
```diff
@@ -256,17 +255,12 @@ def run_train_translate(train_params: str,
     :param dev_source_factor_paths: Optional list of paths to dev source factor files.
     :param test_source_factor_paths: Optional list of paths to test source factor files.
     :param use_prepared_data: Whether to use the prepared data functionality.
+    :param use_target_constraints: Whether to use lexical constraints in the second translation pass.
```
davvil: Not clear what the "second translation pass" means here. It sound like this is related to translate_params_equiv (which is not).

test/common.py
```diff
@@ -324,284 +315,321 @@ def run_train_translate(train_params: str,
...
-            params = "{} {} {}".format(sockeye.score.__file__,
-                                       _SCORE_PARAMS_COMMON.format(model=model_path,
-                                                                   source=test_source_path,
-                                                                   target=out_path,
-                                                                   output=scores_output_file),
-                                       score_params)
-
-            if test_source_factor_paths is not None:
-                params += _SCORE_WITH_FACTORS_COMMON.format(source_factors=" ".join(test_source_factor_paths))
+        if '--max-input-len' not in translate_params and _translate_output_is_valid(translate_outputs):
```
davvil: Did the test for use_prepare_data get lost? Or is it no longer needed?

davvil:
This is certainly an improvement, but I have the feeling the function is still doing too many things. It is good that we test for many things, but it still does many things that one wouldn't expect from a function called run_train_translate. E.g. it tests the evaluate CLI, averaging scores, parameter extraction, etc.

fhieber:
I fully agree, this is really just meant as a first step so that we have a better view on what the tests actually do.
I support splitting this up into separate integration tests for the different CLIs (averaging, extraction, etc.), the main thing to do there would be to have example models available.

fhieber:
this is ready for another review @davvil . I did some major refactor to reduce test time.

test/common.py
```diff
@@ -156,6 +141,10 @@ def tmp_digits_dataset(prefix: str,
                        sort_target: bool = False,
                        seed_train: int = 13, seed_dev: int = 13,
                        with_source_factors: bool = False):
+    """
```
tdomhan: We could maybe even add a `Dict` return type.

test/common.py
```diff
@@ -216,394 +206,299 @@ def tmp_digits_dataset(prefix: str,
 
 _SCORE_WITH_FACTORS_COMMON = " --source-factors {source_factors}"
 
-_EVAL_PARAMS_COMMON = "--hypotheses {hypotheses} --references {references} --metrics {metrics} {quiet}"
 
-_EXTRACT_PARAMS = "--input {input} --names target_output_bias --list-all --output {output}"
+def check_core_features(train_params: str,
```
tdomhan: mh the function name is maybe a little too generic. Maybe we could name this `test_train_translate` analogous to `run_train_translate`.

test/integration/test_seq_copy_int.py
```diff
@@ -158,39 +171,107 @@ def test_seq_copy(train_params: str,
...
+        assert averaged_params
+
+
+def _test_checkpoint_decoder(dev_source_path: str, dev_target_path: str, model_path: str):
+    """
+    Runs checkpoint decoder on 10% of the dev data and checks whether metric keys are present in the result dict.
+    """
+    with open(dev_source_path) as dev_fd:
+        num_dev_sent = sum(1 for _ in dev_fd)
+    sample_size = min(1, int(num_dev_sent * 0.1))
```
tdomhan: used to be 1%. A specific reason to increase? Anyway, shouldn't really matter.

test/common.py
fhieber: done

test/common.py
fhieber: you are right, will rename

test/integration/test_seq_copy_int.py
fhieber: no reason, just thought that it would be more useful to a couple through.

test/common.py
tdomhan: sounds good.

PR #613, opened 2019-01-03
[WIP] Iterative Pruning

mitchellgordon95:
I'd like to implement iterative pruning, as described in [The Lottery Ticket Hypothesis](https://arxiv.org/abs/1803.03635). The procedure is fairly simple:

1. Train the network as you like
2. Pick the weights in the network that have the lowest absolute value (usually 70%)
3. Fix those weights to zero for the rest of training (this is the pruning part)
4. Go back to 1 (this is the iterative part)

We can continue iteratively pruning either for a set number of iterations, or until we can no longer reach some accuracy threshold on the validation set.

It seems hard to get mxnet to pin certain weights to zero during training. Before I continue any further, I'd like to get some input on a proposed method of doing this:

1. Keep a weight mask in TrainingModel
2. Construct the mxnet.module as normal, and run forward_backward / update as normal
3. After updating, multiply relevant params by the weight mask, which will reset the weights to zero.

It's hacky, but the Module / Symbol APIs do not seem to support weight masks (although they do support entire params matrices being fixed). I've included a code-sketch in TrainingModel. Please let me know if this seems like a reasonable thing to do, or if I should take a different approach.

P.S. How should I go about testing changes in TrainingModel? TrainingModel does not seem to be tested explicitly anywhere else.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -56,6 +56,8 @@ class TrainingModel(model.SockeyeModel):
             unrolled to the full length.
     :param gradient_compression_params: Optional dictionary of gradient compression parameters.
     :param fixed_param_names: Optional list of params to fix during training (i.e. their values will not be trained).
+    :param param_masks: Optional dictionary of param names to weight masks. Falsey values in a mask will pin
+            those values to zero during training via element-wise matrix multiplication.
     """
 
     def __init__(self,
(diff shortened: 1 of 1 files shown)


fhieber:
Hi @mitchellgordon95,

thank you for your proposal! Without knowing the approach in detail, your suggestions seems to be the right way to go. It might make sense to spend a little bit of thought on whether the parameters can be multiplied with the masks directly on the GPU. Afaik, `module.get_params()` retrieves them from the (potentially multiple) devices, hence doing a copy (same for `module.set_weights()`). If you want this to have minimal impact on training speed, it might be better to go through the lower-level `ExecutorGroup` interface directly.
As for the TrainingModel question, you are right, we don't have good explicit testing of it. This is something that could be added in the context of this, but in general its hard to construct an instance with all its requirements. You may take a look at #601 whether some of these refactorings would help you with this.

Another thought: would you still run an early-stopping based training algorithm? Reason I am asking is wether we should start thinking about having different Trainer implementations and different TrainingModel implementations (RegularTrainingModel, PruningTrainingModel).

fhieber:
@mitchellgordon95 are you still working on this? If not, may I close this PR for now?

PR #626, opened 2019-01-22
Length ratio prediction 

artemsok:
This implements prediction of length ratios to be used in offsetting hypotheses' scores during beam search to avoid generating overly short translations. Initially based on the BP-Norm method from [here](http://arxiv.org/abs/1808.09582), that learns the length ratios, but includes Poisson loss and can be used with fixed length ratio, without predicting it for every sentence, via specifying the fixed ratio during inference via CLI or by taking from the training data estimates.

Submitting this for review now and will work on unit tests in the meantime.

until you can check this box.
- [X] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -1913,13 +2027,14 @@ def _get_best_from_beam(self,
...
         :param lengths: The lengths of all items in the beam. Shape: (batch * beam). Dtype: int32.
+        :param estimated_ref_lengths: Predicted reference lengths.
         :param constraints: The constraints for all items in the beam. Shape: (batch * beam).
         :param beam_histories: The beam histories for each sentence in the batch.
         :return: List of Translation objects containing all relevant information.
         """
         nbest_translations = []  # type: List[List[Translation]]
         histories = beam_histories if beam_histories is not None else [None] * self.batch_size  # type: List
-
+        ref_lengths = estimated_ref_lengths if estimated_ref_lengths is not None else np.full(self.batch_size * self.beam_size, None)
```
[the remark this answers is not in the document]
artemsok: Yeah, it's used in zip() later, so I thought I would replicate what's done to histories above.

sockeye/inference.py
```diff
@@ -2204,19 +2323,22 @@ class NormalizeAndUpdateFinished(mx.gluon.HybridBlock):
...
 
-    def hybrid_forward(self, F, best_word_indices, max_output_lengths, finished, scores_accumulated, lengths):
+    def hybrid_forward(self, F, best_word_indices, max_output_lengths,
+                       finished, scores_accumulated, lengths, ref_lengths):
         all_finished = F.broadcast_logical_or(best_word_indices == self.pad_id, best_word_indices == self.eos_id)
         newly_finished = F.broadcast_logical_xor(all_finished, finished)
         scores_accumulated = F.where(newly_finished,
-                                     scores_accumulated / self.length_penalty(lengths),
-                                     scores_accumulated)
+                               scores_accumulated / self.length_penalty(lengths) - self.brevity_penalty(lengths, ref_lengths),
```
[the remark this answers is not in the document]
artemsok: You mean having an explicit ```if``` branch where brevity penalty is not used?

sockeye/arguments.py
```diff
@@ -820,10 +820,26 @@ def add_training_args(params):
...
+                              type=str,
+                              default='none',
+                              choices=['none', 'normal', 'poisson'],
+                              help='Train with prediction of the ratio of reference length to input length. Default %(default)s.'
+                              'using the specified links function. Default %(default)s.')
+    train_params.add_argument('--lenratio-layers',
+                              type=int,
+                              default=1,
+                              choices=[1, 2],
+                              help='Number of layers for predicting the length ratio. Default %(default)s.')
```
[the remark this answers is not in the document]
artemsok: It becomes complicated to handle them, because we need to save and to pass all the internal layers' weights to initialize the length ratio prediction block during inference. Maybe we could add more later, after everything else is settled.

sockeye/training.py
```diff
@@ -142,10 +146,21 @@ def sym_gen(seq_lens):
...
+                predicted_lenratios = self.length_ratio(source_encoded)
+
+            for i, loss_symbol in enumerate(self.model_loss):
+                if isinstance(loss_symbol, loss.CrossEntropyLoss) and loss_symbol.name == C.SOFTMAX_NAME:
+                    add_loss = loss_symbol.get_loss(logits, labels)
+                if isinstance(loss_symbol, loss.LenRatioMSELoss):
+                    add_loss = loss_symbol.get_loss(predicted_lenratios, lenratio_labels)
+                if isinstance(loss_symbol, loss.LenRatioPoissonLoss):
+                    add_loss = loss_symbol.get_loss(predicted_lenratios, lenratio_labels)
+                loss_output = loss_output + add_loss if i else add_loss
```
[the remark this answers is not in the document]
artemsok: Good point, rewrote. Need more test since I remmeber I had some difficulties with grouping, but nor sure it was about this place.

sockeye/training.py
```diff
@@ -142,10 +146,21 @@ def sym_gen(seq_lens):
             # logits: (batch_size * target_seq_len, target_vocab_size)
             logits = self.output_layer(target_decoded)
 
-            loss_output = self.model_loss.get_loss(logits, labels)
-
+            if self.config.config_length_ratio.link != 'none':
+                # predicted_lenratios: (batch_size, 1)
+                predicted_lenratios = self.length_ratio(source_encoded)
+
+            for i, loss_symbol in enumerate(self.model_loss):
```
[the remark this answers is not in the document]
artemsok: Simplified a bit

sockeye/training.py
```diff
@@ -92,11 +92,16 @@ def _initialize(self,
...
         labels = mx.sym.reshape(data=mx.sym.Variable(C.TARGET_LABEL_NAME), shape=(-1,))
 
+        # returns a list of losses
         self.model_loss = loss.get_loss(self.config.config_loss)
 
         data_names = [C.SOURCE_NAME, C.TARGET_NAME]
         label_names = [C.TARGET_LABEL_NAME]
 
+        if self.config.config_length_ratio.link != 'none':
+            lenratio_labels = mx.sym.reshape(mx.sym.Variable(C.LENRATIO_LABEL_NAME), shape=(-1,))
```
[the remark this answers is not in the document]
artemsok: I actually had the same question about the ```label``` 10 lines above. Why is it there reshaped?
I think I copied it from above, and since it worked, didn't return to it.

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: not necessarily. It seems that with disabled brevity penalty the BrevityPenalty block is basically a no-op correct?

sockeye/model.py
```diff
@@ -128,6 +130,13 @@ def __init__(self, config: ModelConfig, prefix: str = '') -> None:
                                                weight_normalization=self.config.weight_normalization,
                                                prefix=self.prefix + C.DEFAULT_OUTPUT_LAYER_PREFIX)
 
+        # create length ratio prediction layer(s)
+        if hasattr(self.config, 'config_length_ratio') and self.config.config_length_ratio.link != 'none':
```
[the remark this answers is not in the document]
artemsok: I changed the check to weight > 0.0 everywhere now.

sockeye/loss.py
```diff
@@ -116,16 +147,112 @@ def get_loss(self, logits: mx.sym.Symbol, labels: mx.sym.Symbol) -> List[mx.sym.
...
+        self.ignore_label = ignore_label
+
+
+    def update(self, labels, preds):
+        for label, pred in zip(labels, preds):
+            batch_size = label.shape[0]
+            label = label.as_in_context(pred.context).reshape((label.size,1))
+            # mse: (batch,)
+            mse = mx.nd.square(label - pred)
+            mse = mx.nd.sum(mse)
```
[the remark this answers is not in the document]
artemsok: Added some comments

sockeye/training.py
```diff
@@ -92,11 +92,16 @@ def _initialize(self,
         target_length = utils.compute_lengths(target)
         labels = mx.sym.reshape(data=mx.sym.Variable(C.TARGET_LABEL_NAME), shape=(-1,))
 
+        # returns a list of losses
         self.model_loss = loss.get_loss(self.config.config_loss)
```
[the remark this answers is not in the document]
artemsok: Agreed, fixed

sockeye/training.py
```diff
@@ -833,15 +848,17 @@ def training_state_dirname(self) -> str:
         return os.path.join(self.model.output_dir, C.TRAINING_STATE_DIRNAME)
 
     @staticmethod
-    def _create_eval_metric(metric_name: str) -> mx.metric.EvalMetric:
+    def _create_eval_metric(metric_name: str, index: int) -> mx.metric.EvalMetric:
```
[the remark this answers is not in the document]
artemsok: A fossil from a previous incarnation, removed.

sockeye/training.py
```diff
@@ -850,7 +867,7 @@ def _create_eval_metric_composite(metric_names: List[str]) -> mx.metric.Composit
         """
         Creates a composite EvalMetric given a list of metric names.
         """
-        metrics = [EarlyStoppingTrainer._create_eval_metric(metric_name) for metric_name in metric_names]
+        metrics = [EarlyStoppingTrainer._create_eval_metric(metric_name, i) for i, metric_name in enumerate(metric_names)]
```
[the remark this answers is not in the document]
artemsok: True, removed

sockeye/utils.py
```diff
@@ -142,6 +142,23 @@ def compute_lengths(sequence_data: mx.sym.Symbol) -> mx.sym.Symbol:
     return mx.sym.sum(sequence_data != C.PAD_ID, axis=1)
 
 
+def compute_lenratio_label(input_seq: mx.nd.NDArray, label_seq: mx.nd.NDArray) -> mx.nd.NDArray:
+    """
+    Computes the ratio of reference length to the input length 
+    of the PAD_ID-padded data in sequence_data to zero.
```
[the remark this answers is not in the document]
artemsok: Fixed

sockeye/inference.py
```diff
@@ -1559,6 +1640,30 @@ def _encode(self, sources: mx.nd.NDArray, source_length: int) -> List[ModelState
...
+        for src, model in zip(source_encoded, self.models):
+            if hasattr(model, 'length_ratio'):
+                ratios.append(model.length_ratio(src,
+                                                 model.length_ratio_w_final,
+                                                 model.length_ratio_b_final,
+                                                 model.length_ratio_w_internal,
+                                                 model.length_ratio_b_internal))
+        # average the ratios over the models that actually could predict them
+        if len(ratios):
+            return mx.nd.mean(mx.nd.concat(*ratios, dim=1), axis=1, keepdims=True)
```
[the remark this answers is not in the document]
artemsok: Thanks

sockeye/arguments.py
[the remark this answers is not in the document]
fhieber: Why do we need the internal layers' weights in inference anyway? The reason we have the output layer weights optionally pulled out at inference time is the vocabulary selection feature, where you need to cut the computron graph of the decoder short to do a custom softmax over a reduced set of symbols.
However, your case is independent from that. You can just the symbolic definition of the length prediction subnetwork to the encoder module in the inference model and get the result.

artemsok:
Felix, thanks for all the comments! I'm fixing simple once in the first pass and will work on larger (correct hidden state averages, taking length ratio from the training data, etc.)

sockeye/arguments.py
```diff
@@ -820,10 +820,26 @@ def add_training_args(params):
                               help='How to normalize the loss. By default loss is normalized by the number '
                                    'of valid (non-PAD) tokens (%s).' % C.LOSS_NORM_VALID)
 
+    train_params.add_argument('--lenratio-link',
+                              type=str,
+                              default='none',
+                              choices=['none', 'normal', 'poisson'],
+                              help='Train with prediction of the ratio of reference length to input length. Default %(default)s.'
```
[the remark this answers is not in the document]
artemsok: Fixed that in the next commit, was quite nonsensical.

sockeye/inference.py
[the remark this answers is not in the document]
artemsok: Yes

CHANGELOG.md
```diff
@@ -10,11 +10,16 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.70]
+### Added 
+- Added learning or fixed length ratio to discourage hypotheses that are too short.
```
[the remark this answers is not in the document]
artemsok: Extended a bit.

sockeye/inference.py
```diff
@@ -144,15 +145,27 @@ def initialize(self, max_input_length: int, get_max_output_length_function: Call
...
-                assert self.output_layer.weight_norm is not None
-                weight = self.params[self.output_layer.weight_norm.weight.name].as_in_context(self.context)
-                scale = self.params[self.output_layer.weight_norm.scale.name].as_in_context(self.context)
-                self.output_layer_w = self.output_layer.weight_norm(weight, scale)
-            else:
-                self.output_layer_w = self.params[self.output_layer.w.name].as_in_context(self.context)
-            self.output_layer_b = self.params[self.output_layer.b.name].as_in_context(self.context)
+            self.output_layer_w, self.output_layer_b = self._get_output_layer_params(self.output_layer)
+
+    def _get_output_layer_params(self, layer: Optional[OutputLayer]) -> Tuple[mx.nd.NDArray, mx.nd.NDArray]:
```
fhieber: Return type is not correct, as it can return None

fhieber: is this function still needed?

sockeye/inference.py
```diff
@@ -188,7 +201,15 @@ def sym_gen(source_seq_len: int):
 
             data_names = [C.SOURCE_NAME]
             label_names = []  # type: List[str]
-            return mx.sym.Group(decoder_init_states), data_names, label_names
+            # optionally predict length ratios
+
+            predicted_length_ratios = []
+            length_ratio_data_name = []
+            if hasattr(self, 'length_ratio'):
```
fhieber: why would it not have the attribute? Is this the correct way of checking for it?

sockeye/arguments.py
```diff
@@ -820,10 +838,24 @@ def add_training_args(params):
...
+                              default=0.0,
+                              help='Weight for the loss part that corresponds to learning length ratios. Not used if equal to 0.0. Default %(default)s.')
+    train_params.add_argument('--length-ratio-task-layers',
+                              type=int_greater_or_equal(1),
+                              default=1,
+                              help='Number of fully-connected layers for predicting the length ratio. Default %(default)s.')
+    train_params.add_argument('--length-ratio-task-type',
+                              type=str,
+                              default='ratio',
+                              choices=['ratio', 'absolute'],
```
fhieber: I'd prefer adding third option None (which would be the default), then we don't have to check for the weight. This flag could also be named `--length-ratio-task`, then its clear that the other two flags are subflags related to this.

sockeye/arguments.py
```diff
@@ -820,10 +838,24 @@ def add_training_args(params):
...
+                              help='Predict either float length ratios (uses Gaussian link function) or integer translation lengths (Poisson link function). Default %(default)s.')
```
fhieber: How about: "If not None, adds an auxiliary task during training to predict source/target length ratios (mean squared error loss), or absolute lengths (poisson) loss."

sockeye/arguments.py
```diff
@@ -820,10 +838,24 @@ def add_training_args(params):
                               help='How to normalize the loss. By default loss is normalized by the number '
                                    'of valid (non-PAD) tokens (%s).' % C.LOSS_NORM_VALID)
 
+    train_params.add_argument('--length-ratio-task-weight',
+                              type=float_greater_or_equal(0.0),
+                              default=0.0,
+                              help='Weight for the loss part that corresponds to learning length ratios. Not used if equal to 0.0. Default %(default)s.')
```
fhieber: 'The weight of the auxiliary --length-task loss`

sockeye/arguments.py
```diff
@@ -820,10 +838,24 @@ def add_training_args(params):
...
+    train_params.add_argument('--length-ratio-task-layers',
+                              type=int_greater_or_equal(1),
+                              default=1,
+                              help='Number of fully-connected layers for predicting the length ratio. Default %(default)s.')
+    train_params.add_argument('--length-ratio-task-type',
```
fhieber: i would rename all flags to `--length-task*` since its not always ratio

sockeye/arguments.py
```diff
@@ -1297,23 +1329,43 @@ def add_inference_args(params):
...
                                default=0.0,
                                type=float,
                                help='Beta factor for the length penalty used in beam search: '
-                                    '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
+                                    '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s.')
+
+    decode_params.add_argument('--brevity-penalty-type',
+                               default='learned',
+                               type=str,
+                               choices=['learned', 'fixed'],
```
fhieber: should be defined in constants.py.
Maybe 'mean' instead of 'fixed'?

sockeye/constants.py
```diff
@@ -177,6 +178,10 @@
 SOURCE_LENGTH_NAME = "source_length"
 TARGET_NAME = "target"
 TARGET_LABEL_NAME = "target_label"
+LENRATIO_LABEL_NAME = "lenratio_label"
+LENRATIO_NAME = "lenratio"
```
fhieber: len -> length

sockeye/image_captioning/train.py
```diff
@@ -243,6 +244,8 @@ def create_model_config(args: argparse.Namespace,
                                      config_encoder=config_encoder,
                                      config_decoder=config_decoder,
                                      config_loss=config_loss,
+                                     config_length_ratio_loss=config_length_ratio_loss,
```
fhieber: where do these come from?

sockeye/layers.py
```diff
@@ -225,6 +228,85 @@ def __call__(self,
...
+        # TODO: keep weight here redundant because the weight is also stored in the loss config.
+        # but it's used to test if we need length prediction at all
+        self.weight = weight
+
+
+class LengthRatio:
+    """
+    Defines the length-ratio prediction layer of Sockeye.
+
+    :param hidden_size: Decoder hidden size.
```
fhieber: its the encoder hidden size, no?

sockeye/layers.py
```diff
@@ -225,6 +228,85 @@ def __call__(self,
...
+    :param prefix: Prefix used for naming.
+    """
+
+    def __init__(self,
+                 hidden_size: int,
+                 num_layers: int,
+                 weight_normalization: bool,
```
fhieber: does this really need weight normalization?

sockeye/arguments.py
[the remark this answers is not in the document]
artemsok: True, implemented it all in symbols; actually shortened the implementation a bit. And can now support arbitrary number of FF layers.

sockeye/layers.py
```diff
@@ -225,6 +228,85 @@ def __call__(self,
...
+                 name: str = C.LENRATIO_OUTPUT_NAME) -> None:
+        utils.check_condition(num_layers >= 1, "LengthRatio's num_layers has to be >=1.")
+        self.prefix = prefix
+        self.name = name
+        self.num_layers = num_layers
+
+        self.layers = []
+        # first the internal layers
+        for l in range(self.num_layers - 1):
+            self.layers.append(OutputLayer(hidden_size=hidden_size,
```
fhieber: I guess this works fine, but I find it a bit confusing to use OutputLayer here.
You could just as well do the following in ___call__():
```python
data
for l in range(self.num_layers - 1):
  data = mx.sym.FullyConnected(data=data, num_hidden=hidden_size, flatten=False, name=self.name + 'fc%d' %l)
data = mx.sym.FullyConnected(data=data, num_hidden=1, flatten=false ...)
```

sockeye/arguments.py
```diff
@@ -1309,6 +1325,16 @@ def add_inference_args(params):
                                type=float,
                                help='Beta factor for the length penalty used in beam search: '
                                     '(beta + len(Y))**alpha/(beta + 1)**alpha. Default: %(default)s')
+    decode_params.add_argument('--fixed-length-ratio',
```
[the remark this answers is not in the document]
artemsok: Implemented taking the training data estimate if the CLI argument ```--brevity-penalty-type=learned``` and ```--brevity-penalty-fixed-length-ratio``` is missing.

sockeye/loss.py
```diff
@@ -35,51 +35,84 @@ class LossConfig(config.Config):
...
 
     def __init__(self,
                  name: str,
-                 vocab_size: int,
-                 normalization_type: str,
-                 label_smoothing: float = 0.0) -> None:
+                 vocab_size: Optional[int] = None,
+                 normalization_type: Optional[str] = None,
+                 label_smoothing: float = 0.0,
+                 link: Optional[str] = None,
```
fhieber: these two new parameters should be prefixed with something like 'length_task_' to make clear that they are not for the main loss.
In the future we need to rewrite this anyway.

fhieber: I am also not sure we should call this 'link'. Why not just loss and name the options poisson and mse?

sockeye/model.py
```diff
@@ -59,6 +59,8 @@ def __init__(self,
                  config_encoder: encoder.EncoderConfig,
                  config_decoder: decoder.DecoderConfig,
                  config_loss: loss.LossConfig,
+                 config_length_ratio_loss: loss.LossConfig = None,
```
fhieber: - type should be Optional[loss.LossConfig] if it can be none
- would rename to more general length_task_*

sockeye/train.py
```diff
@@ -361,7 +363,8 @@ def create_data_iters_and_vocabs(args: argparse.Namespace,
             max_seq_len_source=max_seq_len_source,
             max_seq_len_target=max_seq_len_target,
             bucketing=not args.no_bucketing,
-            bucket_width=args.bucket_width)
+            bucket_width=args.bucket_width,
+            use_length_ratios=args.length_ratio_task_weight > 0.0)
```
fhieber: i would decide all of these based on whether args.length_task is not None

sockeye/training.py
```diff
@@ -839,9 +858,12 @@ def _create_eval_metric(metric_name: str) -> mx.metric.EvalMetric:
         """
         # output_names refers to the list of outputs this metric should use to update itself, e.g. the softmax output
         if metric_name == C.ACCURACY:
-            return utils.Accuracy(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME])
+            return utils.Accuracy(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME], label_names=[C.TARGET_LABEL_NAME])
```
fhieber: thanks, its good to be explicit about these!

sockeye/translate.py
```diff
@@ -92,11 +92,21 @@ def run_translate(args: argparse.Namespace):
             restrict_lexicon = TopKLexicon(source_vocabs[0], target_vocab)
             restrict_lexicon.load(args.restrict_lexicon, k=args.restrict_lexicon_topk)
         store_beam = args.output_type == C.OUTPUT_HANDLER_BEAM_STORE
+
+        if args.brevity_penalty_type == 'fixed':
```
fhieber: use constants

sockeye/translate.py
```diff
@@ -108,7 +118,8 @@ def run_translate(args: argparse.Namespace):
                                           store_beam=store_beam,
                                           strip_unknown_words=args.strip_unknown_words,
                                           skip_topk=args.skip_topk,
-                                          sample=args.sample)
+                                          sample=args.sample,
+                                          fixed_length_ratio=constant_length_ratio)
```
fhieber: maybe rename ether the variable to fixed_length_ratio or the argument to constant_length_ratio

sockeye/inference.py
```diff
@@ -35,6 +35,7 @@
 from . import utils
 from . import vocab
 from .log import is_python34
+from .layers import OutputLayer
```
fhieber: no longer needed, no?

sockeye/inference.py
```diff
@@ -144,15 +145,15 @@ def initialize(self, max_input_length: int, get_max_output_length_function: Call
         self.decoder_module.init_params(arg_params=self.params, aux_params=self.aux_params, allow_missing=False)
 
         if self.cache_output_layer_w_b:
-            if self.output_layer.weight_normalization:
+            if layer.weight_normalization:
```
fhieber: this seems off. please revert to original code.

sockeye/inference.py
```diff
@@ -1778,6 +1877,12 @@ def _beam_search(self,
         # (0) encode source sentence, returns a list
         model_states = self._encode(source, source_length)
 
+        # extract predicted length ratio from model states (if present), average them over models 
+        # and return the predicted reference length
+        model_states, estimated_reference_lengths = self._extract_length_ratios(model_states,
```
fhieber: I'd prefer if the 'extraction' would be done directly in the code that runs the module and gets the outputs. Basically hide these internals from the beam search code here a little bit.

sockeye/arguments.py
artemsok: Done

sockeye/inference.py
artemsok: fixed

sockeye/inference.py
artemsok: yes

sockeye/translate.py
artemsok: done

sockeye/translate.py
artemsok: done

sockeye/train.py
artemsok: Agree, changed

sockeye/model.py
artemsok: fixed

sockeye/loss.py
artemsok: Prepended the var names. 

Technically, `link` is more precise because the loss is the same in both cases -- MLE, it just looks differently because of different distributions.

sockeye/layers.py
artemsok: Good question, I guess we don't know yet, but removed for now.

sockeye/image_captioning/train.py
artemsok: Yeah, forgot to remove, removed now.

sockeye/constants.py
artemsok: Done

sockeye/arguments.py
artemsok: Done

sockeye/arguments.py
artemsok: Done

sockeye/arguments.py
artemsok: Done

sockeye/loss.py
fhieber: good point, let's leave 'link'

sockeye/model.py
```diff
@@ -71,7 +73,10 @@ def __init__(self,
         self.config_embed_target = config_embed_target
         self.config_encoder = config_encoder
         self.config_decoder = config_decoder
-        self.config_loss = config_loss
+        self.config_loss = [config_loss] 
+        if config_length_task_loss:
```
fhieber: better to check explicitly for None

sockeye/model.py
```diff
@@ -128,6 +133,15 @@ def __init__(self, config: ModelConfig, prefix: str = '') -> None:
                                                weight_normalization=self.config.weight_normalization,
                                                prefix=self.prefix + C.DEFAULT_OUTPUT_LAYER_PREFIX)
 
+        # create length ratio prediction layer(s)
+        if self.config.config_length_task:
```
fhieber: check for none

sockeye/inference.py
```diff
@@ -188,7 +188,15 @@ def sym_gen(source_seq_len: int):
 
             data_names = [C.SOURCE_NAME]
             label_names = []  # type: List[str]
-            return mx.sym.Group(decoder_init_states), data_names, label_names
+
+            # predict length ratios
+            predicted_length_ratios = []  # type: List[mx.nd.NDArray]
+            length_ratio_data_name = []  # type: List[str]
+            if hasattr(self, 'length_ratio'):
```
fhieber: why not `if self.length_ratio_task is not None`?

sockeye/inference.py
```diff
@@ -306,11 +314,11 @@ def run_encoder(self,
                                 provide_data=self._get_encoder_data_shapes(source_max_length))
 
         self.encoder_module.forward(data_batch=batch, is_train=False)
-        decoder_states = self.encoder_module.get_outputs()
+        decoder_init_states = self.encoder_module.get_outputs()
 
         # replicate encoder/init module results beam size times
-        decoder_states = [mx.nd.repeat(s, repeats=self.beam_size, axis=0) for s in decoder_states]
-        return ModelState(decoder_states)
+        decoder_init_states = [mx.nd.repeat(s, repeats=self.beam_size, axis=0) for s in decoder_init_states]
```
fhieber: I think this is where we should 'extract' the predicted lengths/length ratios and hide this complexity entirely from the Translator. I think it would make sense to expand ModelState to have another member (length/length_ratio) that is optional.

sockeye/arguments.py
```diff
@@ -821,10 +838,25 @@ def add_training_args(params):
                               help='How to normalize the loss. By default loss is normalized by the number '
                                    'of valid (non-PAD) tokens (%s).' % C.LOSS_NORM_VALID)
 
+    train_params.add_argument('--length-task',
+                              type=str,
+                              default=None,
+                              choices=[C.LENGTH_TASK_RATIO, C.LENGTH_TASK_LENGTH],
+                              help='If not None, adds an auxiliary task during training to predict source/target length ratios '
```
tdomhan: As the help string is regarding the CLI I think it would be better to write `if specified ...` instead of `If not None` as there's no `None` on the CLI.

sockeye/arguments.py
```diff
@@ -821,10 +838,25 @@ def add_training_args(params):
...
+                                    '(mean squared error loss), or absolute lengths (poisson) loss. Default %(default)s.')
+    train_params.add_argument('--length-task-weight',
+                              type=float_greater_or_equal(0.0),
+                              default=0.0,
```
tdomhan: So this gives two ways of disabling: `--length-task` and the weight. How about setting it to a reasonable default so that when someone turns it on via `--length-task` it is not indirectly disabled through the weight.

sockeye/arguments.py
```diff
@@ -821,10 +838,25 @@ def add_training_args(params):
                               help='How to normalize the loss. By default loss is normalized by the number '
                                    'of valid (non-PAD) tokens (%s).' % C.LOSS_NORM_VALID)
 
+    train_params.add_argument('--length-task',
```
tdomhan: Would it make sense to rename to make clearer that its about length prediction? `--predict-length` or `length-prediction-task` ...

sockeye/inference.py
artemsok: Added estimated_reference_Length to ModelState

sockeye/inference.py
artemsok: FIxed

sockeye/model.py
artemsok: Fixed (also in another place or two)

sockeye/arguments.py
artemsok: Will change to 1.0 (also there is a warning at [L144 in the sockeye/model.py](https://github.com/awslabs/sockeye/pull/626/files#diff-64c7821f0957b25e2f587b010d67e3afR144))

sockeye/arguments.py
artemsok: Yeah, it was initially called ```--length-ratio...``` but, since it can already predict float ratios or integer lengths depending on the loss, it was a bit confusing, so was changed to just ```--length-task..```. Felix also pointed out that different length-related tasks might be implemented in the future.

sockeye/arguments.py
artemsok: Agree, will fix in the next commit.

sockeye/inference.py
artemsok: Moved now it to Inference Model.

sockeye/inference.py
artemsok: fixed

sockeye/model.py
artemsok: fixed

sockeye/scoring.py
```diff
@@ -166,6 +185,15 @@ def sym_gen(seq_lens):
             zeros = mx.sym.zeros_like(scores)
             sums = mx.sym.sum(mx.sym.where(labels != 0, scores, zeros), axis=1) / (self.length_penalty(target_length - 1))
 
+            # Deal with the potential presence of brevity penalty
+            if False:
```
fhieber: Is that a TODO? :)

sockeye/utils.py
```diff
@@ -142,6 +142,22 @@ def compute_lengths(sequence_data: mx.sym.Symbol) -> mx.sym.Symbol:
...
+def compute_length_ratio(input_seq: mx.nd.NDArray, label_seq: mx.nd.NDArray) -> mx.nd.NDArray:
+    """
+    Computes the ratio of label lengths to the input lengths.
+
+    :param input_seq: Input sequence. Shape: (batch_size, seq_len, 1).
+    :param label_seq: Label sequence. Shape: (batch_size, seq_len).
+    :return: Length data. Shape: (batch_size,).
+    """
+    # input_lengths: (batch_size, )
+    input_lengths = (input_seq != C.PAD_ID).sum(axis=1).squeeze()
```
fhieber: you can avoid the squeeze if you set keepdims=False in sum()

test/common.py
```diff
@@ -406,14 +406,18 @@ def test_scoring(data: Dict[str, Any], translate_params: str):
...
-        params = translate_params.split()
-        for i, param in enumerate(params):
-            if param == '--softmax-temperature':
-                score_params = '--softmax-temperature {}'.format(params[i + 1])
-                break
+    params = translate_params.split()
+    for i, param in enumerate(params):
+        if param in relevant_params:
+            score_params = '{} {}'.format(param, params[i + 1])
+    print(score_params)
```
fhieber: remove once finished

sockeye/evaluate.py
```diff
@@ -90,6 +90,18 @@ def raw_corpus_rougel(hypotheses: Iterable[str], references: Iterable[str]) -> f
...
+def raw_corpus_lenratio(hypotheses: Iterable[str], references: Iterable[str]) -> float:
+    """
+    Simple wrapper around ROUGE-1 implementation.
+
+    :param hypotheses: Hypotheses stream.
+    :param references: Reference stream.
+    :return: ROUGE-L score as float between 0 and 1.
+    """
+    ratios = [ len(h.split())/len(r.split()) for h, r in zip(hypotheses, references) ]
+    return sum(ratios)/len(ratios)
```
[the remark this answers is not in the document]
artemsok: done

test/common.py
```diff
@@ -456,6 +460,7 @@ def test_scoring(data: Dict[str, Any], translate_params: str):
         # the length penalty was applied
         if len(translate_tokens) >= max_len - 2:
             continue
+        print(translate_score, score_score)
```
fhieber: remove once finished

test/integration/test_seq_copy_int.py
```diff
@@ -148,7 +148,48 @@
...
+     False, False),
+    # LSTM encoder-decoder with attention and absolute length prediciton, and test brevity penalty
+    ("--encoder rnn --decoder rnn --num-layers 1 --rnn-cell-type lstm --rnn-num-hidden 8 --num-embed 4 "
+     " --rnn-attention-type mlp"
+     " --rnn-attention-num-hidden 8 --batch-size 2 --loss cross-entropy --optimized-metric perplexity --max-updates 2"
+     " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --batch-type sentence "
+     " --decode-and-evaluate 0 --length-task ratio --length-task-weight 1.0 --length-task-layers 1",
+     "--beam-size 2 --brevity-penalty-type learned --brevity-penalty-weight 1.0",
+     False, False),
+    # LSTM encoder-decoder with attention and absolute length prediciton, and test brevity penalty
```
fhieber: typo

test/integration/test_seq_copy_int.py
```diff
@@ -148,7 +148,48 @@
...
+     False, False),
+    # LSTM encoder-decoder with attention and length ratio prediction, and test brevity penalty
+    ("--encoder rnn --decoder rnn --num-layers 1 --rnn-cell-type lstm --rnn-num-hidden 8 --num-embed 4 "
+     " --rnn-attention-type mlp"
+     " --rnn-attention-num-hidden 8 --batch-size 2 --loss cross-entropy --optimized-metric perplexity --max-updates 2"
+     " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --batch-type sentence "
+     " --decode-and-evaluate 0 --length-task length --length-task-weight 1.0 --length-task-layers 1",
+     "--beam-size 2 --brevity-penalty-type learned --brevity-penalty-weight 1.0",
+     False, False),
+    # LSTM encoder-decoder with attention and absolute length prediciton, and test brevity penalty
```
fhieber: typo

test/integration/test_seq_copy_int.py
```diff
@@ -148,7 +148,48 @@
...
+     False, False),
+    # LSTM encoder-decoder with attention and length ratio prediction
+    ("--encoder rnn --decoder rnn --num-layers 1 --rnn-cell-type lstm --rnn-num-hidden 8 --num-embed 4 "
+     " --rnn-attention-type mlp"
+     " --rnn-attention-num-hidden 8 --batch-size 2 --loss cross-entropy --optimized-metric perplexity --max-updates 2"
+     " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --batch-type sentence "
+     " --decode-and-evaluate 0 --length-task length --length-task-weight 1.0 --length-task-layers 1",
+     "--beam-size 2",
+     False, False),
+    # LSTM encoder-decoder with attention and absolute length prediciton
```
fhieber: tpoy

test/integration/test_seq_copy_int.py
```diff
@@ -148,7 +148,48 @@
      " --batch-size 2 --max-updates 2 --batch-type sentence  --decode-and-evaluate 0"
      " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --lhuc all",
      "--beam-size 2 --beam-prune 1",
-     False, False)]
+     False, False),
```
fhieber: Can you maybe merge some of these tests with the existing ones to avoid adding too many here? I'd suggest using transformer integration tests as these are the default models in Sockeye now.

I would also like to have a system test (one that asserts on ppl and bleu) that uses this new feature (even though its not too interesting to predict length in sequence copy/sort).

sockeye/layers.py
```diff
@@ -225,6 +228,88 @@ def __call__(self,
...
+                 weight_internal: Optional[mx.nd.NDArray] = None,
+                 bias_internal: Optional[mx.nd.NDArray] = None):
+        """
+        Transformation to the length ratio. Returns a scalar.
+
+        :param source_encoded: Encoder representation for n elements. Shape: (n, source_encoded_length, encoder_depth).
+        :return: Logits. Shape(n, 1).
+        """
+        if isinstance(source_encoded, mx.sym.Symbol):
+            data = mx.sym.mean(source_encoded, axis=1)
```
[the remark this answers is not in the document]
artemsok: Very good point about the padding. I fixed this in the one of the last commits.

Regarding using other inputs, shall we have an option?

sockeye/model.py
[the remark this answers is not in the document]
artemsok: And later changed to a None check..

sockeye/utils.py
```diff
@@ -142,6 +142,23 @@ def compute_lengths(sequence_data: mx.sym.Symbol) -> mx.sym.Symbol:
...
+    """
+    Computes the ratio of reference length to the input length 
+    of the PAD_ID-padded data in sequence_data to zero.
+
+    :param input_seq: Input sequence. Shape: (batch_size, seq_len, 1).
+    :param label_seq: Label sequence. Shape: (batch_size, seq_len).
+    :return: Length data. Shape: (batch_size, 1).
+    """
+    # input_seq: (batch_size, seq_len, 1)
+    input_lengths = mx.nd.squeeze(mx.nd.sum(mx.nd.broadcast_not_equal(input_seq, mx.nd.zeros((1,))), axis=1), axis=1)
```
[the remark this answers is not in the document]
artemsok: Fixed in one of the last commits.

fhieber:
(I don't seem to be able to respond to your comment about the Block warnings. Hence doing it here)

I guess this is just a notification from mxnet to make the user aware that you are passing unnecessary inputs to a block (which internally could cause copying of superfluous data). If we cannot get rid of it fine, but maybe we can define the blocks in a way such that we can decide ahead of time whether we need this input data or not, maybe via subclassing.

sockeye/inference.py
```diff
@@ -143,15 +149,32 @@ def initialize(self, max_input_length: int, get_max_output_length_function: Call
...
-                weight = self.params[self.output_layer.weight_norm.weight.name].as_in_context(self.context)
-                scale = self.params[self.output_layer.weight_norm.scale.name].as_in_context(self.context)
-                self.output_layer_w = self.output_layer.weight_norm(weight, scale)
-            else:
-                self.output_layer_w = self.params[self.output_layer.w.name].as_in_context(self.context)
-            self.output_layer_b = self.params[self.output_layer.b.name].as_in_context(self.context)
+            self.output_layer_w, self.output_layer_b = self._get_output_layer_params(self.output_layer)
+
+        if hasattr(self.config, 'config_length_ratio') and self.config.config_length_ratio.link != 'none':
+            if self.config.config_length_ratio.layers == 2:
```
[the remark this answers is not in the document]
artemsok: As per later comments and commits this much if this logic is now in InferenceModel

sockeye/layers.py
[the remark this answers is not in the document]
fhieber: lets leave that for the future

sockeye/image_captioning/train.py
```diff
@@ -42,6 +42,7 @@
 from .. import data_io
 from .. import encoder
 from .. import loss
+from .. import layers
```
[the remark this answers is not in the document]
artemsok: tnx, not any more

sockeye/utils.py
artemsok: I think it's False by default, the squeeze is there to remove the original 3rd axis.

sockeye/inference.py
```diff
@@ -1593,15 +1667,31 @@ def _translate_nd(self,
...
+        if self.constant_length_ratio > 0.0:
+            # override all ratios with the constant value
+            length_ratios = mx.nd.full(val=self.constant_length_ratio, shape=(num_seq, 1), ctx=self.context)
+        else:
+            ratios = [state.estimated_length_ratio for state in model_states if state.estimated_length_ratio is not None]
+            if len(ratios) > 0:  # some model predicted a ratio?
+                # average the ratios over the models that actually we able to predict them
+                length_ratios = mx.nd.mean(mx.nd.stack(*ratios, axis=1), axis=1)
+            else:
+                length_ratios = mx.nd.zeros((num_seq, 1), ctx=self.context)
```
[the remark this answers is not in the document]
artemsok: It receives zero reference lengths and the penalty has no effect.

sockeye/inference.py
```diff
@@ -1593,15 +1667,31 @@ def _translate_nd(self,
...
+ 
+        encoded_source_length=self.models[0].encoder.get_encoded_seq_len(source_length)
+        return model_states, length_ratios * encoded_source_length
```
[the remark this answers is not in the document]
artemsok: Right. Yes, looks like it's not a problem (dividing floats by ints), at least MXNet does not complain.

CHANGELOG.md
```diff
@@ -10,11 +10,14 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.73]
+### Added 
+- Added learning or fixed length ratio to discourage hypotheses that are too short, based on [Yang et al. 2018](http://arxiv.org/abs/1808.09582). Can be enabled in training with a non-zero positive --length-ratio-task-weight.
```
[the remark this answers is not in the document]
artemsok: Added a few sentences.

test/integration/test_seq_copy_int.py
artemsok: OK, leaving just 2 tests with the Transformer model. Will submit a revision after fixing scoring test failures again.

sockeye/layers.py
```diff
@@ -225,6 +228,85 @@ def __call__(self,
...
+        Transformation to the length ratio. Returns a vector.
+
+        :param source_encoded: Encoder representation for n elements. Shape: (n, source_encoded_length, hidden_size).
+        :param source_encoded_length: A vector of encoded sequence lengths. Shape: (n,).
+        :return: Predictions of len(hypothesis)/len(reference). Shape(n, 1).
+        """
+        # data: (n, hidden_size)
+        data = LengthRatio.average_sources(source_encoded, source_encoded_length)
+        # MLP
+        for l in range(self.num_layers - 1):
```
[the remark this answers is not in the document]
artemsok: OK, it worked with Dense but left it with symbolic API for the moment.

sockeye/training.py
```diff
@@ -142,9 +147,23 @@ def sym_gen(seq_lens):
             # logits: (batch_size * target_seq_len, target_vocab_size)
             logits = self.output_layer(target_decoded)
 
-            loss_output = self.model_loss.get_loss(logits, labels)
+            net_outputs = []
+            for loss_instance in self.model_losses:
```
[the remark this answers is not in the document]
fhieber: +1

sockeye/utils.py
```diff
@@ -142,6 +142,22 @@ def compute_lengths(sequence_data: mx.sym.Symbol) -> mx.sym.Symbol:
     return mx.sym.sum(sequence_data != C.PAD_ID, axis=1)
 
 
+def compute_length_ratio(input_seq: mx.nd.NDArray, label_seq: mx.nd.NDArray) -> mx.nd.NDArray:
```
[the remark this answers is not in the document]
fhieber: I like this idea! You don't even need an if-statement there. If you don't use the computed length_ratio later in the graph, it will be pruned from it anyway.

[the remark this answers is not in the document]
artemsok: Very good point! I didn't think of it. Trying out now.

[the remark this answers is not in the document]
artemsok: Implemented this, data_io is not modified now. Had to write a small override for ```update_dict``` in a custom metric to copy outputs to labels to make it work.

sockeye/model.py
```diff
@@ -71,7 +73,10 @@ def __init__(self,
         self.config_embed_target = config_embed_target
         self.config_encoder = config_encoder
         self.config_decoder = config_decoder
-        self.config_loss = config_loss
+        self.config_loss = [config_loss] 
+        if config_length_task_loss is not None:
+            self.config_loss.append(config_length_task_loss)
```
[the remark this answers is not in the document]
artemsok: Should be fixed now.

sockeye/training.py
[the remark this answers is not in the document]
artemsok: Now stored separately.

sockeye/training.py
```diff
@@ -142,9 +151,21 @@ def sym_gen(seq_lens):
...
+                # predicted_length_ratios: (batch_size, 1)
+                predicted_length_ratios = self.length_ratio(source_encoded, source_encoded_length)
+                if isinstance(self.length_task_loss, loss.MSELoss):
+                    loss_symbol = self.length_task_loss.get_loss(predicted_length_ratios, length_ratio)
+                elif isinstance(self.length_task_loss, loss.PoissonLoss):
+                    # convert float predictions to integer lengths for the Poisson loss to be well defined
+                    predicted_reference_lengths = predicted_length_ratios * source_encoded_length.reshape((-1, 1))
+                    loss_symbol = self.length_task_loss.get_loss(predicted_reference_lengths, target_length)
+                # return both the loss symbol and the computed length_ratio to be used in metrics
+                net_outputs.extend([loss_symbol, mx.sym.BlockGrad(length_ratio, name=C.LENRATIO_NAME)])
```
fhieber: this should probably be `[loss_symbol, BlockGrad(predicted_length_ratios, name=LENRATIO_OUTPUT), BlockGrad(length_ratio, name=LENRATIO_LABEL)]`

sockeye/training.py
```diff
@@ -839,9 +860,12 @@ def _create_eval_metric(metric_name: str) -> mx.metric.EvalMetric:
...
         # output_names refers to the list of outputs this metric should use to update itself, e.g. the softmax output
         if metric_name == C.ACCURACY:
-            return utils.Accuracy(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME])
+            return utils.Accuracy(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME], label_names=[C.TARGET_LABEL_NAME])
         elif metric_name == C.PERPLEXITY:
-            return mx.metric.Perplexity(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME])
+            return mx.metric.Perplexity(ignore_label=C.PAD_ID, output_names=[C.SOFTMAX_OUTPUT_NAME], label_names=[C.TARGET_LABEL_NAME], name=C.PERPLEXITY)
+        elif metric_name == C.LENRATIO_MSE:
+            return loss.LengthRatioMSEMetric(name=C.LENRATIO_MSE,
+                                  output_names=[C.LENRATIO_OUTPUT_NAME], label_names=[C.LENRATIO_OUTPUT_NAME])
```
fhieber: these two names should be different, one for the predicted_length_ratio symbol's name, and one for the computed length_ratio symbol's name.

sockeye/layers.py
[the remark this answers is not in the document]
artemsok: Reverted to gluon.Dense

sockeye/train.py
```diff
@@ -694,7 +707,7 @@ def create_training_model(config: model.ModelConfig,
                                             context=context,
                                             output_dir=output_dir,
                                             provide_data=train_iter.provide_data,
-                                            provide_label=train_iter.provide_label,
+                                            provide_label=train_iter.provide_label,# + length_task_provide_label,
```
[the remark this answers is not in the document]
artemsok: forgot to delete

sockeye/training.py
artemsok: Thanks, now should return all three things in the Group (four with the softmax symbol)

sockeye/training.py
artemsok: Yep, fixed this.

sockeye/inference.py
```diff
@@ -2028,15 +2125,19 @@ def _assemble_translation(sequence: np.ndarray,
...
         :param seq_score: Array of length-normalized negative log-probs.
+        :param estimated_reference_length: Estimated reference length (if any).
         :param beam_history: The optional beam histories for each sentence in the batch.
         :return: A Translation object.
         """
         length = int(length)
         sequence = sequence[:length].tolist()
         attention_matrix = attention_lists[:length, :]
         score = float(seq_score)
+        estimated_reference_length=float(estimated_reference_length) if estimated_reference_length else None
```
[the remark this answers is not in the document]
artemsok: Without it the estimated_reference_length is a numpy array of shape (1,1) -- it breaks somehow the output handler (at least that's why one of the tests fail).

sockeye/inference.py
```diff
@@ -2249,19 +2350,22 @@ class NormalizeAndUpdateFinished(mx.gluon.HybridBlock):
...
 
-    def hybrid_forward(self, F, best_word_indices, max_output_lengths, finished, scores_accumulated, lengths):
+    def hybrid_forward(self, F, best_word_indices, max_output_lengths,
+                       finished, scores_accumulated, lengths, reference_lengths):
         all_finished = F.broadcast_logical_or(best_word_indices == self.pad_id, best_word_indices == self.eos_id)
         newly_finished = F.broadcast_logical_xor(all_finished, finished)
         scores_accumulated = F.where(newly_finished,
-                                     scores_accumulated / self.length_penalty(lengths),
-                                     scores_accumulated)
+                               scores_accumulated / self.length_penalty(lengths) - self.brevity_penalty(lengths, reference_lengths),
```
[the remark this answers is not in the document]
artemsok: Oh, that's why the warnings.. Fixed!

sockeye/inference.py
```diff
@@ -2249,19 +2350,22 @@ class NormalizeAndUpdateFinished(mx.gluon.HybridBlock):
...
                  length_penalty_alpha: float = 1.0,
-                 length_penalty_beta: float = 0.0) -> None:
+                 length_penalty_beta: float = 0.0,
+                 brevity_weight: float = 0.0) -> None:
         super().__init__()
         self.pad_id = pad_id
         self.eos_id = eos_id
         with self.name_scope():
             self.length_penalty = LengthPenalty(alpha=length_penalty_alpha, beta=length_penalty_beta)
+            self.brevity_penalty = BrevityPenalty(weight=brevity_weight)
```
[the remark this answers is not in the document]
artemsok: fixed

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: ok

test/integration/test_seq_copy_int.py
```diff
@@ -45,23 +46,23 @@
      " --checkpoint-frequency 2 --optimizer adam --initial-learning-rate 0.01 --batch-type sentence "
      " --decode-and-evaluate 0",
      "--beam-size 2 --softmax-temperature 0.01",
-     False, False),
+     False, False, False),
```
fhieber: so now all tests disable the check against scores?

test/common.py
fhieber: still a print here :)

sockeye/training.py
```diff
@@ -92,11 +92,21 @@ def _initialize(self,
         target_length = utils.compute_lengths(target)
         labels = mx.sym.reshape(data=mx.sym.Variable(C.TARGET_LABEL_NAME), shape=(-1,))
 
+        # returns a list of losses
```
fhieber: no longer true, right?

sockeye/scoring.py
```diff
@@ -213,11 +241,13 @@ class Scorer:
     def __init__(self,
                  model: ScoringModel,
                  source_vocabs: List[vocab.Vocab],
-                 target_vocab: vocab.Vocab) -> None:
+                 target_vocab: vocab.Vocab,
+                 constant_length_ratio = -1) -> None:
```
fhieber: could add a type here

test/integration/test_seq_copy_int.py
artemsok: Yes, disabled after a discussion that we don't need to ensure their closeness in integration tests and because they fail sporadically. Added a reminder in a TODO.

artemsok:
Deleted changes already merged in #644 and rebased.

fhieber:
Looks good to me.

PR #627, opened 2019-01-24
added ability to carry over extra fields in JSON input

mjpost:
This PR causes Sockeye to carry through to the output fields in a JSON input that are not used by Sockeye. This is consistent with a UIMA-style pipeline interface, where each step of the pipeline is adding information non-destructively to an object. For example, if your input is:

    { "text": "Diese Pull-Anfrage ist furchtbar nützlich .", "document_id": 74 }

`sockeye.translate` will output (when called with `--json-input --output-type json`)

    { "text": "Diese Pull-Anfrage ist furchtbar nützlich .", "translation": "This pull request is very useful .", "document_id": 74 }

It also renames the `NBestOutputHandler`, since this is a special case of the more general `JSONOutputHandler`.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -874,6 +881,29 @@ def __init__(self,
...
+    def json(self, align_threshold: float = 0.0) -> Dict:
+        """
+        Returns a dictionary suitable for json.dumps() representing all
+        the information in the class.
+
+        :param align_threshold: If alignments are defined, only print ones over this threshold.
+        :return: A dictionary.
+        """
+        _d = { 'id': self.sentence_id,
+               'translated_text': self.translation,
```
fhieber: maybe rename to 'translation'?

sockeye/inference.py
```diff
@@ -21,7 +21,7 @@
 import time
 from collections import defaultdict
 from functools import lru_cache, partial
-from typing import Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union, Set
+from typing import Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union, Set, Any
```
fhieber: Any doesn't seem to be used

sockeye/inference.py
```diff
@@ -836,6 +842,7 @@ class TranslatorOutput:
     :param score: Negative log probability of generated translation.
     :param beam_histories: List of beam histories. The list will contain more than one
            history if it was split due to exceeding max_length.
+    :param raw_dict: Raw dictionary to append when outputting JSON.
```
fhieber: not used

sockeye/inference.py
```diff
@@ -605,21 +605,24 @@ class TranslatorInput:
     :param tokens: List of input tokens.
     :param factors: Optional list of additional factor sequences.
     :param constraints: Optional list of target-side constraints.
+    :param raw_dict: Optional raw dictionary of arbitrary input data.
```
fhieber: I would suggest renaming to something like `pass_through_{data,dict}`

sockeye/output_handler.py
```diff
@@ -393,32 +393,26 @@ def handle(self,
...
-            extracted_alignments.append(list(get_alignments(alignment_matrix, threshold=self.threshold)))
-
-        d_ = {"translations": t_output.nbest_translations,
-              "scores": t_output.nbest_scores,
-              "alignments": extracted_alignments}
+        d_ = t_output.json(self.align_threshold)
+        # copy over keys from the input that aren't present in the output
+        if t_input.raw_dict is not None:
+            for key, value in t_input.raw_dict.items():
+                if not key in d_:
```
fhieber: this behavior should be documented in the changelog and in the docstring for TranslatorInput: data created by the translator has preference over input data with the same key.

sockeye/inference.py
```diff
@@ -874,6 +881,29 @@ def __init__(self,
...
 
+    def json(self, align_threshold: float = 0.0) -> Dict:
+        """
+        Returns a dictionary suitable for json.dumps() representing all
+        the information in the class.
+
+        :param align_threshold: If alignments are defined, only print ones over this threshold.
+        :return: A dictionary.
+        """
+        _d = { 'id': self.sentence_id,
```
fhieber: TranslatorInput calls it `sentence_id`, should be consistent here for people that use the Input/Output objects in an API.

fhieber: [supernit] I do not like the space after `{` :)

sockeye/output_handler.py
```diff
@@ -393,32 +393,26 @@ def handle(self,
...
-        extracted_alignments = []
-        for alignment_matrix in t_output.nbest_attention_matrices:
-            extracted_alignments.append(list(get_alignments(alignment_matrix, threshold=self.threshold)))
-
-        d_ = {"translations": t_output.nbest_translations,
-              "scores": t_output.nbest_scores,
-              "alignments": extracted_alignments}
+        d_ = t_output.json(self.align_threshold)
+        # copy over keys from the input that aren't present in the output
+        if t_input.raw_dict is not None:
```
fhieber: as said above, I think this logic should be part of TranslatorOutput. What do you think?

mjpost: I think this behavior is a function of the output handler: TranslatorOutput assembles all the fields that Sockeye cares about, and then the output handler takes care of merging them. But I don't feel strongly about it, and mostly I did it this way because TranslatorOutput objects don't have access to the TranslatorInput, so we'd have to add that to its constructor. I'm happy to do that and move it over if you prefer!

sockeye/inference.py
mjpost: It's used in line 894 when the dictionary `d_` is created.

sockeye/output_handler.py
fhieber: Couldn't we add a "pass_through_dict" argument to the constructor of TranslatorOutput (just like your doc string change suggests), and at construction time of the object (https://github.com/awslabs/sockeye/blob/master/sockeye/inference.py#L1499) we pass in `translator_input.pass_through_dict` ? Basically, we make it the responsibility of the `_make_result()` method to carry over this additional data from input to output.

mjpost: Sure, sounds good.

mjpost:
Oh, good idea. I also forgot to document `--source-factors-combine sum`, so I'll add that, too.

PR #629, opened 2019-01-25
Rename "checkpoint frequency" to "checkpoint interval"

TomLippincott:
"frequency" is the opposite of the intended meaning, so applied the following regular expression to entire code repo:

  s/checkpoint( |-|_)frequency/checkpoint\1interval/g

The root 'pytest' and 'pytest test/system' checks pass as before.

- [Y ] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]'
until you can check this box.
- [Y ] Unit tests pass (`pytest`)
- [N ] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?
- [Y ] System tests pass (`pytest test/system`)
- [Y ] Passed code style checking (`./style-check.sh`)
- [N ] You have considered writing a test
- [N ] Updated major/minor version in `sockeye/__init__.py`. Major version bump if this is a backwards incompatible change.
- [N ] Updated CHANGELOG.md


TomLippincott:
Quick comment: this command-line parameter is confusing to newcomers simply because what it governs is conceptually straightforward and the meaning (seems) intuitive.  Having to instead remember that "higher frequency means less often for this value" may not be enough to justify merging this PR, but it bothers me enough to create it!

tdomhan:
That is very understandable! I have also thought about renaming this parameter several times. The main reason I haven't is that I didn't want to break existing training scripts. Maybe we could temporarily still accept `--checkpoint-frequency` and print a deprecation warning. Let's see what others think.

TomLippincott:
@tdomhan Apologies for the delay: I've changed it to accept either switch with identical behavior, but to print a deprecation warning for '--checkpoint-frequency'.

mjpost:
Looks like you've fallen behind master and will need to merge that in.

TomLippincott:
@tdomhan @mjpost OK, I believe this is all set: let me know if anything else is needed!

sockeye/arguments.py
```diff
@@ -847,7 +847,8 @@ def add_training_args(params):
                               type=int,
                               default=None,
                               help='Maximum number of samples. Default: %(default)s.')
-    train_params.add_argument(C.TRAIN_ARGS_CHECKPOINT_FREQUENCY,
+    train_params.add_argument(C.TRAIN_ARGS_CHECKPOINT_INTERVAL,
```
fhieber: Thanks for updating! how about adding the `--checkpoint-interval` argument separately and changing the existing `--checkpoint-frequency` argument to be suppressed (`help=argparse.SUPPRESS`, `default=argparse.SUPPRESS`) and be mapped to `--checkpoint-interval` directly via `dest=checkpoint_interval`? (see for example the `fill-up` argument).

This would not break anyone's training scripts but the option is now hidden and new users would simply use --checkpoint-interval. In the next major version update we could then remove the suppressed flags.

sockeye/train.py
```diff
@@ -908,4 +908,6 @@ def train(args: argparse.Namespace) -> training.TrainState:
         return training_state
 
 if __name__ == "__main__":
+    if any([a == "--checkpoint-frequency" or a.startswith("--checkpoint-frequency=") for a in sys.argv]):
```
fhieber: I would vote for removing this in favor of my suggestion above.

tdomhan: Does `SUPPRESS` produce some kind of warning message? If not, I'd still have something around to make users aware of the deprecation.

fhieber: I dont think it does. Is there a better way than checking sys.argv?

TomLippincott: I think the main alternative would be checking the parser object itself when instantiated for training: this was my initial impulse.  I went with checking sys.argv so it wouldn't be nested deeper than the training script (since it's a stop-gap) but maybe that wasn't a good reason.  I'll try the SUPPRESS approach, and see if I can still have the old usage trigger the warning, hopefully w/o the sys.argv string comparisons.

TomLippincott:
@fhieber How does this look?  It involves a new logger action subclass, but it's now 1) using argparse.SUPPRESS, 2) checking the actual argument object for the deprecated switch use, and 3) printing the warning via the logging system after it's been initialized.

PR #630, opened 2019-01-28
Replaced expand_dims with reshape ops to avoid data copies

fhieber:
According to https://github.com/apache/incubator-mxnet/issues/13998 `expand_dims` creates copies of the data, instead of returning a view on the original data with different shape (like `reshape` does). This removes usage of `expand_dims` in Sockeye and replaces with `reshape` (which is sometimes a bit more complicated).

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,13 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.72]
+### Changed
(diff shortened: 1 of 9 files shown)


tdomhan:
A little harder to read, but given the memory usage this makes a lot of sense!

szha:
Did you measure the effect of this PR? My impression is that it should not matter, as the memory planning pass will utilize the FInplaceIdentity attribute of these ops and do in-place update when possible. If you observe otherwise, it might indicate a bug and I would love to use the workload to improve existing memory plan.

fhieber:
I didn't at first, but just did a quick test. It actually makes 0 difference in terms of GPU memory usage or samples/sec. I'll post a PR to revert it for improved readability.

PR #641, opened 2019-02-15
Refactored batch_size in InferenceModel and Translator to allow for dynamic batch sizes

fhieber:
This adds support for dynamic batch sizes during inference, reviving #516.
The main changes are:
- `Translator.translate()` will now adapt batch sizes to the number of inputs/chunks up to `Translator.max_batch_size`
- Added ability to set the loglevel for score/train/translate CLIs to DEBUG.
- Slightly changed logging style in translate CLI so that reported number of batches used for translating the input is still correct with dynamic batch sizes.

There are some refactorings on how the batch size information is stored and used across internal methods. Specifically it does the following:
- Removed the `batch_size` argument from `InferenceModel` constructor. the `initialize()` method now has a new arguement `default_batch_size`, which is similar to `default_bucket_key` supposed to be the largest batch size expected during an inference run.
- `Translator` still has a `batch_size` member, currently still set to the default batch size of the model. However, there are no constants initialized in the constructor anymore, that depend on the batch size. This is done entirely in the respective HybridBlocks and/or at every call to `_beam_search()`, allowing to change the batch size in the future with every request.

I needed to do two rather involved changes:
- in the first iteration of beam search, we only look at the 'first rows' of the scores array, since all hypotheses have the same history. I changed the logic around that a little bit so that the TopK block always can expect the scores array input to be of shape `(batch*beam, vocab_size)`. Before it was `(batch, vocab_size)` for `t==1`, which was impossible to get working without explicit batch size information in the TopK HybridBlock.
- `SampleK` still has a batch size argument, which is now expected to be the default batch size. This is used to create a default `best_hyp_indices` array, that is `sliced` like an input to the HybridBlock. The constraint here is again that no batch size of a request is larger than the default batch size of the translator/inference model.

**revision 2**:
I realized that supporting dynamic batch sizes is actually quite trivial. The last version of this PR supports this now. A `Translator` instance has a new property `max_batch_size` (which internally maps to the existing `self.batch_size` to not break existing user scripts. This defines the maximum batch size that can be translated in one beam search run. If the user passes more `TranslatorInputs` to `translate()` than `max_batch_size`, Sockeye will translate them in sequence.
Also, we now use a smaller batch size if the last batch from a sequence is 'underfilled'.

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
@mr-lee FYI

fhieber:
Right now, batch decoding after this change does not seem to work with lexical constraints:
```
        for row, col, seq_score in zip(best_ids, best_word_ids, sequence_scores):
            row = int(row.asscalar())
            col = int(col.asscalar())
>           if hypotheses[row].is_valid(col):
E           IndexError: list index out of range
```

There are two integration tests in `lexical_constraints.py`, the one with `batch_size=1` passes.

fhieber:
@mjpost 
[This](https://github.com/awslabs/sockeye/blob/master/sockeye/inference.py#L1307) seems to be a member that is used across `translate()` calls when using lexical constraints. Could you briefly describe what kind of state is stored there?

mjpost:
You're too fast for me.

sockeye/inference.py
```diff
@@ -1823,17 +1836,19 @@ def _beam_search(self,
                 best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores, target_dists, finished)
             else:
                 # On the first timestep, all hypotheses have identical histories, so force topk() to choose extensions
-                # of the first row only
+                # of the first row only by setting all other rows to inf
                 if t == 1 and not self.skip_topk:
-                    scores = scores[self.batch_indices, :]
```
fhieber: this change is actually creating a little bit of a performance hit. It is needed to abstract batch size away from the TopK HybridBlock. Does anyone have a better idea on how to do this?

fhieber: using a mask, allocated before the beam search iteration is faster. Still a tiny slowdown for larger batch sizes (-0.1 sent/sec), but that should be fine.

sockeye/inference.py
```diff
@@ -1372,11 +1372,17 @@ def _log_linear_interpolation(predictions):
     def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutput]:
         """
         Batch-translates a list of TranslatorInputs, returns a list of TranslatorOutputs.
-        Splits oversized sentences to sentence chunks of size less than max_input_length.
+        Empty or bad inputs are skipped.
+        Splits inputs larger than Translator.max_input_length into chunks of size max_input_length.
+        Translates in batches of Translator.max_batch_size or less if there are fewer inputs/chunks.
+        If there are more inputs/chunks than Translator.max_batch_size, multiple batches of size
+        min(len(remaining_chunks), max_batch_size) are translated in sequence.
```
davvil: I think the min() expression here confuses more than it helps.

sockeye/inference.py
```diff
@@ -1435,26 +1441,23 @@ def translate(self, trans_inputs: List[TranslatorInput]) -> List[TranslatorOutpu
                             "constraint" if len(trans_input.constraints) == 1 else "constraints",
                             ", ".join(" ".join(x) for x in trans_input.constraints))
 
+        num_bad_empty = len(translated_chunks)
```
davvil: Is this correct? Perhaps the github diff doesn't show if this is modified somewhere else.

sockeye/inference.py
fhieber: Do you have a better phrasing? I tried and failed :)

sockeye/inference.py
fhieber: it should be. translated_chunks is populated above for bad and empty inputs and then extended below for the actual translations.

sockeye/inference.py
mjpost: "Splits inputs longer than Translator.max_input_length into segments of size max_input_length, and then groups segments into batches of at most Translator.max_batch_size. Too-long segments that were split are reassembled into a single output after translation."

fhieber: wonderful prose, thank you!

mjpost: I tried to work in a Jane Austen reference, but couldn't pull it off.

PR #642, opened 2019-02-18
Disable dropout layers in inference

fhieber:
According to [this issue](https://github.com/apache/incubator-mxnet/issues/13825) and [this one](https://github.com/apache/incubator-mxnet/issues/14198), the Dropout operator causes some slowdown, even in the forward pass of non-training inference runs.
This PR disables the creation of dropout layers when constructing the model graph for inference at loading time of the models. Its a bit hacky, but provides consistent speedups when measured on a p3 with different batch sizes. (~ +0.2 sent/sec).

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/config.py
+++ b/sockeye/config.py
@@ -139,8 +139,8 @@ class Config(yaml.YAMLObject, metaclass=TaggedYamlObjectMetaclass):
 
     def disable_dropout(self):
         """
-        Sets the value of all float-valued attributes that contain 'dropout' in their name to 0.0.
-        Causes model graph construction to skip dropout layers.
+        Sets the value of all float-valued attributes in this config (or any of its children) that contain 'dropout'
+        in their name to 0.0.
         """
         for attr, val in self.__dict__.items():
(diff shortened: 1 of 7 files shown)


sockeye/config.py
```diff
@@ -133,3 +136,15 @@ def copy(self, **kwargs):
...
+    def disable_dropout(self):
+        """
+        Sets the value of all attributes that contain 'dropout' to 0.0.
+        """
+        for attr, val in self.__dict__.items():
+            if 'dropout' in attr:
+                logger.debug("Setting %s to 0.0", attr)
+                setattr(self, attr, 0.0)
+
+            elif isinstance(val, Config):
```
davvil: I would change the order of the checks. If I see it correctly in the current form, if a subconfig has an attribute called 'dropout' it won't recurse into it.

fhieber: Good point about the edge case of a subconfig named 'dropout' :)
Will revise.

PR #646, opened 2019-02-25
Detect divergence and report it with the convergence status in the metrics file

artemsok:
Sometimes training diverges with average training and validation perplexities sky-rocketing. To detect this early, we can test if the average validation perplexity surpasses double the the target vocabulary size and stop training. Additionally, we report the convergence and divergence status in the metrics file to provide some more information on why training stopped: because it actually converged, diverged or stopped for some the reason, like, hitting one of the max_{epochs, updates} limits.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -591,7 +591,7 @@ class EarlyStoppingTrainer:
                 # (2) determine improvement
                 has_improved = False
                 previous_best = self.state.best_metric
-                # at this point state.self..metrics doesn't have validation results yet
+                # at this point state.self.metrics doesn't have validation results yet
                 current_checkpoint_val_metric = {"%s-val" % name:val for name, val in metric_val.get_name_value()}
                 for checkpoint, metric_dict in enumerate(self.state.metrics + [current_checkpoint_val_metric], 1):
                     value = metric_dict.get("%s-val" % early_stopping_metric, self.state.best_metric)
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -592,8 +592,8 @@ class EarlyStoppingTrainer:
                 has_improved = False
                 previous_best = self.state.best_metric
                 # at this point state.self..metrics doesn't have validation results yet
-                current_checkpoint_val_metric = [{"%s-val" % name:val for name, val in metric_val.get_name_value()}]
-                for checkpoint, metric_dict in enumerate(self.state.metrics + current_checkpoint_val_metric, 1):
+                current_checkpoint_val_metric = {"%s-val" % name:val for name, val in metric_val.get_name_value()}
+                for checkpoint, metric_dict in enumerate(self.state.metrics + [current_checkpoint_val_metric], 1):
                     value = metric_dict.get("%s-val" % early_stopping_metric, self.state.best_metric)
                     if utils.metric_value_is_better(value, self.state.best_metric, early_stopping_metric):
                         self.state.best_metric = value
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -764,9 +764,13 @@ class EarlyStoppingTrainer:
         for name, value in metric_train.get_name_value():
             if np.isfinite(value):
                 checkpoint_metrics["%s-train" % name] = value
+            else:
+                logger.warning("A non-finite value skipped for training metric %s", name)
         for name, value in metric_val.get_name_value():
             if np.isfinite(value):
                 checkpoint_metrics["%s-val" % name] = value
+            else:
+                logger.warning("A non-finite value skipped for validation metric %s", name)
 
         if process_manager is not None:
             result = process_manager.collect_results()
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -588,10 +588,21 @@ class EarlyStoppingTrainer:
                 for name, val in metric_val.get_name_value():
                     logger.info('Checkpoint [%d]\tValidation-%s=%f', self.state.checkpoint, name, val)
 
-                # (2) determine improvement
+                # (2) wait for checkpoint decoder results and fill self.state.metrics
+                if process_manager is not None:
+                    result = process_manager.collect_results()
+                    if result is not None:
+                        decoded_checkpoint, decoder_metrics = result
+                        # The first checkpoint before any gradient updates is ignored
+                        if decoded_checkpoint > 0:
+                            self.state.metrics[decoded_checkpoint - 1].update(decoder_metrics)
+                            self.tflogger.log_metrics(decoder_metrics, decoded_checkpoint)
+                    process_manager.start_decoder(self.state.checkpoint)
+
+                # (3) determine improvement
                 has_improved = False
                 previous_best = self.state.best_metric
-                # at this point state.self.metrics doesn't have validation results yet
+                # at this point state.self.metrics doesn't have perplexity validation results yet
                 current_checkpoint_val_metric = {"%s-val" % name:val for name, val in metric_val.get_name_value()}
                 for checkpoint, metric_dict in enumerate(self.state.metrics + [current_checkpoint_val_metric], 1):
                     value = metric_dict.get("%s-val" % early_stopping_metric, self.state.best_metric)
@@ -611,7 +622,7 @@ class EarlyStoppingTrainer:
                     logger.info("Validation-%s has not improved for %d checkpoints, best so far: %f",
                                 early_stopping_metric, self.state.num_not_improved, self.state.best_metric)
 
-                # (3) determine stopping
+                # (4) determine stopping
                 if 0 <= max_num_not_improved <= self.state.num_not_improved:
                     logger.info("Maximum number of not improved checkpoints (%d) reached: %d",
                                 max_num_not_improved, self.state.num_not_improved)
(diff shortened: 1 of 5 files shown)


sockeye/training.py
```diff
@@ -586,7 +588,27 @@ def fit(self,
...
+                        self.state.converged = False
+
+                    if min_updates is not None and self.state.updates < min_updates:
+                        logger.info("Minimum number of updates (%d) not reached yet: %d",
+                                    min_updates, self.state.updates)
+                        self.state.converged = False
+
+                    if min_samples is not None and self.state.samples < min_samples:
+                        logger.info("Minimum number of samples (%d) not reached yet: %d",
+                                    min_samples, self.state.samples)
```
fhieber: is there actually a `self.state.converged = False` missing?

artemsok: Strange. looks like it's was missing all along, in the previous code. Adding in the next commit.

sockeye/utils.py
```diff
@@ -791,7 +791,7 @@ def write_metrics_file(metrics: List[Dict[str, Any]], path: str):
     """
     with open(path, 'w') as metrics_out:
         for checkpoint, metric_dict in enumerate(metrics, 1):
-            metrics_str = "\t".join(["{}={}".format(name, value) for name, value in sorted(metric_dict.items())])
+            metrics_str = "\t".join(["{}={:f}".format(name, value) for name, value in sorted(metric_dict.items())])
```
tdomhan: Why the change to fixed point?

sockeye/training.py
```diff
@@ -747,11 +757,16 @@ def _update_metrics(self,
...
         gpu_memory_usage = utils.get_gpu_memory_usage(self.model.context)
         checkpoint_metrics['used-gpu-memory'] = sum(v[0] for v in gpu_memory_usage.values())
+        checkpoint_metrics['converged'] = self.state.converged
+        checkpoint_metrics['diverged'] = self.state.diverged
 
+        # copy values from checkpoint_metrics, skipping NaNs and infinities that break numpy's histograms
         for name, value in metric_train.get_name_value():
-            checkpoint_metrics["%s-train" % name] = value
+            if np.isfinite(value):
+                checkpoint_metrics["%s-train" % name] = value
```
tdomhan: how about adding a warning here and below?

sockeye/utils.py
artemsok: This is for the read_metrics() @[L780](https://github.com/awslabs/sockeye/blob/09068cdeba9be538fccd78496a6b60a26d2a995f/sockeye/utils.py#L780) from utils.py that assumes everything is float.

sockeye/training.py
artemsok: Added

sockeye/training.py
```diff
@@ -747,11 +757,16 @@ def _update_metrics(self,
...
         for name, value in metric_val.get_name_value():
-            checkpoint_metrics["%s-val" % name] = value
+            if np.isfinite(value):
```
tdomhan: actually what are we guarding against here? I tried invoking an issue by setting value to `np.inf`, but that didn't cause any issues. Was this related to Tensorboard? If so I think we should remove values only for tensorboard logging (or for whichever part can not handle `inf`). For the metric file itself I would argue that having the `inf` is actually valuable for debugging, no?

tdomhan: So this seems to be caused by `add_histogram` in `TensorboardLogger`. So to guard against this problem we should add the check (+ warning) there. The check for finite for scalars we can then keep.

sockeye/training.py
```diff
@@ -611,6 +634,13 @@ def fit(self,
                     logger.info("Validation-%s has not improved for %d checkpoints, best so far: %f",
                                 early_stopping_metric, self.state.num_not_improved, self.state.best_metric)
 
+                # (5) detect divergence with respect to perplexity value at the last checkpoint
+                if self.state.metrics and not has_improved:
+                    # perplexity cannot exceed the number of target words, using the double to avoid precision issues
+                    last_ppl_value = self.state.metrics[-1]["%s-val" % C.PERPLEXITY]
+                    if not np.isfinite(last_ppl_value) or last_ppl_value > 2 * len(self.target_vocab):
+                        self.state.diverged = True
```
tdomhan: I think when we detect that the model diverged a logging message would be appropriate.

sockeye/training.py
artemsok: Moving the check to TensorboardLogger.log_metrics() as per offline discussion.

sockeye/utils.py
artemsok: Now doing the check in the read_metrics().

sockeye/training.py
artemsok: Added

sockeye/training.py
```diff
@@ -586,7 +588,28 @@ def fit(self,
                 for name, val in metric_val.get_name_value():
                     logger.info('Checkpoint [%d]\tValidation-%s=%f', self.state.checkpoint, name, val)
 
-                # (3) update training metrics
+                # (2) determine stopping
```
tdomhan: Why did this move before the actual update of `num_not_improved`? Wouldn't lead this to models training for one checkpoint longer than before?

sockeye/training.py
```diff
@@ -611,6 +634,13 @@ def fit(self,
                     logger.info("Validation-%s has not improved for %d checkpoints, best so far: %f",
                                 early_stopping_metric, self.state.num_not_improved, self.state.best_metric)
 
+                # (5) detect divergence with respect to perplexity value at the last checkpoint
+                if self.state.metrics and not has_improved:
+                    # perplexity cannot exceed the number of target words, using the double to avoid precision issues
```
tdomhan: maybe phrase as "should not"

tdomhan:
mypy seems unhappy now
```
mypy --ignore-missing-imports --follow-imports=silent @typechecked-files --no-strict-optional
sockeye/utils.py:783: error: Incompatible types in assignment (expression has type "float", target has type "bool")
The command "mypy --ignore-missing-imports --follow-imports=silent @typechecked-files --no-strict-optional" exited with 1.

```

sockeye/training.py
artemsok: Reverted to (almost) the original step order and now manually adding validation metric to the loop that sets has_improved, since it wasn't yet available from self.state.metrics

sockeye/training.py
```diff
@@ -586,14 +588,12 @@ def fit(self,
...
-                # (3) update training metrics
-                self._update_metrics(metric_train, metric_val, process_manager)
-                metric_train.reset()
-
-                # (4) determine improvement
+                # (2) determine improvement
                 has_improved = False
                 previous_best = self.state.best_metric
-                for checkpoint, metric_dict in enumerate(self.state.metrics, 1):
+                # at this point state.self..metrics doesn't have validation results yet
```
tdomhan: nit: `self.state.metrics`

tdomhan:
Now there are a couple of pylint errors which seem to be unrelated (this could be a new version of pylint). Maybe we should fix pylint to the old version for Travis for now.
```
34.00s$ pylint --rcfile=pylintrc sockeye -E
************* Module sockeye.image_captioning.utils
sockeye/image_captioning/utils.py:194:38: E1133: Non-iterable value diff_shape is used in an iterating context (not-an-iterable)
The command "pylint --rcfile=pylintrc sockeye -E" exited with 2.
21.64s$ pylint --rcfile=pylintrc test -E
************* Module test.unit.test_loss
test/unit/test_loss.py:124:27: E1130: bad operand type for unary -: any (invalid-unary-operand-type)
test/unit/test_loss.py:131:27: E1130: bad operand type for unary -: any (invalid-unary-operand-type)
test/unit/test_loss.py:149:29: E1130: bad operand type for unary -: any (invalid-unary-operand-type)
************* Module test.unit.test_encoder
test/unit/test_encoder.py:202:11: E1101: Function 'any' has no 'all' member (no-member)
The command "pylint --rcfile=pylintrc test -E" exited with 2.
```

sockeye/training.py
```diff
@@ -1061,7 +1077,10 @@ def log_metrics(self, metrics: Dict[str, Union[float, int, mx.nd.NDArray]], chec
 
         for name, value in metrics.items():
             if isinstance(value, mx.nd.NDArray):
-                self.sw.add_histogram(tag=name, values=value, bins=100, global_step=checkpoint)
+                if mx.ndarray.contrib.isfinite(value).sum() == value.size():
```
tdomhan: when was this added to mxnet? It does not seem to be present in my version (`1.3.1`).

tdomhan: It's mentioned in the 1.4.0 changelog: https://github.com/apache/incubator-mxnet/blob/master/NEWS.md So maybe we  should use numpy for now and add a TODO to use the mxnet version with the next release.

tdomhan: Or simply add the three line function to `utils.py` for now https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/ndarray/contrib.py#L491

sockeye/training.py
```diff
@@ -586,14 +588,12 @@ def fit(self,
...
-                self._update_metrics(metric_train, metric_val, process_manager)
-                metric_train.reset()
-
-                # (4) determine improvement
+                # (2) determine improvement
                 has_improved = False
                 previous_best = self.state.best_metric
-                for checkpoint, metric_dict in enumerate(self.state.metrics, 1):
+                # at this point state.self.metrics doesn't have validation results yet
+                current_checkpoint_val_metric = {"%s-val" % name:val for name, val in metric_val.get_name_value()}
```
tdomhan: this will not actually work if the optimized metric is `BLEU` as it will only later be added to the metrics.

artemsok: I now moved the waiting for collect_result() out of the _update_metrics() to let state.metrics have the bleu-val if it's computed. Tested locally.

PR #648, opened 2019-02-26
[WIP] increase default inference batch size

mjpost:
Now that the batch size is dynamic, it seems like the default size should be increased. I chose this value as a compromise between the results below and aesthetics.

Timing test (four layer FR-EN Transformer on a 1080 Ti) on sacrebleu://mtnt1.1/test:

```
batch_size sents_per_second
50      45.3929
40      47.6237
30      45.7444
20      41.3569
10      32.0709
5       21.4622
1       7.6021
```

until you can check this box.
- [ ] Updated CHANGELOG.md


sockeye/arguments.py
```diff
@@ -1244,7 +1244,7 @@ def add_inference_args(params):
                                     'or when a finished hypothesis is in (first) position. Default: %(default)s.')
     decode_params.add_argument('--batch-size',
                                type=int_greater_or_equal(1),
-                               default=1,
+                               default=32,
```
artemsok: Use C.DEFAULT_BATCH_SIZE here as well?

tdomhan:
The dynamic batching shouldn't have a large impact for the CLI translation use case as we read the input in chunks of `500` * batch_size anyway. I believe the original motivation for a batch size of one was to have the interactive CLI use when the input comes from `stdin`. Would it make sense to make the default batch size dependent on the input source (1 for stdin and a larger batch size for an input from a file). What do you think?

mjpost:
Ah, yes, that's a good point, it needs to be 1 for `sys.stdin`. Changing the default when dealing with a file would be great, though.

mjpost:
Do you know if there's a way in `arpgarse` to tell if a variable was explicitly set via the command line? We should only override the batch size for files if the user hasn't explicitly set `--batch-size 1`, but that's also the default value.

tdomhan:
Unfortunately I do not know a better way than setting the default to `None` and then testing for that later in the file/or set to the default value.

tdomhan:
and yes, setting the batch size to something larger than `1` if the input does not come from stdin makes a lot of sense!

mjpost:
I started with this approach but it creates some problems later on. I'll figure it out.

sockeye/image_captioning/captioner.py
```diff
@@ -127,6 +127,13 @@ def caption(args: argparse.Namespace):
         check_condition(len(args.checkpoints) == len(args.models),
                         "must provide checkpoints for each model")
 
+    # If the user has not explicitly set a batch size, set it based on the file type
+    if batch_size is None:
```
fhieber: I think you are missing a `batch_size = args.batch_size` here.
It might also make sense to factor this out into a function (def `set_batch_size(args)`)to not replicate code in captioner.py and translate.py.

PR #652, opened 2019-02-28
Serialize metrics file with final BLEU for training continuation.

tdomhan:
The training state pickle did not contain the final BLEU score. When continuing training this lead to the metrics file missing the BLEU score of the checkpoint at which the training was terminated.

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -597,6 +597,8 @@ class EarlyStoppingTrainer:
                         if decoded_checkpoint > 0:
                             self.state.metrics[decoded_checkpoint - 1].update(decoder_metrics)
                             self.tflogger.log_metrics(decoder_metrics, decoded_checkpoint)
+                            utils.write_metrics_file(self.state.metrics, self.metrics_fname)
+                    # Start the decoder for the next checkpoint
                     process_manager.start_decoder(self.state.checkpoint)
 
                 # (3) determine improvement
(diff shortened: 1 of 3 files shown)


sockeye/training.py
```diff
@@ -801,7 +802,10 @@ def _cleanup(self, lr_decay_opt_states_reset: str, process_manager: Optional['De
                 decoded_checkpoint, decoder_metrics = result
                 self.state.metrics[decoded_checkpoint - 1].update(decoder_metrics)
                 self.tflogger.log_metrics(decoder_metrics, decoded_checkpoint)
-            utils.write_metrics_file(self.state.metrics, self.metrics_fname)
+
+
+                utils.write_metrics_file(self.state.metrics, self.metrics_fname)
+                self.state.save(os.path.join(self.training_state_dirname, C.TRAINING_STATE_NAME))
```
tdomhan: This makes sure that the state is written to disk after updating `self.state.metrics` so that the pickle and the `metrics` text file are in sync.

sockeye/training.py
```diff
@@ -1079,7 +1084,7 @@ def log_metrics(self, metrics: Dict[str, Union[float, int, mx.nd.NDArray]], chec
         for name, value in metrics.items():
             if isinstance(value, mx.nd.NDArray):
                 # TODO: switch to mx.ndarray.contrib.isfinite after upgrade to MxNet 1.4.*
-                if utils.isfinite(value).sum() == value.size():
+                if utils.isfinite(value).astype('int32').sum().asscalar() == value.size:
```
tdomhan: The check did not actually work as `value.size` is an integer.

sockeye/training.py
```diff
@@ -759,8 +761,7 @@ def _evaluate(self, val_iter: data_io.BaseParallelSampleIter, val_metric: mx.met
 
     def _update_metrics(self,
                         metric_train: mx.metric.EvalMetric,
-                        metric_val: mx.metric.EvalMetric,
```
tdomhan: I removed the `process_manager` here as it was no longer used.

tdomhan:
rebased + added changelog.

PR #657, opened 2019-03-06
Update to MXNet 1.4.0

fhieber:
Update to mxnet 1.4.0 after 'solving' the issue documented [here](https://github.com/apache/incubator-mxnet/issues/13862).

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
Integration and system tests pass on my laptop, but one integration test fails on Travis. Need to investigate.

fhieber:
It seems that the failing integration test is not related to the unravel_index hack we use. I checked by always using the topk function (which doesn't rely on this functionality of unravel_index) and Travis (and a GPU run of the integration tests) still fail. The GPU run fails in a different test and actually in a nondeterministic way.
This means that with mxnet 1.4.0 something has changed that makes these poorly trained integration test models behave nondeterministically. Given that mxnet changed the type of a couple of typedefs this is not too surprising.
The system tests that properly train models still pass. The question is, whether we should disable more assertions in the integration tests to make them pass on different machines reliably (@tdomhan, @davvil, @artemsok)?

tdomhan:
The integration tests on non-trained models, which produce strange outputs, have always been flaky. I think just testing that nothing crashes for the integration tests and testing the correct outputs for the system tests would be fine with me.

PR #661, opened 2019-03-12
Stop after n checkpoints

tuglat:
This change add the ability to stop training after a given number of checkpoints.
Useful for situations where one is required to share scarce GPU resources.
Added a new command line argument '--max-num-checkpoint'

- [y] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]'
until you can check this box.
- [y] Unit tests pass (`pytest`)
- [n] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?
- [y] System tests pass (`pytest test/system`)
- [y, as far as any files I touched] Passed code style checking (`./style-check.sh`)
- [y] You have considered writing a test
- [y] Updated major/minor version in `sockeye/__init__.py`. Major version bump if this is a backwards incompatible change.
- [y] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [1.18.85]
+### Fixed
+- Added parameter to force training to stop after a given number of checkpoints. Useful when forced to share limited GPU resources.
+
 ## [1.18.84]
 ### Fixed
 - Fixed lexical constraints bugs that broke batching and caused large drop in BLEU.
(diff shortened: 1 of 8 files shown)


fhieber:
Thank you for the contribution!
I am wondering, if this feature is worth adding several lines of code when it is already supported through `--max-updates` and `--checkpoint-interval`: to make training stop after n checkpoints, you set `--max-updates` to `n * checkpoint_interval`.
If we want to expose an additional argument `--max-checkpoints` and avoid leaving the above calculation to the user, we could run the calculation in the beginning internally to avoid adding more if statements in the training loop, which complicates the logic quite a bit.
Does that make sense?

tuglat:
Yes, it does. Are you proposing to make that change yourself, or should I
reimplement to work as you describe?

fhieber:
Please go ahead and implement it, if you like.

tuglat:
These changes are ready for your review. Or should I put in a new PR for the changes?

tuglat:
thanks for adding

PR #663, opened 2019-03-14
Support various parameter fixing strategies.

mjdenkowski:
This code makes it easier to fix various parameters during training by specifying a named strategy.  Rather than manually specifying each parameter name to freeze, the user can specify a single option that freezes everything except a certain part of the model (`--fixed-param-strategy NAME`).  Most of these simple strategies are taken from "Compact Personalized Models for Neural Machine Translation" ([Wuebker et al., 2018](https://arxiv.org/abs/1811.01990)).  They include:
- Updating only the decoder.
- Updating only the first and last encoder and decoder layer.
- Updating only the embeddings.
- Updating only the output projection layer.

MXNet's convention of specifying which parameters to fix (vs which to train) makes the naming a little awkward, but the options should be clear enough.  This also provides an avenue for adding other parameter fixing strategies in the future.

The log message reporting parameters during training is updated to add the number and percentage of parameters that will be fixed and learned.

until you can check this box.
- [x] Unit tests covering the new code are added.
- [x] Updated CHANGELOG.md


sockeye/training.py
```diff
@@ -67,11 +69,13 @@ def __init__(self,
                  default_bucket_key: Tuple[int, int],
                  bucketing: bool,
                  gradient_compression_params: Optional[Dict[str, Any]] = None,
-                 fixed_param_names: Optional[List[str]] = None) -> None:
+                 fixed_param_names: Optional[List[str]] = None,
+                 fixed_param_strategy: Optional[str] = None) -> None:
```
fhieber: can you add a docstring for the new parameter above?

sockeye/training.py
```diff
@@ -183,6 +193,57 @@ def sym_gen(seq_lens):
         self.save_version(self.output_dir)
         self.save_config(self.output_dir)
 
+    def _generate_fixed_param_names(self, param_names: List[str], strategy: str):
```
fhieber: Could you add a return type?

sockeye/training.py
```diff
@@ -311,13 +372,23 @@ def log_parameters(self):
...
         for name, array in sorted(arg_params.items()):
             info.append("%s: %s" % (name, array.shape))
-            total_parameters += reduce(lambda x, y: x * y, array.shape)
+            num_parameters = reduce(lambda x, y: x * y, array.shape)
+            total_parameters += num_parameters
+            if name in self.module._fixed_param_names:
+                fixed_parameters += num_parameters
+            else:
+                learned_parameters += num_parameters
+        percent_fixed = 100 * (fixed_parameters / total_parameters)
```
fhieber: should we guard against `total_parameters = 0`? Admitted, its very unlikely :)

test/unit/test_fixed_param_strategy.py
```diff
@@ -0,0 +1,213 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fhieber: 2019

sockeye/training.py
mjdenkowski: Good catch!

PR #666, opened 2019-03-20
[WIP] No dependency on sequence length as int

fhieber:
Our APIs currently require knowing the size of the sequence dimension as an integer. This is preventing us to convert our layer implementations to Gluon blocks eventually: HybridBlocks do not take integer arguments.

This is a first step to get rid of this dependency by removing the dependency on `seq_len` for RNNEncoder and CNNEncoders. 

This PR is on top of the MXNet 1.4 change for now.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/convolution.py
+++ b/sockeye/convolution.py
@@ -88,14 +88,12 @@ class ConvolutionBlock:
 
     def __call__(self,
                  data: mx.sym.Symbol,
-                 data_length: mx.sym.Symbol,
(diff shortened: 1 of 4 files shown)


fhieber:
Current problem is failing shape inference for the VariationalDropoutCell. It seems that MXNet cannot infer the shape of the batch dimension of the variational mask (created with `ones_like`) if foreach is used. Unclear how to work around it.

PR #668, opened 2019-03-25
Remove sacrebleu from sockeye. Add dependency on pip's sacrebleu

fhieber:
until you can check this box.
- [x] Updated CHANGELOG.md


mjpost:
Ah, I was halfway through doing this myself.

Would this pull down sacrebleu 1.3.0? I think with that I will introduce some backwards-incompatible method signatures, so that I can expand the set of smoothing techniques that are available (https://github.com/mjpost/sacreBLEU/tree/smoothing). I could submit the Sockeye changes, but want to make sure it won't accidentally break.

fhieber:
Good point @mjpost. It probably makes sense to fix the version in the requirements for now to `<1.3.0`. We can then consciously update to 1.3.0 by making the corresponding changes in Sockeye.

PR #675, opened 2019-04-06
Convert all transformer-relevant layers to Gluon HybridBlocks.

fhieber:
This is a rather large no-op change :)
Specifically, it converts all Transformer-related layer implementations (Layer Normalization, LHUC, TransformerFeedForward, TransformerProcess, MultiHead(Self)Attention) into Gluon HybridBlocks, even though they are not yet used as Gluon Blocks. 

Since one can call HybridBlocks with Symbols as input and receives Symbols, technically nothing changes for now using our Module-based training and inference.

I checked that the parameter names created from the symbol blocks are the same as before (the logic with how prefixes are passed onto child blocks is a bit smarter in Gluon). 

Some minor hacks required nevertheless:
- We use LayerNormalization also for RNN cells that may pass in an existing parameter Variable. For these cases I had to branch on whether the HybridBlock itself creates a parameter or its already passed in. This complicated logic can be removed if we move to Gluon at some point.
- Gluon HybridBlocks have some involved way of passing inputs (and parameters) through when `__call__` is called, passing to `forward()`, passing to `hybrid_forward`. I had to remove keyword-based argument passing in a couple of places to make this work.

This should be safe to merge without bumping any major version, but is a first step towards Gluon-based Sockeye.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/encoder.py
+++ b/sockeye/encoder.py
@@ -1013,7 +1013,7 @@ class ConvolutionalEncoder(Encoder):
         return self.config.cnn_config.num_hidden
 
 
-class TransformerEncoder(Encoder):
+class TransformerEncoder(Encoder, mx.gluon.HybridBlock):
     """
     Non-recurrent encoder based on the transformer architecture in:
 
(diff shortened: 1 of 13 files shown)


sockeye/encoder.py
```diff
@@ -1008,19 +1027,44 @@ class TransformerEncoder(Encoder):
...
+                                                                            name="bias")
+            self.final_process = transformer.TransformerProcessBlock(sequence=config.preprocess_sequence,
+                                                                     dropout=config.dropout_prepost,
+                                                                     prefix="final_process_")
+
+    def hybrid_forward(self, F, data, data_length):
+        return self._encode(F, data, data_length)
+
+    def _encode(self, F, data: mx.sym.Symbol, data_length: mx.sym.Symbol) -> mx.sym.Symbol:
+
```
artemsok: also nit: replace the vertical space with a doc string?

sockeye/encoder.py
```diff
@@ -1008,19 +1027,44 @@ class TransformerEncoder(Encoder):
...
+        data = utils.cast_conditionally(F, data, self.dtype)
+        if self.config.dropout_prepost > 0.0:
+            data = F.Dropout(data=data, p=self.config.dropout_prepost)
+
+        # (batch_size * heads, 1, seq_len)
+        bias = F.expand_dims(self.valid_length_mask(data, data_length), axis=1)
+        bias = utils.cast_conditionally(F, bias, self.dtype)
+        for layer in self.layers:
```
artemsok: is it because of the bias that we are not calling self.layers.forward() here instead of the explicit loop?

sockeye/encoder.py
fhieber: removed the vertical space :)

sockeye/encoder.py
fhieber: yes, HybridSequential only supports a single input for forward calls.

PR #690, opened 2019-05-03
Update to pyyaml 5.1

fhieber:
This should remove the github security warnings. As we need to load custom objects using yaml, we are bound to use the `UnsafeLoader`.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/config.py
```diff
@@ -119,7 +121,7 @@ def load(fname: str) -> 'Config':
         :return: Configuration.
         """
         with open(fname) as inp:
-            obj = yaml.load(inp)
+            obj = yaml.load(inp, Loader=yaml.UnsafeLoader)  # type: ignore
```
davvil: Would it be better to use yaml_loader here?

fhieber: can you elaborate? I am not familiar with yaml that much.

PR #698, opened 2019-06-20
Updated the install instructions

rherbrich74:
(description of the change)

until you can check this box.
- [ ] Updated CHANGELOG.md


docs/setup.md
```diff
@@ -51,16 +51,27 @@ where `${CUDA_VERSION}` can be `80` (8.0), `90` (9.0), `92` (9.2), or `100` (10.
 
 Developers will be better served by pointing `$PYTHONPATH` to the root of the git-cloned source.
 
-### → on AWS...
+### → on AWS and Azure ...
```
tdomhan: As this really mainly about Anaconda now and not about either Azure or AWS: How about we call this section:
```
### in an Anaconda environment
```
and then 
```
In an Anaconda environment such as the one provided by the [AWS DeepLearning AMI](https://aws.amazon.com/amazon-ai/amis/) users only need to run the following line to install sockeye (on an instance without a GPU):
```

docs/setup.md
```diff
@@ -51,16 +51,27 @@ where `${CUDA_VERSION}` can be `80` (8.0), `90` (9.0), `92` (9.2), or `100` (10.
...
-For other environments, you can choose between installing via pip or directly from source. Note that for the
-remaining instructions to work you will need to use `python3` instead of `python` and `pip3` instead of `pip`.
+On an instance with a GPU, the following commands will work
+
+```bash
+> conda create -n sockeye python=3.6
+> source activate sockeye
+> wget https://raw.githubusercontent.com/awslabs/sockeye/master/requirements/requirements.gpu-cu100.txt
+> pip install sockeye --no-deps -r requirements.gpu-cu100.txt
+rm requirements.gpu-cu100.txt
```
tdomhan: Could you replace this with `${CUDA_VERSION}`?

docs/setup.md
```diff
@@ -51,16 +51,27 @@ where `${CUDA_VERSION}` can be `80` (8.0), `90` (9.0), `92` (9.2), or `100` (10.
...
+```
+
+For other environments, you can choose between installing via pip or directly from source. This will also work with Azure when using the [Data Science Virtual Machine](http://aka.ms/dsvm/discover) image.
```
tdomhan: So Data Science Virtual Machine doesn't come with Anaconda and therefore user can install either via pip or from source?

docs/setup.md
rherbrich74: I did not check with all four versions. Shall I still specify all four or just the one that worked?

docs/setup.md
rherbrich74: I like this. Shall I make this change and re-submit?

docs/setup.md
rherbrich74: It comes with Anaconda; I suggest we remove the first sentence. Agreed?

rherbrich74:
I made the changes as suggested and pushed again to rherbrich74/sockeye. Anything else I need to do?

docs/setup.md
tdomhan: I'd use the `CUDA_VERSION` environment variable and add a sentence as done in the pip section:
```
where ${CUDA_VERSION} can be 80 (8.0), 90 (9.0), 92 (9.2), or 100 (10.0).
```

Or maybe just once outside the pip and anaconda sections mention that `CUDA_VERSION` in the following refers to the cuda version present on the system.

docs/setup.md
tdomhan: I'd use the `CUDA_VERSION` environment variable and add a sentence as done in the pip section:
```
where ${CUDA_VERSION} can be 80 (8.0), 90 (9.0), 92 (9.2), or 100 (10.0).
```

Or maybe just once outside the pip and anaconda sections mention that `CUDA_VERSION` in the following refers to the cuda version present on the system.

tdomhan:
Thanks! Looks good. I will merge now.

Issue #684, opened 2019-04-23
Problem with multiprocessing context


adamkusmirek:
I am currently experiencing the following issue when running `sockeye-train`:
```
[INFO:sockeye.training] Training started.
[INFO:sockeye.utils] Releasing GPU 0.
[ERROR:root] Uncaught exception
Traceback (most recent call last):
  File "/usr/local/bin/sockeye-train", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/sockeye/train.py", line 825, in main
    train(args)
  File "/usr/local/lib/python3.6/dist-packages/sockeye/train.py", line 955, in train
    existing_parameters=args.params)
  File "/usr/local/lib/python3.6/dist-packages/sockeye/training.py", line 627, in fit
    process_manager = DecoderProcessManager(self.model.output_dir, decoder=decoder)
  File "/usr/local/lib/python3.6/dist-packages/sockeye/training.py", line 1282, in __init__
    self.ctx = mp_utils.get_context()  # type: ignore
  File "/usr/local/lib/python3.6/dist-packages/sockeye/multiprocessing_utils.py", line 71, in get_context
    assert __context is not None, ("Multiprocessing context not initialized. Please call "
AssertionError: Multiprocessing context not initialized. Please call sockeye.multiprocessing_utils.initialize() right after interpreter startup.
```
Any idea what could be causing this?

I am running sockeye using mxnet-cu100mkl inside a singularity container.

fhieber:
Thanks for reporting. I am not sure what the problem is. `sockeye.train` calls `mp.initialize()` right at start up. Are you using that module somehow in your own script? What is a singularity container?

adamkusmirek:
> Thanks for reporting. I am not sure what the problem is. `sockeye.train` calls `mp.initialize()` right at start up. 

That's what confused me too after having a look at the source.

> Are you using that module somehow in your own script?
I am calling the module inside a bash script:
```
sockeye-train -d ${TRAIN_DATA_DIR} \
              -vs ${SRC_CORP} \
              -vt ${TGT_CORP} \
              --encoder rnn \
              --decoder rnn \
              --num-embed 256 \
              --rnn-num_hidden 512 \
              --rnn-attention-type dot \
              --max-seq-len 60 \
              --decode-and-evaluate 500 \
              -o ${MODEL_DIR}
```
Basically I have been following the WMT tutorial.

> What is a singularity container?

[...]

PR #704, opened 2019-07-07
Fix:Always initialize the multiprocessing context.

tdomhan:
With this change we will always initialize the multiprocessing context to address https://github.com/awslabs/sockeye/issues/684. As the multiprocessing context initialization code checks whether it has been called before and if not simply returns this should not change existing behavior.


fhieber:
I think you commited src/recommonmark by accident

PR #711, opened 2019-08-05
Sockeye 2 cpdecoder merged

tdomhan:
(description of the change)


sockeye/checkpoint_decoder.py
```diff
@@ -73,8 +78,8 @@ def __init__(self,
                  max_output_length_num_stds: int = C.DEFAULT_NUM_STD_MAX_OUTPUT_LENGTH,
                  ensemble_mode: str = 'linear',
                  sample_size: int = -1,
-                 random_seed: int = 42) -> None:
-        self.context = context
+                 random_seed: int = 42,
+                 model_folder: str = '') -> None:
```
fhieber: do we want to default to an empty string?

sockeye/checkpoint_decoder.py
```diff
@@ -39,29 +40,33 @@ class CheckpointDecoder:
...
 
     def __init__(self,
-                 context: mx.context.Context,
                  inputs: List[str],
                  references: str,
-                 model: str,
+                 source_vocabs: List[vocab.Vocab],
+                 target_vocab: vocab.Vocab,
+                 model: sockeye.model.SockeyeModel,
+                 contexts: List[mx.Context],
```
fhieber: Could we use `Union[mx.Context, List[mx.Context]]`?

sockeye/checkpoint_decoder.py
```diff
@@ -109,44 +114,41 @@ def __init__(self,
...
+            length_penalty=inference.LengthPenalty(self.length_penalty_alpha, self.length_penalty_beta),
+            brevity_penalty=inference.BrevityPenalty(weight=0.0),
+            beam_prune=0.0,
+            beam_search_stop='all',
+            nbest_size=self.nbest_size,
+            models=[self.model],
+            source_vocabs=source_vocabs,
+            target_vocab=target_vocab,
+            restrict_lexicon=None,
+            store_beam=False)
```
fhieber: would it make sense to expose hybridize here (and also in the CheckpointDecoder class)?

sockeye/decoder.py
```diff
@@ -159,30 +162,31 @@ def init_state_from_encoder(self,
...
         :param encoder_outputs: Encoder outputs. Shape: (batch, source_length, encoder_dim).
         :param encoder_valid_length: Valid lengths of encoder outputs. Shape: (batch,).
-        :param is_inference: Whether to return states for inference or for training.
         :return: Initial states.
         """
         source_mask = self.valid_length_mask(encoder_outputs, encoder_valid_length)
 
-        if is_inference:
+        # (batch_size, 1)
+        step = mx.nd.expand_dims(mx.nd.zeros_like(encoder_valid_length), axis=1)
```
fhieber: Why the shape change from (1,) to (1,1)?

sockeye/decoder.py
```diff
@@ -219,53 +224,66 @@ def forward(self, step_input, states):
...
 
         # run decoder op
         target, self_attention_key_values = super().forward(step_input, states)
 
         if is_inference:
-            # During inference, length dimension of decoder output has size 1, remove it
+            # During inference, length dimension of decoder output has size 1, squeeze it
             # (batch, num_hidden)
-            target = target.reshape((-1, self.config.model_size))
+            target = mx.nd.reshape(target, shape=(-1, self.get_num_hidden()))
```
fhieber: do we still want that interface method?

sockeye/inference.py
```diff
@@ -820,12 +820,12 @@ def __init__(self,
                  ensemble_mode: str,
                  length_penalty: LengthPenalty,
                  batch_size: int,
-                 beam_size: int,
                  beam_prune: float,
                  beam_search_stop: str,
                  models: List[SockeyeModel],
                  source_vocabs: List[vocab.Vocab],
                  target_vocab: vocab.Vocab,
+                 beam_size: int = 5,
```
fhieber: Can we remove the default value here again and explicitly create it on CheckpointDecoder with beam_size=5? I like the user to be forced to think about beam size when constructing a Translator.

sockeye/layers.py
```diff
@@ -659,7 +659,7 @@ def hybrid_forward(self, F, data, steps, weight):  # pylint: disable=arguments-d
         Applies positional embeddings to input data.
 
         :param data: Input data. Shape: (batch, length or 1, num_embed)
-        :param steps: Optional steps input. If given, shape is (batch,)
+        :param steps: Optional steps input. If given, shape is (batch_size or 1, seq_len,)
```
fhieber: parentheses missing?

sockeye/model.py
```diff
@@ -96,10 +96,11 @@ class SockeyeModel(mx.gluon.Block):
     time.
 
     :param config: Model configuration.
+    :param inference_only: Use the model only for inference, enabling optimizations.
     :param prefix: Name prefix for all parameters of this model.
     """
 
-    def __init__(self, config: ModelConfig, prefix: str = '', **kwargs) -> None:
+    def __init__(self, config: ModelConfig, inference_only: bool = False, prefix: str = '', **kwargs) -> None:
```
fhieber: thats a good addition. Can we move other optimization into here and bind them to the setting of that flag? For example `config.disable_dropout()`?

sockeye/train.py
```diff
@@ -815,7 +816,7 @@ def train(args: argparse.Namespace) -> training.TrainState:
                                       update_on_kvstore=None)
         losses = create_losses(args)
 
-        hybridize = True
+        hybridize = False
```
fhieber: set to True again

sockeye/train.py
```diff
@@ -828,9 +829,12 @@ def train(args: argparse.Namespace) -> training.TrainState:
             loss_functions=losses,
             context=context,
             dtype=args.dtype
-        )
+        )        
+        
+        cp_decoder = create_checkpoint_decoder(args, exit_stack, context,
+                                            training_model, source_vocabs, target_vocab)
```
fhieber: indendation is off

sockeye/training.py
```diff
@@ -24,11 +24,12 @@
 from math import sqrt
 from typing import Callable, Dict, List, Optional, Iterable, Tuple, Union
 
+import gluonnlp
```
fhieber: no longer used

sockeye/training.py
```diff
@@ -24,11 +24,12 @@
...
 import mxnet as mx
 import numpy as np
+from mxnet import gluon
```
fhieber: needed?

sockeye/training.py
```diff
@@ -203,6 +204,10 @@ def fit(self,
 
             if self.state.updates > 0 and self.state.batches % (
                     self.config.checkpoint_interval * self.config.update_interval) == 0:
+                # ############
```
fhieber: weird formatting. Do we need it? :)

sockeye/training.py
```diff
@@ -337,6 +352,7 @@ def _determine_improvement(self, val_metrics: List[loss.LossMetric]) -> bool:
                     self.state.best_checkpoint = self.state.checkpoint
                     self.state.num_not_improved = 0
                     return True
+        assert value is not None, "Early stoppin metric %s not found in validation metrics." % self.config.early_stopping_metric
```
fhieber: typo: stopping

sockeye/inference.py
tdomhan: Why? I think it would be good to simply have a reasonable default with the option to change it. More often than not we don't want to change the beam size, no?

fhieber: If we force the user to think about less important parameters such as beam_prune and beam_search_stop, I think we can ask her/him for a reasonable beam_size :)

sockeye/checkpoint_decoder.py
tdomhan: sure, i'll expose this.

sockeye/decoder.py
tdomhan: because I wanted the graph to be the same for training and decoding. So the step always has the time dimension now.

sockeye/decoder.py
tdomhan: why not?

sockeye/inference.py
tdomhan: maybe we should add more defaults also for those parameters. :)

sockeye/layers.py
tdomhan: is it? So the first dimension is either be 1 or batch_size. Anyway, this is maybe confusing.

sockeye/model.py
tdomhan: yeah, makes sense to me.

sockeye/training.py
tdomhan: ups, come back with the merge I think.

sockeye/training.py
tdomhan: we have this similarly in Sockeye 1: 
https://github.com/awslabs/sockeye/blob/master/sockeye/training.py#L698

Anyway, we can remove it. Just wanted to point out the location where checkpointing is happening.

sockeye/decoder.py
fhieber: ah I see, thanks!

sockeye/inference.py
fhieber: Ok, but then we need to keep them in sync with the default values in arguments.py, which I find tedious :)
It's fine for now, but we should do a pass over the code for such small things at some point.

sockeye/train.py
fhieber: change still needed

tdomhan:
I merged in the latest version of the branch and added a hybridization CLI flag.

fhieber:
Let's wait for Travis and then merge

PR #712, opened 2019-08-06
Sockeye 2 Training Update

mjdenkowski:
This merges in the most stable parts of the "sockeye_2_training" branch.  Major updates include:

#### Distributed Training
- Sockeye now supports multi-instance training with Horovod and OpenMPI (see requirements.horovod.txt or use the Docker build).
- Use `horovodrun` and the `--horovod` flag to activate distributed mode, where multiple workers (copies of sockeye.train) run in parallel and automatically synchronize updates.
- Each worker loads a different slice of each data shard.
- File management and decisions like learning rate decay are also synchronized.
- Sample commands for running distributed training are included in the Docker readme.

#### Training Options
- Optimizers have been trimmed down to Adam and SGD.
- Learning rate schedulers have been trimmed down to `plateau-reduce`, `inv-sqrt-decay`, and `linear-decay`.
- Time scale can be specified for decay-based schedulers.
- Users must now specify a stopping criteria and training can always be continued with a different criteria.
- Basic AMP support is included (recipe from the AMP tutorial).  Sockeye is not yet able to take full advantage of AMP (issues with SoftmaxOutput among others).

#### Docker Build
- Sockeye now has a Docker build to simplify running, especially useful for distributed training.
- The Python build script creates an image of the current code tagged with the latest commit.  Code changes are recognized (users can rebuild without a new commit).
- The Dockerfile is a recipe for downloading and building software.  We do not distribute any of the referenced files (other than Sockeye).

#### Misc
- A script for plotting Sockeye's "metrics" files is included under sockeye_contrib.  This can be used to visualize different aspects of training runs, including derived data like time since last improvement or average improvement over a given window.
- The seldom-used Autopilot module is removed.

until you can check this box.
- [X] System tests pass (`pytest test/system`) (as many as other sockeye_2* branches)
- [X] Updated CHANGELOG.md


sockeye/train.py
```diff
@@ -607,9 +625,8 @@ def create_optimizer_config(args: argparse.Namespace) -> OptimizerConfig:
     config.set_lr_scheduler(lr_sched)
     logger.info("Optimizer: %s | kvstore=%s | params=%s | initializer=%s",
                 config.name, config.kvstore, config.params, config.initializer)
-    if args.update_interval > 1:
-        logger.info("Gradient accumulation over %d batches. Effective batch size: %d",
-                    args.update_interval, effective_batch_size)
+    logger.info("Gradient accumulation over %d batch(es) by %d worker(s). Effective batch size: %d",
+                args.update_interval, num_workers, effective_batch_size)
     return config
```
mjpost: You all should consider requiring Python 3.7+ for Sockeye 2, just for the f-strings.

fhieber:
Thanks Michael, I will take a look soon. I find the PR description too good to be forgotten. Would it make sense to already start an updated README.md and CHANGELOG.md, using the description for Distributed Training etc. you have above?

CHANGELOG.md
```diff
@@ -11,15 +11,28 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
 ## [2.0.0]
+
 ### Changed
```
fhieber: I think it would be nice to use the PR descriptions on Distributed Training, Training Options, Docker build, Misc here as well with a nested list. We will need some 'release notes' document anyway.

sockeye/arguments.py
```diff
@@ -854,27 +845,21 @@ def add_training_args(params):
...
                               help='Learning rate scheduler type. Default: %(default)s.')
+    train_params.add_argument('--learning-rate-t-scale',
+                              type=float,
+                              default=1.0,
+                              help="Step number is multiplied by this value when determining learning rate for the "
+                                   "current step. Default: %(default)s.")
     train_params.add_argument('--learning-rate-reduce-factor',
                               type=float,
-                              default=0.7,
+                              default=0.9,
```
fhieber: Do we want to update the default arguments here? Users might have sockeye.train scripts with this parameter not specified and we don't want to change the overall behavior too much.

sockeye/arguments.py
```diff
@@ -915,7 +900,7 @@ def add_training_args(params):
 
     train_params.add_argument('--seed',
                               type=int,
-                              default=13,
+                              default=1,
```
fhieber: Same here, I would suggest going back to `13`.

sockeye/constants.py
```diff
@@ -40,13 +40,7 @@
 EMBEDDING_PREFIX = "embed_"
 ATTENTION_PREFIX = "att_"
 COVERAGE_PREFIX = "cov_"
-BIDIRECTIONALRNN_PREFIX = ENCODER_PREFIX + "birnn_"
```
fhieber: thanks a lot for cleaning up various unused constants!

sockeye/lr_scheduler.py
```diff
@@ -62,132 +60,63 @@ def new_evaluation_result(self, has_improved: bool) -> bool:
...
 
-    :param updates_per_checkpoint: Number of batches between checkpoints.
-    :param half_life: Half life of the learning rate in number of checkpoints.
-    """
-
-    def __init__(self, updates_per_checkpoint: int, half_life: int, warmup: int = 0) -> None:
-        super().__init__(warmup)
-        check_condition(updates_per_checkpoint > 0, "updates_per_checkpoint needs to be > 0.")
-        check_condition(half_life > 0, "half_life needs to be > 0.")
+    This is the schedule used by Devlin et al. in the BERT paper
```
fhieber: thanks for adding!

sockeye/lr_scheduler.py
```diff
@@ -201,7 +130,7 @@ class LearningRateSchedulerPlateauReduce(AdaptiveLearningRateScheduler):
 
     def __init__(self, reduce_factor: float, reduce_num_not_improved: int, warmup: int = 0) -> None:
         super().__init__(warmup)
-        check_condition(0.0 < reduce_factor <= 1, "reduce_factor should be in ]0,1].")
+        check_condition(0.0 < reduce_factor < 1, "reduce_factor should be between (0, 1).")
```
fhieber: good catch :)

sockeye/lr_scheduler.py
```diff
@@ -251,49 +180,47 @@ def __repr__(self):
...
 
 def get_lr_scheduler(scheduler_type: str,
-                     updates_per_checkpoint: int,
-                     learning_rate_half_life: int,
+                     learning_rate_t_scale: float,
                      learning_rate_reduce_factor: float,
                      learning_rate_reduce_num_not_improved: int,
-                     learning_rate_schedule: Optional[List[Tuple[float, int]]] = None,
-                     learning_rate_warmup: Optional[int] = 0) -> Optional[LearningRateScheduler]:
+                     learning_rate_warmup: Optional[int] = 0,
```
fhieber: `Optional` still required?

sockeye/train.py
fhieber: Yes they are nice, but I think there is still value in supporting Python 3.6

sockeye/train.py
```diff
@@ -689,11 +705,39 @@ def train(args: argparse.Namespace) -> training.TrainState:
...
+        using_horovod = True
+        horovod_mpi.hvd.init()
+        # Each worker uses a separate output directory.  The primary worker
+        # (rank 0) writes files to the root of the output directory (standard
+        # behavior).  Secondary workers write files to rank-named
+        # sub-directories.
+        if horovod_mpi.hvd.rank() > 0:
+            args.output = os.path.join(args.output, C.HOROVOD_SECONDARY_WORKERS_DIRNAME, str(horovod_mpi.hvd.rank()))
+            # Do not keep extensive checkpoint histories for secondary workers
+            args.keep_last_params = 1
```
fhieber: nice!

sockeye/train.py
```diff
@@ -751,6 +796,14 @@ def train(args: argparse.Namespace) -> training.TrainState:
 
         training_model = model.SockeyeModel(model_config)
 
+        # Stopping criteria
```
fhieber: would this be better-placed inside `check_arg_compatibility`?

sockeye/train.py
```diff
@@ -767,35 +820,36 @@ def train(args: argparse.Namespace) -> training.TrainState:
...
-            trainer_config.max_epochs = None
-
         optimizer_config = create_optimizer_config(args)
         training_model.initialize(optimizer_config.initializer, ctx=context)
         if args.params is not None:  # load existing parameters if present
             training_model.load_parameters(filename=args.params,
                                            ctx=context,
-                                           allow_missing=args.allow_missing_params or model_config.lhuc)
+                                           allow_missing=args.allow_missing_params or model_config.lhuc,
+                                           cast_dtype=True)
```
fhieber: could you also make explicit the other argument (`dtype_source`) here?

sockeye/train.py
```diff
@@ -767,35 +820,36 @@ def train(args: argparse.Namespace) -> training.TrainState:
...
+        # across all workers by broadcasting worker 0's values.  This is not
+        # required when resuming training as synchronized training states
+        # already exist.
+        if using_horovod and not resume_training:
+            for ctx in context:
+                with mx.Context(ctx):
+                    horovod_mpi.hvd.broadcast_parameters(params, root_rank=0)
+
         if args.dtype == C.DTYPE_FP16:
             training_model.cast(C.DTYPE_FP16)
```
fhieber: I don't know much about AMP. Should one skip casting when using amp?

sockeye/training.py
```diff
@@ -328,9 +337,28 @@ def _determine_improvement(self, val_metrics: List[loss.LossMetric]) -> bool:
...
+                # check of whether metric value has improved and broadcasts the
+                # result to secondary workers.  Non-determinism in the order of
+                # GPU operations can lead to slight numeric variation across
+                # workers, causing potential desync if each worker makes its own
+                # check for key training decisions (reducing learning rate,
+                # early stopping, etc.).
+                if self.config.using_horovod and horovod_mpi.hvd.rank() > 0:
+                    # Horovod secondary workers: wait for primary worker to send
+                    # result.
+                    value_is_better = None  # type: bool
```
fhieber: shouldn't this be `# type: Optional[bool]`?

sockeye/training.py
```diff
@@ -328,9 +337,28 @@ def _determine_improvement(self, val_metrics: List[loss.LossMetric]) -> bool:
...
-                                                self.state.best_metric,
-                                                self.config.early_stopping_metric):
+                # When using Horovod, the primary worker makes an authoritative
+                # check of whether metric value has improved and broadcasts the
+                # result to secondary workers.  Non-determinism in the order of
+                # GPU operations can lead to slight numeric variation across
+                # workers, causing potential desync if each worker makes its own
+                # check for key training decisions (reducing learning rate,
+                # early stopping, etc.).
+                if self.config.using_horovod and horovod_mpi.hvd.rank() > 0:
```
fhieber: For code clarity, would it make sense to define these checks in a separate method, something along the lines of `self.isMaster()` and `self.isWorker()`?

sockeye_contrib/docker/README.md
```diff
@@ -0,0 +1,65 @@
+# Sockeye Docker Image
```
fhieber: Thanks for adding this !

CHANGELOG.md
mjdenkowski: Expanding the changelog is a good idea.  We've kept it minimal to avoid clutter, but anyone actually reading it is probably interested in the details.

sockeye/arguments.py
mjdenkowski: This could be considered part of the training update and general 2.0 release.  Since we're already not going to be backward compatible, we have a chance to update our default training recipe to the best current settings.

We can also give users some sample training commands for 1 GPU, multiple GPUs, and distributed training in the updated readme.

sockeye/arguments.py
mjdenkowski: The idea is that `1` would be more intuitive and perhaps `0` would be even better.  Again since we aren't backward compatible anyway, we can pick defaults that are either stronger or more intuitive.

sockeye/lr_scheduler.py
mjdenkowski: Good catch

sockeye/train.py
mjdenkowski: Good idea

sockeye/train.py
mjdenkowski: Are you referring to passing `dtype='current'`?

sockeye/train.py
mjdenkowski: The [sample AMP recipe](https://mxnet.incubator.apache.org/versions/master/tutorials/amp/amp_tutorial.html) doesn't involve any casting, so AMP should take care of data types automatically.  Sockeye would generally run with either `--amp` or `--dtype float16` but not both.

That said, we still aren't 100% sure AMP is working properly due to SoftmaxOutput and only include it for testing.  We could drop it from this update and keep it in the sockeye_2_training branch.

sockeye/training.py
mjdenkowski: Good catch

sockeye/training.py
mjdenkowski: That's a good idea.  We can add those convenience checks to the `horovod_mpi` module.

sockeye/arguments.py
fhieber: ok, as long as system tests still pass, we'd should be fine here.

sockeye/train.py
fhieber: yes `dtype_source='current'` instead of `saved', meaning that it will cast the on-disk parameter to the precision of the Model block and not vice versa.

sockeye/train.py
fhieber: Thanks for the explanation! Let's just add a TODO here that we might need to condition casting on args.dtype AND args.amp

sockeye/training.py
mjdenkowski: I found another way to refactor this.  Adding a dedicated `using_horovod()` function to the `horovod_mpi` module means we don't have to pass around `using_horovod` values and can always check directly with the source (all things related to Horovod or MPI go through that optional module).

Taking another look, I do like explicitly making two checks: whether Horovod is running and what type of worker the current process is.

What do you think?

fhieber:
I am going to merge this to sockeye_2 for now. We can resolve the remaining things in a separate change.

PR #731, opened 2019-10-01
Sockeye 2 beam search fix

tdomhan:
This fixes an issue with beam search introduced with the beam search refactoring (and reenables the refactoring..


sockeye/layers.py
```diff
@@ -456,7 +456,7 @@ def hybrid_forward(self, F,
             updated_keys = F.concat(previous_keys, keys, dim=1)
             keys = _remove_first_step(F, updated_keys)
 
-        updated_values = keys
+        updated_values = values
```
tdomhan: this bug didn't actually cause any harm because `previous_values` is not `None` in practice.

sockeye/beam_search.py
```diff
@@ -674,7 +674,7 @@ def forward(self,
                 break
 
             # (5) update models' state with winning hypotheses (ascending)
-            _sort_states(model_states, best_hyp_indices)
+            model_states = _sort_states(model_states, best_hyp_indices)
```
tdomhan: This was the bug.

tdomhan:
For me reranking tests are failing, but this was already the case before the change. This is likely due to my setup. I'll look deeper into this.

sockeye/beam_search.py
fhieber: 👏👏 great catch! Thanks!

fhieber: 👏👏 great catch! Thanks!

tdomhan: Took a while to debug because I comparing the model states by taking the sum over ndarrays which matched even when entries were reshuffled. Anyway, good to have the refactoring back. Having beam search separated from the rest of inference looks much cleaner!

PR #737, opened 2019-10-24
Test selection through testpaths.

tdomhan:
Small modification to pytest. By moving the directory subselection to `testpaths` we still only run the unit and integration tests as before. The advantage is that one can run single tests in IDEs  instead of having to always run all unit and integration tests.


fhieber:
Useful, thanks!

PR #747, opened 2019-11-11
Set ParallelModel threads to daemons, nicer parameter printing, re-enable tensorboard logging

fhieber:
- Re-enables tensorboard logging
- nicer parameter logging for `sockeye.train`
- make ParallelModel threads daemons, causing them to stop once main thread exits (e.g. via KeyboardInterrupt).

until you can check this box.

### DIFF
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -22,6 +22,7 @@ import itertools
 import logging
 import math
 import os
+import pprint
 import random
 import shutil
 import subprocess
@@ -764,6 +765,18 @@ def split(data: mx.nd.NDArray,
     return ndarray_or_list
 
 
+_DTYPE_TO_STRING = {
+    np.float32: 'float32',
+    np.float16: 'float16',
+    np.int8: 'int8',
+    np.int32: 'int32'
+}
+
+
+def _print_dtype(dtype):
+    return _DTYPE_TO_STRING.get(dtype, str(dtype))
+
+
 def log_parameters(params: mx.gluon.ParameterDict):
     """
     Logs information about model parameters.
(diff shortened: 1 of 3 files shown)


sockeye/utils.py
```diff
@@ -774,16 +787,15 @@ def log_parameters(params: mx.gluon.ParameterDict):
...
             learned_parameter_names.append(repr)
     #percent_fixed = 100 * (fixed_parameters / max(1, total_parameters))
     #percent_learned = 100 * (learned_parameters / max(1, total_parameters))
-    logger.info("Trainable parameters: %s", ", ".join(learned_parameter_names))
-    logger.info("Fixed model parameters: %s", ", ".join(fixed_parameter_names))
+    logger.info("Trainable parameters:\n%s", pprint.pformat(learned_parameter_names))
+    logger.info("Fixed model parameters:\n%s", pprint.pformat(fixed_parameter_names))
     #logger.info("Fixing %d parameters (%0.2f%%)", fixed_parameters, percent_fixed)
     #logger.info("Learning %d parameters (%0.2f%%)", learned_parameters, percent_learned)
     #logger.info("Total # of parameters: %d", total_parameters)
```
mjdenkowski: Great update!  While we're doing work here, can we reenable printing these parameter totals?

fhieber: I'd love to, the problem is deferred initialization of some parameters so that the shape of these is still `0` in some dimensions at this point. Their actual size can only be known after a first batch of data has been passed to the model. Doing this here would be pretty ugly. Do you have a better idea?

mjdenkowski: That's a good reason not to report parameters here.  We could move the report to the first batch or checkpoint with a label the user could grep for.  That's probably out of scope for this commit.

PR #752, opened 2019-11-14
[WIP] Sockeye 2 Performance Optimizations

blchu:
Made changes to Sockeye 2 to improve the performance of the Transformer model in machine translation. Current changes only apply to inference; optimizations to training are planned for later but before completion of the pull request.

A list of changes detailed below:
1. Replaced the `batch_dot` ops in multihead attention with ops that do not require folding heads in with batch dimension; one caveat is that batch and sequence_length dimensions are swapped, requiring some adjustments to other parts of the code to account for the change
2. Removed the `take` ops that were applied to the encoder states, as they do not change on different beams; this effectively cuts compute time for these `take`s in half
3. Gathered the input token ids into a numpy array on CPU before sending them all to the GPU at the beginning of beam search, rather than sending each batch element to the GPU one at a time
4. Set the data type of arrays during beam search computation to match the model's data type, rather than explicitly setting it to fp32

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -31,6 +31,10 @@ logger = logging.getLogger(__name__)
 class _Inference(ABC):
 
     @abstractmethod
+    def state_structure(self):
+        raise NotImplementedError()
+
+    @abstractmethod
     def encode_and_initialize(self,
                               inputs: mx.nd.NDArray,
                               valid_length: Optional[mx.nd.NDArray] = None):
@@ -91,7 +95,7 @@ class _EnsembleInference(_Inference):
     def state_structure(self) -> str:
         structure = ''
         for model in self._models:
-            structure.append(model.state_structure())
+            structure += model.state_structure()
         return structure
 
     def encode_and_initialize(self, inputs: mx.nd.NDArray, valid_length: Optional[mx.nd.NDArray] = None):
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -67,7 +67,7 @@ class _SingleModelInference(_Inference):
         # TODO: change back to casting first when fp32 safe accumulation is fixed
         if not self._skip_softmax:
             logits = logits.log_softmax(axis=-1)
-        scores = -logits.astype('float32', copy=False)
+        scores = -logits
         return scores, states
 
 
@@ -104,7 +104,6 @@ class _EnsembleInference(_Inference):
         outputs, new_states = [], []
         for model, model_states in zip(self._models, states):
             logits, model_states, _ = model.decode_step(step_input, model_states, vocab_slice_ids)
-            logits = logits.astype('float32', copy=False)
             probs = logits.softmax(axis=-1)
             outputs.append(probs)
             new_states.append(model_states)
@@ -276,11 +275,13 @@ class SortNormalizeAndUpdateFinished(mx.gluon.HybridBlock):
     """
 
     def __init__(self,
+                 dtype: str,
                  pad_id: int,
                  eos_id: int,
                  scorer: CandidateScorer,
                  **kwargs) -> None:
         super().__init__(**kwargs)
+        self.dtype = dtype
         self.pad_id = pad_id
         self.eos_id = eos_id
         self._scorer = scorer
@@ -298,7 +299,7 @@ class SortNormalizeAndUpdateFinished(mx.gluon.HybridBlock):
         newly_finished = F.broadcast_logical_xor(all_finished, finished)
         scores_accumulated = F.where(newly_finished,
                                      self._scorer(scores_accumulated,
-                                                  F.cast(F.expand_dims(lengths, axis=1), 'float32'),
+                                                  F.cast(F.expand_dims(lengths, axis=1), self.dtype),
                                                   reference_lengths),
                                      scores_accumulated)
 
@@ -434,6 +435,7 @@ class BeamSearch(mx.gluon.Block):
 
     def __init__(self,
                  beam_size: int,
+                 dtype: str,
                  bos_id: int,
                  eos_id: int,
                  context: Union[mx.Context, List[mx.Context]],
@@ -446,6 +448,7 @@ class BeamSearch(mx.gluon.Block):
                  sample: Optional[int] = None) -> None:
         super().__init__(prefix='beam_search_')
         self.beam_size = beam_size
+        self.dtype = dtype
         self.bos_id = bos_id
         self.eos_id = eos_id
         self.output_vocab_size = output_vocab_size
@@ -461,6 +464,7 @@ class BeamSearch(mx.gluon.Block):
             self._scorer = scorer
             self._sort_norm_and_update_finished = SortNormalizeAndUpdateFinished(
                 prefix='sort_norm_and_update_finished_',
+                dtype=self.dtype,
                 pad_id=C.PAD_ID,
                 eos_id=eos_id,
                 scorer=scorer)
@@ -525,12 +529,12 @@ class BeamSearch(mx.gluon.Block):
 
         # locations of each batch item when first dimension is (batch * beam)
         batch_indices = mx.nd.arange(0, batch_size * self.beam_size, self.beam_size, dtype='int32', ctx=self.context)
-        first_step_mask = mx.nd.full((batch_size * self.beam_size, 1), val=np.inf, ctx=self.context, dtype='float32')
+        first_step_mask = mx.nd.full((batch_size * self.beam_size, 1), val=np.inf, ctx=self.context, dtype=self.dtype)
         first_step_mask[batch_indices] = 1.0
         pad_dist = mx.nd.full((batch_size * self.beam_size, self.output_vocab_size - 1), val=np.inf,
-                              ctx=self.context, dtype='float32')
+                              ctx=self.context, dtype=self.dtype)
         eos_dist = mx.nd.full((batch_size * self.beam_size, self.output_vocab_size), val=np.inf,
-                              ctx=self.context, dtype='float32')
+                              ctx=self.context, dtype=self.dtype)
         eos_dist[:, C.EOS_ID] = 0
 
         # Best word and hypotheses indices across beam search steps from topk operation.
@@ -544,7 +548,7 @@ class BeamSearch(mx.gluon.Block):
         max_output_lengths = mx.nd.repeat(max_output_lengths, self.beam_size)
 
         # scores_accumulated: chosen smallest scores in scores (ascending).
-        scores_accumulated = mx.nd.zeros((batch_size * self.beam_size, 1), ctx=self.context, dtype='float32')
+        scores_accumulated = mx.nd.zeros((batch_size * self.beam_size, 1), ctx=self.context, dtype=self.dtype)
 
         # If using a top-k lexicon, select param rows for logit computation that correspond to the
         # target vocab for this sentence.
@@ -729,6 +733,7 @@ def get_beam_search(models: List[SockeyeModel],
     global_avoid_trie = None if avoid_list is None else constrained.get_avoid_trie(avoid_list, vocab_target)
     bs = BeamSearch(
         beam_size=beam_size,
+        dtype=models[0].dtype,
         bos_id=C.BOS_ID,
         eos_id=C.EOS_ID,
         context=context,
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -54,7 +54,7 @@ class _SingleModelInference(_Inference):
         self._skip_softmax = skip_softmax
         self._const_lr = constant_length_ratio
 
-    def state_structure(self):
+    def state_structure(self) -> str:
         return self._model.state_structure()
 
     def encode_and_initialize(self, inputs: mx.nd.NDArray, valid_length: Optional[mx.nd.NDArray] = None):
@@ -88,10 +88,11 @@ class _EnsembleInference(_Inference):
             raise ValueError()
         self._const_lr = constant_length_ratio
 
-    def state_structure(self):
+    def state_structure(self) -> str:
         structure = ''
         for model in self._models:
             structure.append(model.state_structure())
+        return structure
 
     def encode_and_initialize(self, inputs: mx.nd.NDArray, valid_length: Optional[mx.nd.NDArray] = None):
         model_states = []  # type: List[List[mx.nd.NDArray]]
@@ -396,15 +397,18 @@ class SampleK(mx.gluon.HybridBlock):
         return best_hyp_indices, best_word_indices, values
 
 
-def _repeat_states(states: List, beam_size, state_structure) -> List:
+def _repeat_states(states: List, beam_size: int, state_structure: str) -> List:
     repeated_states = []
-    for i in range(len(state_structure)):
-        if state_structure[i] == 's':
+    for i, state_format in enumerate(state_structure):
+        if state_format == C.STEP_STATE or state_format == C.BIAS_STATE:
             # Steps and source_bias have batch dimension on axis 0
-            state = states[i].repeat(repeats=beam_size, axis=0)
-        else:
+            repeat_axis = 0
+        elif state_format == C.DECODER_STATE or state_format == C.ENCODER_STATE:
             # Decoder and encoder layer states have batch dimension on axis 1
-            state = states[i].repeat(repeats=beam_size, axis=1)
+            repeat_axis = 1
+        else:
+            raise ValueError("Provided state format %s not recognized." % state_format)
+        state = states[i].repeat(repeats=beam_size, axis=repeat_axis)
         repeated_states.append(state)
     return repeated_states
 
@@ -417,16 +421,18 @@ class SortStates(mx.gluon.HybridBlock):
 
     def hybrid_forward(self, F, best_hyp_indices, *states):
         sorted_states = []
-        for i in range(len(self.state_structure)):
-            if self.state_structure[i] == 's':
+        for i, state_format in enumerate(self.state_structure):
+            if state_format == C.STEP_STATE or state_format == C.BIAS_STATE:
                 # Steps and source_bias have batch dimension on axis 0
                 state = F.take(states[i], best_hyp_indices)
-            elif self.state_structure[i] == 'd':
+            elif state_format == C.DECODER_STATE:
                 # Decoder layer states have batch dimension on axis 1
                 state = F.take(states[i], best_hyp_indices, axis=1)
-            else:
+            elif state_format == C.ENCODER_STATE:
                 # No need for takes on encoder layer states
                 state = states[i]
+            else:
+                raise ValueError("Provided state format %s not recognized." % state_format)
             sorted_states.append(state)
         return sorted_states
 
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -54,6 +54,9 @@ class _SingleModelInference(_Inference):
         self._skip_softmax = skip_softmax
         self._const_lr = constant_length_ratio
 
+    def state_structure(self):
+        return self._model.state_structure()
+
     def encode_and_initialize(self, inputs: mx.nd.NDArray, valid_length: Optional[mx.nd.NDArray] = None):
         states, predicted_output_length = self._model.encode_and_initialize(inputs, valid_length, self._const_lr)
         predicted_output_length = predicted_output_length.expand_dims(axis=1)
@@ -84,6 +87,11 @@ class _EnsembleInference(_Inference):
         else:
             raise ValueError()
         self._const_lr = constant_length_ratio
+
+    def state_structure(self):
+        structure = ''
+        for model in self._models:
+            structure.append(model.state_structure())
 
     def encode_and_initialize(self, inputs: mx.nd.NDArray, valid_length: Optional[mx.nd.NDArray] = None):
         model_states = []  # type: List[List[mx.nd.NDArray]]
@@ -388,25 +396,32 @@ class SampleK(mx.gluon.HybridBlock):
         return best_hyp_indices, best_word_indices, values
 
 
-def _repeat_states(states: List, beam_size) -> List:
+def _repeat_states(states: List, beam_size, state_structure) -> List:
     repeated_states = []
-    for i in range(len(states)):
-        if i < 2:
+    for i in range(len(state_structure)):
+        if state_structure[i] == 's':
+            # Steps and source_bias have batch dimension on axis 0
             state = states[i].repeat(repeats=beam_size, axis=0)
         else:
+            # Decoder and encoder layer states have batch dimension on axis 1
             state = states[i].repeat(repeats=beam_size, axis=1)
         repeated_states.append(state)
     return repeated_states
 
 
 class SortStates(mx.gluon.HybridBlock):
+
+    def __init__ (self, state_structure, prefix):
+        mx.gluon.HybridBlock.__init__(self, prefix=prefix)
+        self.state_structure = state_structure
+
     def hybrid_forward(self, F, best_hyp_indices, *states):
         sorted_states = []
-        for i in range(len(states)):
-            if i < 2:
+        for i in range(len(self.state_structure)):
+            if self.state_structure[i] == 's':
                 # Steps and source_bias have batch dimension on axis 0
                 state = F.take(states[i], best_hyp_indices)
-            elif i >= ((len(states) - 2) // 2) + 2:
+            elif self.state_structure[i] == 'd':
                 # Decoder layer states have batch dimension on axis 1
                 state = F.take(states[i], best_hyp_indices, axis=1)
             else:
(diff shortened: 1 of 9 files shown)


sockeye/encoder.py
```diff
@@ -291,11 +291,13 @@ def hybrid_forward(self, F, data, valid_length):
 
         # (batch_size * heads, 1, seq_len)
         bias = F.expand_dims(self.valid_length_mask(data, valid_length), axis=1)
+        data = F.transpose(data, axes=(1, 0, 2))
```
fhieber: please add a comment with shapes

sockeye/encoder.py
```diff
@@ -291,11 +291,13 @@ def hybrid_forward(self, F, data, valid_length):
...
 
         for block in self.layers:
             data = block(data, bias)
 
         data = self.final_process(data, None)
+        data = F.transpose(data, axes=(1, 0, 2))
```
fhieber: please add a comment with shapes

sockeye/inference.py
```diff
@@ -893,9 +920,9 @@ def _get_inference_input(self,
         """
         batch_size = len(trans_inputs)
         lengths = [len(inp) for inp in trans_inputs]
-        source_length = mx.nd.array(lengths, ctx=self.context, dtype='float32')  # shape: (batch_size,)
+        source_length = mx.nd.array(lengths, ctx=self.context, dtype=self.dtype)  # shape: (batch_size,)
         max_length = max(len(inp) for inp in trans_inputs)
-        source = mx.nd.zeros((batch_size, max_length, self.num_source_factors), ctx=self.context, dtype='float32')
+        source_npy = np.zeros((batch_size, max_length, self.num_source_factors), dtype=np.float32)
```
fhieber: nice tweak, thanks!

sockeye/layers.py
```diff
@@ -428,8 +381,7 @@ def hybrid_forward(self, F,
                        inputs: mx.sym.Symbol,
                        input_lengths: Optional[mx.sym.Symbol] = None,
                        bias: Optional[mx.sym.Symbol] = None,
-                       previous_keys: Optional[mx.sym.Symbol] = None,
-                       previous_values: Optional[mx.sym.Symbol] = None):  # mypy: ignore
+                       previous_kv: Optional[mx.sym.Symbol] = None):  # mypy: ignore
```
fhieber: would prefer `previous_keys_values` here

fhieber: also requires update of docstring

sockeye/layers.py
```diff
@@ -444,33 +396,18 @@ def hybrid_forward(self, F,
...
-        # (batch, max_length, depth)
-        # pylint: disable=unbalanced-tuple-unpacking
-        queries, keys, values = F.split(combined, num_outputs=3, axis=2)
 
-        updated_keys = keys
-        if previous_keys is not None:
-            updated_keys = F.concat(previous_keys, keys, dim=1)
-            keys = _remove_first_step(F, updated_keys)
+        proj = self.ff_in(inputs)
+        queries, kv_1, kv_2 = F.split(proj, num_outputs=3, axis=2)
```
fhieber: this 3-way split, followed by concat of two elements seems wasteful. Can we solve this with `slice` somehow and would that be more efficient?

sockeye/layers.py
```diff
@@ -444,33 +396,18 @@ def hybrid_forward(self, F,
...
+        kv = F.concat(kv_1, kv_2, dim=2)
 
-        updated_values = values
-        if previous_values is not None:
-            updated_values = F.concat(previous_values, values, dim=1)
-            values = _remove_first_step(F, updated_values)
+        updated_kv = kv
+        if previous_kv is not None:
+            updated_kv = F.concat(previous_kv, kv, dim=0)
+            # TODO: remove slicing and initialize with empty array when available in 1.6
```
fhieber: thanks for adding this TODO! The dummy states are an annoyance

sockeye/transformer.py
```diff
@@ -174,27 +174,23 @@ def hybrid_forward(self, F,
...
+                       enc_att_kv: Optional[mx.sym.Symbol] = None) -> Tuple[mx.sym.Symbol,
                                                                            mx.sym.Symbol,
                                                                            mx.sym.Symbol]:
         # self-attention
-        target_self_att, keys, values = self.self_attention(self.pre_self_attention(target, None),
-                                                            None,
-                                                            target_bias,
-                                                            self_att_k,
-                                                            self_att_v)
+        target_self_att, kv = self.self_attention(self.pre_self_attention(target, None),
```
fhieber: would prefer `keys_values`

sockeye/inference.py
```diff
@@ -706,6 +707,32 @@ def __init__(self,
             self.strip_ids.add(self.unk_id)
         self.models = models
 
+        # interleave qkv with heads for fast attention inference
+        for model in self.models:
+            for layer in [l for l in model.encoder.layers] + [l for l in model.decoder.layers]:
```
fhieber: could you add some comments adding shape information for easier understanding?
I wonder, if this can be moved to the `SockeyeModel` class instead. It has an `inference_only` flag to control behavior for some speed optimizations. WOuld it be possible to trigger the interleaving below based on this flag, and move this code there? It could be a separate method similar to `disable_dropout`, see https://github.com/awslabs/sockeye/blob/sockeye_2/sockeye/model.py#L442

sockeye/beam_search.py
```diff
@@ -102,7 +104,6 @@ def decode_step(self,
         outputs, new_states = [], []
         for model, model_states in zip(self._models, states):
             logits, model_states, _ = model.decode_step(step_input, model_states, vocab_slice_ids)
-            logits = logits.astype('float32', copy=False)
```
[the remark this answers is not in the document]
blchu: Actually, looking at the TODO above, the cast is not something that's needed anymore since I've made changes to the rest of beam search so that the dtype matches the model's dtype. I'll remove the TODO regarding that.

sockeye/beam_search.py
```diff
@@ -396,28 +391,31 @@ def hybrid_forward(self, F, scores, target_dists, finished, best_hyp_indices):
...
-            state = mx.nd.take(state, best_hyp_indices)
-        else:
-            ValueError("state list can only be nested list or NDArrays")
-        sorted_states.append(state)
-    return sorted_states
+class SortStates(mx.gluon.HybridBlock):
+    def hybrid_forward(self, F, best_hyp_indices, *states):
+        sorted_states = []
+        for i in range(len(states)):
+            if i < 2:
```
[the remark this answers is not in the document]
blchu: I think splitting this into separate blocks based on shape would work as a solution, as well as turning the list into a dictionary of annotation -> state list mappings. Personally, between the two I would prefer using the dictionary, to keep the sorting step as one unified call. Alternatively, if we want to retain this as a single HybridBlock, we could use the parallel list of annotations as part of initialization to indicate the order of states passed into its call, which would require that the structure of the states passed in doesn't change during the entire beam search processes (which is required anyway for any HybridBlock).

sockeye/layers.py
blchu: Right, using `slice` didn't cross my mind for some reason when I was implementing this part. I'll try that and see if it provides benefit over what I'm doing now, which I believe it should.

sockeye/inference.py
blchu: This part of the code is intended to be only temporary until I have training working with these optimizations. The reason is that the `interleaved_matmul_encdec` ops for the faster multi-head attention assume a different structure to the key_value projections, which I address by modifying the projection weights after initializing the model. This shouldn't be needed anymore once models can be trained with these ops.

blchu: Actually, one reason to retain this code is for compatibility with older models, which would then also require a model version check. If that's the case, this can be moved to a more appropriate location.

sockeye/beam_search.py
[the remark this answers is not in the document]
fhieber: +1 on the model/decoder class having a method to announce its state organization which can be used to initialie the sorting block at inference time.

sockeye/inference.py
fhieber: If we want to merge this PR rather soon, we should maintain backwards compatibility and in that case a more appropriate location would be preferred by me. 
Once this optimization is available for training as well, we can do a major version bump (2.1).

fhieber:
Hi @blchu it seems like MXNet 1.6 will actually have these ops available or both GPU and CPU. Is this correct?

blchu:
@fhieber I've been informed that it won't, however I can test it out to see if that is the case.

fhieber:
@blchu According to the latest release vote (https://lists.apache.org/thread.html/87ae98b5449459f40b536af12bfe4b701305a28071fe7bd2173465d8%40%3Cdev.mxnet.apache.org%3E) they added CPU implementations.
Do you want to update this PR then? Once 1.6 is released we can merge this I think.

blchu:
Oh, in which case that would be good. There shouldn't need to be much changes to account for that, just tests to make sure that it runs as intended for both.

sockeye/beam_search.py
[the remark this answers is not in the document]
blchu: @fhieber my latest commit included changes addressing what was discussed here, I used a string to inform the sorting/repeating functions how to treat each state, retrieved by a method call to the model. Let me know your thoughts on this.

sockeye/decoder.py
```diff
@@ -148,6 +152,22 @@ def __init__(self,
                                                                      prefix="final_process_",
                                                                      num_hidden=self.config.model_size)
 
+    def state_structure(self):
```
fhieber: Could you add a return type annotation?
```suggestion
    def state_structure(self) -> str:
```

sockeye/decoder.py
```diff
@@ -82,6 +82,10 @@ def get_decoder(cls, config: DecoderConfig, inference_only: bool, prefix: str) -
     def __init__(self):
         super().__init__()
 
+    @abstractmethod
+    def state_structure(self):
```
fhieber:
```suggestion
    def state_structure(self) -> str:
```

sockeye/beam_search.py
```diff
@@ -85,6 +88,11 @@ def __init__(self,
             raise ValueError()
         self._const_lr = constant_length_ratio
 
+    def state_structure(self):
+        structure = ''
+        for model in self._models:
+            structure.append(model.state_structure())
```
fhieber: I think this is missing a return statement.

sockeye/beam_search.py
```diff
@@ -388,25 +396,32 @@ def hybrid_forward(self, F, scores, target_dists, finished, best_hyp_indices):
         return best_hyp_indices, best_word_indices, values
 
 
-def _repeat_states(states: List, beam_size) -> List:
+def _repeat_states(states: List, beam_size, state_structure) -> List:
```
fhieber:
```suggestion
def _repeat_states(states: List, beam_size: int, state_structure: str) -> List:
```

sockeye/beam_search.py
```diff
@@ -85,6 +88,11 @@ def __init__(self,
             raise ValueError()
         self._const_lr = constant_length_ratio
 
+    def state_structure(self):
```
fhieber:
```suggestion
    def state_structure(self) -> str:
```

sockeye/beam_search.py
```diff
@@ -54,6 +54,9 @@ def __init__(self,
         self._skip_softmax = skip_softmax
         self._const_lr = constant_length_ratio
 
+    def state_structure(self):
```
fhieber:
```suggestion
    def state_structure(self) -> str:
```

sockeye/beam_search.py
```diff
@@ -388,25 +396,32 @@ def hybrid_forward(self, F, scores, target_dists, finished, best_hyp_indices):
         return best_hyp_indices, best_word_indices, values
 
 
-def _repeat_states(states: List, beam_size) -> List:
+def _repeat_states(states: List, beam_size, state_structure) -> List:
     repeated_states = []
-    for i in range(len(states)):
-        if i < 2:
+    for i in range(len(state_structure)):
```
fhieber: I'd prefer
```suggestion
    for i, format in enumerate(state_structure):
```

sockeye/beam_search.py
```diff
@@ -388,25 +396,32 @@ def hybrid_forward(self, F, scores, target_dists, finished, best_hyp_indices):
...
+        if state_structure[i] == 's':
+            # Steps and source_bias have batch dimension on axis 0
             state = states[i].repeat(repeats=beam_size, axis=0)
         else:
```
fhieber: better safe than sorry and check for `format == 'e' or format == 'd'`, and raise a ValueError otherwise.

I'd also suggest setting a variable `axis` based on the if statement and then have only a single 'repeat' line.

sockeye/decoder.py
```diff
@@ -148,6 +152,22 @@ def __init__(self,
                                                                      prefix="final_process_",
                                                                      num_hidden=self.config.model_size)
 
+    def state_structure(self):
+        """
+        Returns the structure of states used for manipulation of the states.
```
fhieber: I like this interface method.
Since we are relying on the string coding here, can we introduce 's', 'e', and 'd' as constants in `constants.py` and then use them here, e.g. `C.DECODER_STATE / C.ENCODER_STATE / C.INFERENCE_STATE` ?

sockeye/model.py
```diff
@@ -460,9 +463,10 @@ def load_model(model_folder: str,
         cast_dtype = True
         dtype_source = 'current'
 
+    # TODO: change allow_missing back to false when fast attention is implemented for training
```
fhieber: Is fast attention available for training in 1.6 now?

fhieber:
@blchu did you also experiment with using your(?) optimized self-attention ops for Sockeye (`interleaved_matmul_selfatt_qk` & `interleaved_matmul_selfatt_valatt`)?

gluon-nlp uses them for Bert and I am wondering if this is something we should do as well: https://github.com/dmlc/gluon-nlp/pull/1136

blchu:
@fhieber the self-attention ops require the queries to have the same sequence length as the keys and values, and so can't be used even for decoder self-attention during inference, which is why I'm using the encdec ops for both encoder-decoder attention and self-attention.

fhieber:
I see, thanks for clarifying! Since the use of the fused ops is now also enabled in training, have you observed differences in perplexity between this version and current `sockeye_2` branch? Is this be expected given that you change the structure of the model parameters, ie. fusing input projections in the attention layer (https://github.com/awslabs/sockeye/pull/752/files#diff-74c804465509cbab074609c50c13a8f7R434) ?

blchu:
The only differences I've observed are due to using more fp16 ops (and even then, it affects the score by <0.1%). Combining input projections only changes how the tensors are presented, and are mathematically equivalent, and thus don't affect the actual output/predictions at all.

fhieber:
Yes, I understand that for inference, changes are likely not noticable, what I am referring to is training: Since you fused the key/value parameter matrix, given the same random seed you'll get a slightly different parameter initialization. I just ran the first system test (`pytest test/system`) until the first checkpoint on `sockeye_2` and your branch (`sockeye_2_perf_optim`):

branch: `sockeye_2_perf_optim`
```
[INFO:sockeye.training] Training started.
[INFO:sockeye.training] Epoch[0] Batch [50]	Speed: 1038.94 samples/sec 10389.37 tokens/sec 64.93 updates/sec	perplexity=11.058745
[INFO:sockeye.training] Epoch[0] Batch [100]	Speed: 1026.58 samples/sec 10265.83 tokens/sec 64.16 updates/sec	perplexity=8.913289
[INFO:sockeye.training] Epoch[0] Batch [150]	Speed: 1017.80 samples/sec 10177.97 tokens/sec 63.61 updates/sec	perplexity=7.567821
[INFO:sockeye.training] Epoch[0] Batch [200]	Speed: 1015.62 samples/sec 10156.21 tokens/sec 63.48 updates/sec	perplexity=6.624727
[INFO:sockeye.training] Epoch[0] Batch [250]	Speed: 1023.01 samples/sec 10230.09 tokens/sec 63.94 updates/sec	perplexity=5.952862
[INFO:sockeye.training] Epoch[0] Batch [300]	Speed: 1026.37 samples/sec 10263.73 tokens/sec 64.15 updates/sec	perplexity=5.395916
[INFO:sockeye.training] Epoch[0] Batch [350]	Speed: 1012.51 samples/sec 10125.10 tokens/sec 63.28 updates/sec	perplexity=4.982689
[INFO:sockeye.training] Epoch[0] Batch [400]	Speed: 999.30 samples/sec 9992.96 tokens/sec 62.46 updates/sec	perplexity=4.624363
[INFO:sockeye.training] Epoch[0] Batch [450]	Speed: 980.17 samples/sec 9801.69 tokens/sec 61.26 updates/sec	perplexity=4.330886
[INFO:sockeye.training] Epoch[0] Batch [500]	Speed: 979.53 samples/sec 9795.30 tokens/sec 61.22 updates/sec	perplexity=4.074068
[INFO:sockeye.training] Epoch[0] Batch [550]	Speed: 826.07 samples/sec 8260.70 tokens/sec 51.63 updates/sec	perplexity=3.849256
[INFO:sockeye.training] Epoch[0] Batch [600]	Speed: 988.35 samples/sec 9883.46 tokens/sec 61.77 updates/sec	perplexity=3.651894
[INFO:sockeye.training] Epoch[1] Batch [650]	Speed: 972.31 samples/sec 9723.11 tokens/sec 60.77 updates/sec	perplexity=3.476266
[INFO:sockeye.training] Epoch[1] Batch [700]	Speed: 965.61 samples/sec 9656.07 tokens/sec 60.35 updates/sec	perplexity=3.315779
[INFO:sockeye.training] Epoch[1] Batch [750]	Speed: 968.91 samples/sec 9689.05 tokens/sec 60.56 updates/sec	perplexity=3.180055
[INFO:sockeye.training] Epoch[1] Batch [800]	Speed: 980.64 samples/sec 9806.38 tokens/sec 61.29 updates/sec	perplexity=3.055236
[INFO:sockeye.training] Epoch[1] Batch [850]	Speed: 977.26 samples/sec 9772.64 tokens/sec 61.08 updates/sec	perplexity=2.938140
[INFO:sockeye.training] Epoch[1] Batch [900]	Speed: 977.68 samples/sec 9776.82 tokens/sec 61.11 updates/sec	perplexity=2.829028
[INFO:sockeye.training] Epoch[1] Batch [950]	Speed: 978.95 samples/sec 9789.46 tokens/sec 61.18 updates/sec	perplexity=2.730155
[INFO:sockeye.training] Epoch[1] Batch [1000]	Speed: 978.57 samples/sec 9785.69 tokens/sec 61.16 updates/sec	perplexity=2.641466
[INFO:root] Saved params to "/tmp/test_seq_copyrfc8dciy/train_translate/model/params.00001"
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=20.179 Updates/sec=49.557
[INFO:sockeye.training] Checkpoint [1]	Train-perplexity=2.641466
[INFO:sockeye.training] Checkpoint [1]	Validation-perplexity=1.032465
```
branch: `sockeye_2`
```
[INFO:sockeye.training] Epoch[0] Batch [50]	Speed: 953.36 samples/sec 9533.61 tokens/sec 59.59 updates/sec	perplexity=13.272038
[INFO:sockeye.training] Epoch[0] Batch [100]	Speed: 929.06 samples/sec 9290.63 tokens/sec 58.07 updates/sec	perplexity=11.060930
[INFO:sockeye.training] Epoch[0] Batch [150]	Speed: 934.17 samples/sec 9341.75 tokens/sec 58.39 updates/sec	perplexity=9.474380
[INFO:sockeye.training] Epoch[0] Batch [200]	Speed: 913.22 samples/sec 9132.20 tokens/sec 57.08 updates/sec	perplexity=8.334352
[INFO:sockeye.training] Epoch[0] Batch [250]	Speed: 849.70 samples/sec 8497.00 tokens/sec 53.11 updates/sec	perplexity=7.469096
[INFO:sockeye.training] Epoch[0] Batch [300]	Speed: 943.43 samples/sec 9434.32 tokens/sec 58.96 updates/sec	perplexity=6.785138
[INFO:sockeye.training] Epoch[0] Batch [350]	Speed: 901.35 samples/sec 9013.48 tokens/sec 56.33 updates/sec	perplexity=6.238354
[INFO:sockeye.training] Epoch[0] Batch [400]	Speed: 932.25 samples/sec 9322.49 tokens/sec 58.27 updates/sec	perplexity=5.770828
[INFO:sockeye.training] Epoch[0] Batch [450]	Speed: 894.95 samples/sec 8949.50 tokens/sec 55.93 updates/sec	perplexity=5.386468
[INFO:sockeye.training] Epoch[0] Batch [500]	Speed: 935.37 samples/sec 9353.70 tokens/sec 58.46 updates/sec	perplexity=5.041607
[INFO:sockeye.training] Epoch[0] Batch [550]	Speed: 936.78 samples/sec 9367.81 tokens/sec 58.55 updates/sec	perplexity=4.750448
[INFO:sockeye.training] Epoch[0] Batch [600]	Speed: 938.63 samples/sec 9386.34 tokens/sec 58.66 updates/sec	perplexity=4.494768
[INFO:sockeye.training] Epoch[1] Batch [650]	Speed: 938.61 samples/sec 9386.11 tokens/sec 58.66 updates/sec	perplexity=4.266690
[INFO:sockeye.training] Epoch[1] Batch [700]	Speed: 939.29 samples/sec 9392.92 tokens/sec 58.71 updates/sec	perplexity=4.054684
[INFO:sockeye.training] Epoch[1] Batch [750]	Speed: 940.36 samples/sec 9403.62 tokens/sec 58.77 updates/sec	perplexity=3.874874
[INFO:sockeye.training] Epoch[1] Batch [800]	Speed: 941.35 samples/sec 9413.54 tokens/sec 58.83 updates/sec	perplexity=3.711587
[INFO:sockeye.training] Epoch[1] Batch [850]	Speed: 934.86 samples/sec 9348.56 tokens/sec 58.43 updates/sec	perplexity=3.562968
[INFO:sockeye.training] Epoch[1] Batch [900]	Speed: 942.74 samples/sec 9427.43 tokens/sec 58.92 updates/sec	perplexity=3.426815
[INFO:sockeye.training] Epoch[1] Batch [950]	Speed: 936.59 samples/sec 9365.94 tokens/sec 58.54 updates/sec	perplexity=3.306326
[INFO:sockeye.training] Epoch[1] Batch [1000]	Speed: 934.77 samples/sec 9347.73 tokens/sec 58.42 updates/sec	perplexity=3.192111
[INFO:root] Saved params to "/tmp/test_seq_copy26_cz08f/train_translate/model/params.00001"
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=21.206 Updates/sec=47.157
[INFO:sockeye.training] Checkpoint [1]	Train-perplexity=3.192111
[INFO:sockeye.training] Checkpoint [1]	Validation-perplexity=1.154931
```

I don't think this is an issue, but we need to make sure training converges similarly when training a proper model before merging this change.

fhieber:
Can you merge the updated `sockeye_2` branch also? We can then start running more tests on our side, thanks!

blchu:
I realized I misunderstood your earlier comment. The results that you showed look like they can probably be explained by different randomizations even from the same seed like you suggested, but I also did try training a full model on WMT afterwards and am noting that convergence is significantly slower. I'm investigating it, but I'll also merge in changes from the `sockeye_2` branch as I'm doing so.

blchu:
@fhieber As I'm debugging the current test errors, I've noticed that in the `test_beam_search` unit test an integer state gets passed into `_repeat_states` and `_sort_states`, which should cause a ValueError, but in the main sockeye_2 branch that doesn't produce errors in the test because the errors are actually not raised. Is this intentional?

fhieber:
@blchu you are right this test is still rather weak. There is a `TODO` to make it a more useful test. Feel free to adapt the test to your needs.

fhieber:
@blchu What is the status on this PR? Did you find the issue with the perplexity differences when using the fused attention ops? If not, would it make sense to split this PR into two, one for the fused ops, one for the other improvements (hybrid state sorting, copy input to GPU at once etc.)?

blchu:
I've managed to trace this to some code within the operator itself that results in running calculations at reduced precision. Currently I'm testing some possible solutions, but since this requires a change in MXNet itself, it won't be fixed until the next update (a 1.6.1 bugfix patch at the earliest). If this process takes too long, I may split the PR as you suggested.

fhieber:
Great to hear that you have root-caused the issue with this! I think it would be useful to merge the contributions that we can have right now, so if splitting the PR is not too much work, I‘d be happy to pull in the other optimizations.

blchu:
@fhieber just created a separate PR for the other optimizations, you can find it here https://github.com/awslabs/sockeye/pull/798

fhieber:
Hi @blchu, I saw that this change of yours got merged into MXNET: https://github.com/apache/incubator-mxnet/pull/17994

Would this unblock this PR?

blchu:
@fhieber yes, that PR fixes the issue that had resulted in the suboptimal training I observed, though we'll need to wait for it to be included in the 1.7 release

fhieber:
@blchu, should this be closed in favor of #814 ?

PR #763, opened 2019-12-17
Added more flexibility for source factors combination

davvil:
Added more flexibility for source fators combination

- The combination strategy, sum, concat or average (new) can be
  specified for each factor individually
- Source factors can share the same embedding matrix as the input

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/arguments.py
```diff
@@ -645,8 +672,14 @@ def add_model_parameters(params):
                                    '(validation) source factor files. Default: %(default)s.')
     model_params.add_argument('--source-factors-combine', '-sfc',
                               choices=C.SOURCE_FACTORS_COMBINE_CHOICES,
-                              default=C.SOURCE_FACTORS_COMBINE_CONCAT,
+                              default=[C.SOURCE_FACTORS_COMBINE_CONCAT],
+                              nargs='+',
                               help='How to combine source factors. Default: %(default)s.')
```
fhieber: could you add a sentence in help to indicate that this is a list?

sockeye/arguments.py
```diff
@@ -645,8 +672,14 @@ def add_model_parameters(params):
...
+    model_params.add_argument('--source-factors-share-embedding',
+                              type=bool_str(),
+                              nargs='+',
+                              default=[False],
+                              help='Share the embeddings with the source language. Default: do not share.')
```
fhieber: could you add a sentence in help to indicate that this is a list?

sockeye/encoder.py
```diff
@@ -93,10 +93,16 @@ def get_max_seq_len(self) -> Optional[int]:
 
 class FactorConfig(config.Config):
 
-    def __init__(self, vocab_size: int, num_embed: int) -> None:
+    def __init__(self,
+                 vocab_size: int,
+                 num_embed: int,
+                 combine: str, # From C.SOURCE_FACTORS_COMBINE_CHOICES
```
fhieber: any value in making this an Enum?

sockeye/encoder.py
```diff
@@ -143,20 +147,38 @@ def __init__(self,
...
-                    self.factor_embeds.add(mx.gluon.nn.Embedding(fc.vocab_size, fc.num_embed,
-                                                                 prefix="factor%d_" % i))
-
-    def hybrid_forward(self, F, data, valid_length, embed_weight):  # pylint: disable=arguments-differ
-        factor_embeds = []
+                for i, fc in enumerate(self.config.factor_configs):
+                    factor_weight_name = 'factor%d_weight' % i
+                    factor_weight = embed_weight if fc.share_source_embedding else \
+                        self.params.get('factor%d_weight' % i, shape=(fc.vocab_size, fc.num_embed))
+                    # We set the attribute of the class to trigger the hybrid_forward parameter creation "magic"
```
fhieber: What is the magic here?

sockeye/encoder.py
```diff
@@ -143,20 +147,38 @@ def __init__(self,
...
+                    setattr(self, factor_weight_name, factor_weight)
+
+    def hybrid_forward(self, F, data, valid_length, embed_weight, **kwargs):  # pylint: disable=arguments-differ
+        # We will catch the optional factor weights in kwargs
```
fhieber: not necessarily of type symbol, rather `Union[Symbol, ndarray]`

sockeye/arguments.py
davvil: Done, although the help produced by argparse already shows this.

sockeye/arguments.py
davvil: Done, although the help produced by argparse already shows this.

sockeye/encoder.py
davvil: It's a bit cleaner, although the whole parameter validation is already performed by argparse.

sockeye/encoder.py
davvil: mxnet defines the parameters of the block (and thus passed to hybrid_forward) as the members of the block which are of class Parameter. The goal is to use something like
```
self.my_param = self.params.get('blablabla', ...)
```
But in our case we create the parameters dynamically. Simply storing them in a list does not work (`self.factor_weights = [self.params.get(...) for ...]`), as `factor_weights` would be of type List. Therefore we create a number of class attributes of the appropriate class.

PR #773, opened 2020-01-17
Vocabulary selection w/ mxnet arrays

fhieber:
Requires MXNet 1.6.

Makes use of new set operations in MXNet 1.6.
Requires passing context to TopKLexicon class.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -554,15 +554,16 @@ class BeamSearch(mx.gluon.Block):
         vocab_slice_ids = None  # type: Optional[mx.nd.NDArray]
         if restrict_lexicon:
             source_words = utils.split(source, num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
-            vocab_slice_ids = restrict_lexicon.get_trg_ids(source_words.astype("int32").asnumpy())
+            vocab_slice_ids = restrict_lexicon.get_trg_ids(source_words)
             if any(raw_constraint_list):
                 # Add the constraint IDs to the list of permissibled IDs, and then project them into the reduced space
-                constraint_ids = np.array([word_id for sent in raw_constraint_list for phr in sent for word_id in phr])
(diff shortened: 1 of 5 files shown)


sockeye/beam_search.py
```diff
@@ -554,15 +554,16 @@ def forward(self,
...
         if restrict_lexicon:
             source_words = utils.split(source, num_outputs=self.num_source_factors, axis=2, squeeze_axis=True)[0]
-            vocab_slice_ids = restrict_lexicon.get_trg_ids(source_words.astype("int32").asnumpy())
+            vocab_slice_ids = restrict_lexicon.get_trg_ids(source_words)
             if any(raw_constraint_list):
                 # Add the constraint IDs to the list of permissibled IDs, and then project them into the reduced space
-                constraint_ids = np.array([word_id for sent in raw_constraint_list for phr in sent for word_id in phr])
-                vocab_slice_ids = np.lib.arraysetops.union1d(vocab_slice_ids, constraint_ids)
+                constraint_ids = [word_id for sent in raw_constraint_list for phr in sent for word_id in phr]
+                vocab_slice_ids = mx.nd.numpy.unique(
```
tdomhan: I wonder if it's possible to avoid the back-and-forth copy (we already had this before with asnumpy though of course). Do the mx.nd.numpy arrays live on the CPU or GPU?

We should benchmark this in comparison to using sparse ops for the union
```
set1 = mx.nd.sparse.row_sparse_array(([[1]], [27]), shape=(32, 1))
set2 = mx.nd.sparse.row_sparse_array(([[1], [1]], [29, 32]), shape=(32, 1))
union = (set1 + set2) > 0
```

fhieber: yes, good callout. I don't know how the conversion between npx and nd arrays is done in the backend and whether its actually copying the data. I agree that if it does there is probably no benefit in this. 

+1 on comparing with sparse op union.

PR #798, opened 2020-03-21
Sockeye 2 Inference Optimizations

blchu:
Made changes to Sockeye 2 to improve the performance of the Transformer model in machine translation for inference.

A list of changes detailed below:
1. Removed the take ops that were applied to the encoder states, as they do not change on different beams; this effectively cuts compute time for these takes in half
2. Gathered the input token ids into a numpy array on CPU before sending them all to the GPU at the beginning of beam search, rather than sending each batch element to the GPU one at a time
3. Set the data type of arrays during beam search computation to match the model's data type, rather than explicitly setting it to fp32

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -270,14 +270,6 @@ class CandidateScorer(mx.gluon.HybridBlock):
         return (scores + bp) * self._lp(lengths)
 
 
-class SortByIndex(mx.gluon.HybridBlock):
-    """
-    A HybridBlock that sorts args by the given indices.
-    """
-    def hybrid_forward(self, F, indices, *args):
-        return [F.take(arg, indices) for arg in args]
-
-
 class SortNormalizeAndUpdateFinished(mx.gluon.HybridBlock):
     """
     A HybridBlock for normalizing newly finished hypotheses scores with LengthPenalty.
@@ -409,17 +401,21 @@ def _repeat_states(states: List, beam_size) -> List:
     return repeated_states
 
 
-def _sort_states(states: List, best_hyp_indices: mx.nd.NDArray) -> List:
-    sorted_states = []
-    for state in states:
-        if isinstance(state, List):
-            state = _sort_states(state, best_hyp_indices)
-        elif isinstance(state, mx.nd.NDArray):
-            state = mx.nd.take(state, best_hyp_indices)
-        else:
-            ValueError("state list can only be nested list or NDArrays")
-        sorted_states.append(state)
-    return sorted_states
+class SortStates(mx.gluon.HybridBlock):
+    def hybrid_forward(self, F, best_hyp_indices, *states):
+        sorted_states = []
+        for i in range(len(states)):
+            if i < 2:
+                # Steps and source_bias have batch dimension on axis 0
+                state = F.take(states[i], best_hyp_indices)
+            elif i >= ((len(states) - 2) // 2) + 2:
+                # Decoder layer states have batch dimension on axis 1
+                state = F.take(states[i], best_hyp_indices, axis=1)
+            else:
+                # No need for takes on encoder layer states
+                state = states[i]
+            sorted_states.append(state)
+        return sorted_states
 
 
 # TODO (fhieber): add full fp16 decoding with mxnet > 1.5
@@ -461,7 +457,7 @@ class BeamSearch(mx.gluon.Block):
         self.global_avoid_trie = global_avoid_trie
 
         with self.name_scope():
-            self._sort_by_index = SortByIndex(prefix='sort_by_index_')
+            self._sort_states = SortStates(prefix='sort_states_')
             self._update_scores = UpdateScores(prefix='update_scores_')
             self._scorer = scorer
             self._sort_norm_and_update_finished = SortNormalizeAndUpdateFinished(
@@ -675,7 +671,7 @@ class BeamSearch(mx.gluon.Block):
                 break
 
             # (5) update models' state with winning hypotheses (ascending)
-            model_states = _sort_states(model_states, best_hyp_indices)
+            model_states = self._sort_states(best_hyp_indices, *model_states)
 
         logger.debug("Finished after %d out of %d steps.", t, max_iterations)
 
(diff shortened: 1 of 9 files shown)


sockeye/beam_search.py
```diff
@@ -84,6 +92,12 @@ def __init__(self,
             raise ValueError()
         self._const_lr = constant_length_ratio
 
+    def state_structure(self) -> str:
+        structure = ''
+        for model in self._models:
+            structure.append(model.state_structure())
```
fhieber: `structure` is a string, not a list.

sockeye/beam_search.py
```diff
@@ -30,6 +30,10 @@
 
 class _Inference(ABC):
 
+    @abstractmethod
+    def state_structure(self):
```
fhieber: It seems that we now got rid of the nested state lists in SortStates. This also means that we need to return a flat list of states in the case of ensembles. How about having `state_structure` always return a string, and `encode_and_initialize` and `decode_step` always a flat list of states? This would require changes in lines 93 and 108 (`+=` instead of `append`).

blchu: Actually, now that I'm considering it more, it may be better to revert back to using treating nested state lists in SortStates. I have a quick question about that though, in that case, are the model states guaranteed to be either a list of arrays (single model), or a list of lists of arrays (ensemble)? Are there any cases of lists of lists of lists of arrays, etc?

fhieber: Yes, there is no arbitrary depth of nesting, just 2 levels in case of ensembling.

blchu:
@fhieber I eventually arrived at a solution where the states are still a flat list (so that SortStates can still be hybridized), but the nesting structure is retained in the state structure, which is now a list of strings. That way, for ensemble inference, the state structure can be used to determine which states are associated with each model for decoding.

sockeye/layers.py
```diff
@@ -431,7 +432,7 @@ def hybrid_forward(self, F,
         :param input_lengths: Optional lengths of inputs to mask attention scores. Shape: (batch, 1).
         :param bias: Optional 3d bias tensor to mask attention scores.
         :param previous_keys: Optional previous input projections of keys. Shape: (batch, max_length+1, depth_att).
-        :param previous_values: Optional previous input projections of values. Shape: (batch, max_length+1, depth_att).
+        :param previous_keys: Optional previous input projections of values. Shape: (batch, max_length+1, depth_att).
```
fhieber:
```suggestion
        :param previous_values: Optional previous input projections of values. Shape: (batch, max_length+1, depth_att).
```

mjdenkowski:
Thanks for working on this, Brenton!

When running a benchmark, we discovered an issue with FP16 inference.  Using the current version from this PR, we hit an exception due to operations not supporting FP16: 
```
Traceback (most recent call last):
  File "/home/ec2-user/miniconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/ec2-user/miniconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/ec2-user/sockeye/sockeye/translate.py", line 268, in <module>
    main()
  File "/home/ec2-user/sockeye/sockeye/translate.py", line 46, in main
    run_translate(args)
  File "/home/ec2-user/sockeye/sockeye/translate.py", line 154, in run_translate
    input_is_json=args.json_input)
  File "/home/ec2-user/sockeye/sockeye/translate.py", line 236, in read_and_translate
    chunk_time = translate(output_handler, chunk, translator)
  File "/home/ec2-user/sockeye/sockeye/translate.py", line 259, in translate
    trans_outputs = translator.translate(trans_inputs)
  File "/home/ec2-user/sockeye/sockeye/inference.py", line 836, in translate
    batch_translations = self._translate_nd(*self._get_inference_input(translator_inputs))
  File "/home/ec2-user/sockeye/sockeye/inference.py", line 1015, in _translate_nd
    max_output_lengths))
  File "/home/ec2-user/miniconda3/lib/python3.7/site-packages/mxnet/gluon/block.py", line 693, in __call__
    out = self.forward(*args)
  File "/home/ec2-user/sockeye/sockeye/beam_search.py", line 721, in forward
    constraints = [constraints[x] for x in best_hyp_indices.asnumpy()]
  File "/home/ec2-user/miniconda3/lib/python3.7/site-packages/mxnet/ndarray/ndarray.py", line 2535, in asnumpy
    ctypes.c_size_t(data.size)))
  File "/home/ec2-user/miniconda3/lib/python3.7/site-packages/mxnet/base.py", line 255, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))
mxnet.base.MXNetError: [19:01:42] src/operator/tensor/./ordering_op-inl.h:613: This operation does not support float16
```

Felix suggested that this was due to changes such as not casting logits before the softmax and moving arrays used in beam search to FP16.  The issue should be reproducible by adding `--dtype float16` to the `sockeye.translate` call for any trained model.  Adding `--no-hybridization` can also help with debugging.

blchu:
looks like argsort doesn't like fp16, I'll have a fix up shortly

fhieber:
Thanks @blchu for the quick update! We should probably raise an issue with MXNet that `argsort` does not yet support fp16.
@mjdenkowski can you try your test once again? If all good, I think this is ready for merging.

mjdenkowski:
Thanks Brenton!  FP16 inference works now.

blchu:
are there any other potential issues that would keep these changes from being merged in?

sockeye/inference.py
```diff
@@ -888,9 +888,9 @@ def _get_inference_input(self,
         """
         batch_size = len(trans_inputs)
         lengths = [len(inp) for inp in trans_inputs]
-        source_length = mx.nd.array(lengths, ctx=self.context, dtype='float32')  # shape: (batch_size,)
+        source_length = mx.nd.array(lengths, ctx=self.context, dtype=self.dtype)  # shape: (batch_size,)
         max_length = max(len(inp) for inp in trans_inputs)
-        source = mx.nd.zeros((batch_size, max_length, self.num_source_factors), ctx=self.context, dtype='float32')
+        source_npy = np.zeros((batch_size, max_length, self.num_source_factors), dtype=np.float32)
```
tdomhan, line 893: What's the reason to switch to numpy here?

fhieber: "Gathered the input token ids into a numpy array on CPU before sending them all to the GPU at the beginning of beam search, rather than sending each batch element to the GPU one at a time"

Copying the input to the GPU at once is faster.

tdomhan: sure, but the same could have been achieved via `ctx=mx.cpu()`, no? Anyway, not a big deal. We can always change this to MXNet if we need this in the future.

PR #800, opened 2020-03-23
Option for setting parameters in model

annacurrey:
Added some additional options to models/model loading, including a function to set model parameters from a dictionary

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/model.py
+++ b/sockeye/model.py
@@ -15,7 +15,7 @@ import copy
 import time
 import logging
 import os
-from typing import cast, Optional, Tuple, Union, List
+from typing import cast, Dict, Optional, Tuple, Union, List
 
 import mxnet as mx
 from sockeye import __version__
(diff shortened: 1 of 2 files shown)


sockeye/model.py
```diff
@@ -317,6 +317,17 @@ def load_parameters(self,
                                 cast_dtype=cast_dtype, dtype_source=dtype_source)
         logger.info('Loaded params from "%s" to "%s"', filename, mx.cpu() if ctx is None else ctx)
 
+    def set_parameters(self, new_params: dict):
```
fhieber: Can you change the type to `Dict[str, mx.gluon.parameter.Parameter]` to make clear that this expects a mapping of Gluon parameters?

sockeye/model.py
```diff
@@ -317,6 +317,17 @@ def load_parameters(self,
...
+        """
+        Update model params with new values from a dictionary.
+
+        :param new_params: Dictionary containing the new parameters.
+        """
+        model_params = self.collect_params()
+        for k in new_params:
+            if k in model_params:
+                model_params[k].set_data(new_params[k].data())
```
fhieber: I assume this would fail if `model_params[k]` or `new_params[k]` are not initialized yet. Can you add a unit test for this? Would it make sense to fail here instead of relying on the implementation of `set_data`?

sockeye/model.py
fhieber: This method is very similar to `load_parameters()` from `mx.gluon.Block`, except that it does not load from disk.
Would it make sense to support some of its flags as well (i.e. `allow_missing`, `ignore_extra`)?
```python
def load_parameters(self, filename, ctx=None, allow_missing=False, ignore_extra=False,
    cast_dtype=False, dtype_source='current')```

File: sockeye/model.py [hunk 3]
```diff
@@ -317,6 +317,17 @@ def load_parameters(self,
                                 cast_dtype=cast_dtype, dtype_source=dtype_source)
         logger.info('Loaded params from "%s" to "%s"', filename, mx.cpu() if ctx is None else ctx)
 
+    def set_parameters(self, new_params: dict):
+        """
+        Update model params with new values from a dictionary.
```
fhieber: What is the expected behavior if the shape/context of the new parameter does not match the existing parameter? Can you document this and validate via some unit tests?

sockeye/model.py
annacurrey: Thanks for the suggestion! I added failures here and some unit tests to address this.

sockeye/model.py
annacurrey: Good idea -- this now has `allow_missing` and `ignore_extra` flags.

sockeye/model.py
annacurrey: The parameters are updated on all contexts of the model, regardless of the contexts of the new parameters. I've updated the docstring and added a relevant unit test.

If the shape doesn't match, this will fail, so I added a check for the shapes as well.

PR #814, opened 2020-05-22
Sockeye 2 Interleaved Multi-head Attention Operators

blchu:
Replaced batched dot product in multi-head attention with interleaved_matmul attention operators to improve performance. Also changes the batch-major data to time-major format while in the model to comply with the new operator requirements

until you can check this box.
- [ ] Updated CHANGELOG.md


fhieber:
Thanks! This should be mergable once MXNet 1.7 is out, right? Would you expect perplexity scores for training to be exactly the same with and without this change?

sockeye/encoder.py
```diff
@@ -330,11 +330,13 @@ def hybrid_forward(self, F, data, valid_length):
 
         # (batch_size * heads, 1, seq_len)
         bias = F.expand_dims(self.valid_length_mask(data, valid_length), axis=1)
+        data = F.transpose(data, axes=(1, 0, 2))
 
         for block in self.layers:
             data = block(data, bias)
 
         data = self.final_process(data, None)
+        data = F.transpose(data, axes=(1, 0, 2))
```
fhieber, line 339: I'd put the transpose before the final_process LayerNorm.

sockeye/decoder.py
```diff
@@ -288,39 +282,39 @@ def hybrid_forward(self, F, step_input, states):
...
-            new_self_att_kv += [new_self_att_k, new_self_att_v]
+        for layer, _self_att_kv, _enc_att_kv in zip(self.layers, self_att_kv, enc_att_kv):
+            target, _new_self_att_kv = layer(target,
+                                             mask,
+                                             source_encoded,
+                                             source_mask,
+                                             _self_att_kv,
+                                             _enc_att_kv)
+            new_self_att_kv.append(_new_self_att_kv)
         target = self.final_process(target, None)
```
fhieber, line 316: I'd put the transpose before the final_process LayerNorm.

sockeye/layers.py
```diff
@@ -466,40 +422,18 @@ def hybrid_forward(self, F,
...
-        queries = split_heads(F, queries, self.depth_per_head, self.heads)
-        keys = split_heads(F, keys, self.depth_per_head, self.heads)
-        values = split_heads(F, values, self.depth_per_head, self.heads)
 
-        updated_keys = keys
-        if previous_keys is not None:
-            updated_keys = F.concat(previous_keys, keys, dim=2)
-            keys = _remove_first_step(F, updated_keys)
+        proj = self.ff_in(inputs)
+        queries, kv_1, kv_2 = F.split(proj, num_outputs=3, axis=2)
```
fhieber, line 427: could this be done with two slice ops instead to avoid the concat? Would it be faster?

sockeye/layers.py
```diff
@@ -466,40 +422,18 @@ def hybrid_forward(self, F,
...
         """
-        # combined: (batch, max_length, depth * 3)
-        combined = self.ff_in(inputs)
-        # split into query, keys and values
-        # (batch, max_length, depth)
-        # pylint: disable=unbalanced-tuple-unpacking
-        queries, keys, values = F.split(combined, num_outputs=3, axis=2)
-
-        # scale by sqrt(depth_per_head)
-        queries = queries * (self.depth_per_head ** -0.5)
```
fhieber, line 477: I assume that scaling is done in the interleaved op?

sockeye/model.py
```diff
@@ -535,6 +535,7 @@ def load_model(model_folder: str,
         cast_dtype = True
         dtype_source = 'current'
 
+    # TODO: change allow_missing back to false when fast attention is implemented for training
```
fhieber, line 538: leftover comment?

blchu:
Hm, depending on what you're testing, my best guess is that the weights may be initialized slightly differently by the xavier initializer since the key and value weights are now a single array. Could you clarify what you mean by undoing the merged input projections (i.e. what changes did you make to do that)? In my own tests I didn't observe any more negative impact in training after my fix, but I can check it again to make sure.

sockeye/layers.py
blchu: yes

fhieber:
Here's a commit that still gives me different perplexity values when compared to sockeye_2 branch: https://github.com/awslabs/sockeye/commit/1676d6a5a79615d56387f4d2f206d9f6d6877d68
I might have missed something though.

blchu:
Looking at the changes in your commit, it looks like the discrepancy is likely still there because of the way that you've combined the weights together. The `interleaved_matmul_*` ops don't assume that the key and value projections are simply concatenated, but rather are striped together over the head dimension (hence the name "interleaved").

You can reference the code changes in sockeye/inference.py from 6b25ad550a07027c5531e359c35cbb3a0f680eef to see how the weights would need to be combined.

fhieber:
> Looking at the changes in your commit, it looks like the discrepancy is likely still there because of the way that you've combined the weights together. The `interleaved_matmul_*` ops don't assume that the key and value projections are simply concatenated, but rather are striped together over the head dimension (hence the name "interleaved").
> 
> You can reference the code changes in sockeye/inference.py from [6b25ad5](https://github.com/awslabs/sockeye/commit/6b25ad550a07027c5531e359c35cbb3a0f680eef) to see how the weights would need to be combined.

Right, thanks for the clarification. I will try interleaving them instead of concatenating.

blchu:
I tried testing the perplexity values myself, as it turns out since the model now transposes the batch and time dimensions, not only do the attention input arrays need to be interleaved together if the keys and values are initialized separately, but also that they need to be transposed for the dropout and layer norms to produce the same results. Accounting for all of this, I'm observing that the array values largely align with the main sockeye_2 branch.
There still is a small difference in perplexity values in the system test, and that's likely due to a ~1e-8 difference in outputs from the attention operator itself, since I believe its implementation uses a different GEMM than batch_dot does. I don't believe that this should result in any significant issue though.

I've pushed the changes I've made to run the tests onto a branch from my fork here https://github.com/blchu/sockeye/commits/sockeye_2_test_interleaved_perplex

mjdenkowski:
Now that we have a MXNet 1.7 branch, we can update/review this PR and merge it into that branch so it will be immediately usable for anyone who wants to manually install a MXNet pre-release.  When MXNet 1.7 wheels are available, `mx17` can merge into `master` and the improvements in this PR will be included automatically.

mjdenkowski:
It looks like there are some merge conflicts between this PR and the SSRU updates.  @barbaradarques and @blchu, as the experts on these parts of the code, can you help us find the best way to resolve the conflicts?

blchu:
Looking over the SSRU PR, there is significant overlap since both modify the decoder autoregressive layer. Most of the conflicts should be relatively straightforward to resolve, but one of my changes transposes the batch and time dimensions for the duration of the model to improve attention performance. @barbaradarques is that compatible with the SSRU layer, or can be made compatible? Otherwise I'll look for some workarounds.

tdomhan:
Actually I think there's some positive synergy between the changes as the SSRU code currently temporarily transposes to time major in order to use the foreach operator: https://github.com/awslabs/sockeye/blob/master/sockeye/layers.py#L893
When merging we'd just to make sure to remove that transpose in favor your transpose for the model.

barbaradarques:
Hi @blchu , as @tdomhan mentioned, the transposition shouldn't be a problem for SSRU layers. On top of removing the internal `transpose()` operations, you'd also need update [`get_state_shape()`](https://github.com/awslabs/sockeye/blob/master/sockeye/layers.py#L867), but that should be it.

mjdenkowski:
We've officially updated Sockeye to MXNet 1.7, so we can merge this into master as soon as the conflicts are resolved.

blchu:
I'm currently working on resolving the merge conflicts. Right now I have all tests passing; however, I'm seeing some strange performance and perplexity numbers in training, so I'm trying to get that resolved.

barbaradarques:
@blchu, feel free to commit your WIP, maybe we can help figuring out the inconsistencies.

blchu:
With further testing, I did notice that while perplexity was noticeable worse during training than what I had measured before, this is also the case with the mx17 branch without any of the changes from the PR, so it may be due to something entirely unrelated. The BLEU scores at each training checkpoint were no different.

There's still an issue with the inference during a training checkpoint using an SSRU decoder, though I think I've narrowed the cause to a different state shape than in pure inference ([batch, dim] vs the expected [time, batch, dim]).

fhieber:
thanks @blchu though I think something went wrong with the merge/conflict resolution (the PR shows 49 changed files).

blchu:
Ah hm, that might have been because the way I did the merge and the fact that I was originally building off of the sockeye_2 branch. I think the files changed is using a diff off of that branch, and not master perhaps? I'm not entirely sure how to fix that.

barbaradarques:
Hi @blchu , since a lot has changed since the original fork, it might help if you create a new branch from master and copy+paste on top of it the changes of those 6 files you actually changed.

blchu:
I've just created a new PR by cherry-picking my commits to a new branch off of master, since I can't modify the source branch of this PR. The new PR can be seen here https://github.com/awslabs/sockeye/pull/884.

PR #827, opened 2020-06-08
Distributed training update

mjdenkowski:
This PR adds and updates training settings for multi-process and mixed precision training.

Models are backward compatible and training settings are compatible with minor changes (`--no-bucket-scaling` option replaced with `--bucket-scaling` option, `--round-batch-sizes-to-multiple-of` option renamed to `--batch-sentences-multiple-of`).

Changes:

- A new `max-word` batch type sizes batches based on max words (versus average) and includes padding tokens in the count (versus non-padding only).  This sizing method allows more direct control over GPU memory usage.  Existing `word` batching remains the default and its unit tests are unchanged.  New unit tests cover `max-word` batching.

- The argument `--round-batch-sizes-to-multiple-of` is renamed `--batch-sentences-multiple-of` for clarity, with a default of 8.  Sizes are rounded up/down with `word` batching (nearest multiple) and always rounded down with `max-word` batching.

- Default bucketing settings are changed to width 8, max sequence length 95 (96 including BOS/EOS tokens), and no bucket scaling.  These are the recommended settings when using AMP and are observed not to degrade quality when not using AMP.

- Argument `--no-bucket-scaling` replaced with `--bucket-scaling`, which is False by default.

until you can check this box.
- [x] Updated CHANGELOG.md


tdomhan:
What's the reason for per device batch sizes? This seems less intuitive from a user perspective, doesn't it?

fhieber:
Given that we change user-facing behavior here, should we bump the major version?

tdomhan:
This would invalidate all previously trained models by the user, so I think we should avoid if at all possible. How about: Create a new argument `--device-batch-size` and fail with a message to change to the new argument when users use `--batch-size`? (similar for the other renamings/changes)

This way the change is not hidden to the user and we don't need to bump. What do you think?

mjdenkowski:
The latest commits rework this PR to be backward compatible and provide the most effective settings for different training scenarios:

- We keep the default `word` batching and rename the new version `max-word`.  Benchmarks show that `word` batching can yield slightly higher BLEU scores (0.1-0.2) for small batch training while `max-word` batching can yield slightly higher scores (again 0.1-0.2) for large batch training.

- We keep the current behavior where `batch_size` is per-process.

This PR is now marked as complete.

tdomhan:
Thanks Michael. Have you consider my suggestion to make it noticeable to the user that the behavior of the argument(s) changed?

mjdenkowski:
It's a good point that we can make life easier for our users with specific notifications when arguments are removed or renamed.  The latest commit adds such notifications for `--no-bucket-scaling` and `--round-batch-sizes-to-multiple-of`.

mjdenkowski:
The latest commit updates the sequence sorting tests to use `word` and `max-word` batching instead of `sentence` with values chosen to produce similar effective batch sizes.  System test results are confirmed stable.

PR #832, opened 2020-06-22
Quantization: use tied weight parameter

kpuatamazon:
MXNet 1.6 introduced optional deduplication when files are saved.  This effectively relies on the same weight objects being shared when a model is loaded, but previously quantization broke that.

This breaks existing models quantized to int8 format but is consistent with its own quantize.py.  There aren't many of those models out there though.

until you can check this box.
- [x] Unit tests pass (`pytest`): the same 20 related to chrf fail before/after.
- [x] Passed code style checking (`./style-check.sh`): I get the same 8 errors in 5 files.
- [ ] Updated CHANGELOG.md


kpuatamazon:
Regarding OS X 3.7, I've no idea how this change (which only changes a branch not called in the test) would reduce BLEU on copying for float32 from 0.98 to 0.9731009815384613.

PR #835, opened 2020-06-23
Cross-entropy loss without SoftmaxOutput

fhieber:
Use cross entropy loss implementation that avoids SoftmaxOutput by default.

Added tests to verify equivalence to SoftmaxOutput as much as possible (for the case without label smoothing). For label-smoothed cross-entropy, there actually seems to be an issue with SoftmaxOutputs gradients: I was not able to reproduce them with an independent implementation.

Speed-wise, the SoftmaxOutput-less cross-entropy loss is about 5-10% slower at equal memory cost.

However, given the issues with SoftmaxOutput smoothed gradients, I am inclined to change the default to a verifiable correct implementation.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.1.10]
+
+### Changed
+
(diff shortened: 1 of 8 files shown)


fhieber:
> Looks good!
> 
> Is the 5-10% slowdown for computing loss or for overall training throughput?

I had another look at the training logs. The slowdown is actually only 3-4% as measured in overall training throughput (time cost for until next checkpoint, or updates/sec)

PR #842, opened 2020-07-07
Add Inference Dropout feature

Deseaus:
Adding an option, `--mc-dropout`, which enables dropout during inference. This leads to non-deterministic translations that can be used to quantify model certainty on its predictions (Monte Carlo Dropout). This feature enables users to implement the multi-hypothesis evaluation techniques described here: https://www.aclweb.org/anthology/2020.acl-main.113.pdf

To enable the feature we have to turn on training mode on gluon, as otherwise dropout layers are disabled. The context manager can be put in various places, I just chose one that seemed convenient but happy to find it a better home :) Will extend the tests once the implementation is settled. 

Also, the feature will enable dropout using the parameters the model with trained with, enabling customisation of this via code seems overkill. For now, users can modify the config sockeye files if they want to use different dropout setting than those used in training. 

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber:
Thanks for the contribution!
When `mx-dropout` is enabled you disable the `inference_only` flag, presumably you want to avoid disabling the removal of dropout layers in `load_model()`. But it also has an effect on decoder state caching. Is this intended?

Have you also checked whether the `train_mode` causes allocation of gradient arrays in the inference model? I would assume no, because we set `grad_req=null` in `load_model` before, but it might be wort checking to avoid memory consumption.

sockeye/inference.py
```diff
@@ -833,7 +836,14 @@ def translate(self, trans_inputs: List[TranslatorInput], fill_up_batches: bool =
                 batch = batch + [batch[0]] * rest
 
             translator_inputs = [indexed_translator_input.translator_input for indexed_translator_input in batch]
-            batch_translations = self._translate_nd(*self._get_inference_input(translator_inputs))
+
+            if self.mc_dropout:
```
fhieber, line 840: not a big fan of having this in the `Translator` class. Is there a way to keep this local to the model class somehow?

Deseaus: I found a way to move it there, although split into two functions.

Deseaus:
I managed to separate the two parameters and keep `inference_only` set to `True`. Regarding the gradients, the NDArrays have the attribute, but they're empty. I imagine MXNet wouldn't fill them until `.backward()` is called. Either way I wouldn't worry too much about memory consumption, since this is purely a feature for testing and evaluating a model, not for real inference.

sockeye/inference.py
fhieber: Looks much better, thanks! Could you add a logging message in `load_model()` when `mc_dropout` is enabled? This would help debugging if users accidentally enable it and wonder about nondeterministic outputs :)

test/unit/test_inference.py
```diff
@@ -50,6 +50,7 @@ def mock_translator(batch_size: int = 1,
                                                   beam_search_stop=None,
                                                   nbest_size=None,
                                                   models=None,
+                                                  mc_dropout=None,
```
fhieber, line 53: This is no longer needed, no?

Deseaus: Fixed!

Deseaus:
Are you ok proceeding without tests? I tired modifying the seq copy integration test and adding one that translates *n* times with MC Dropout and checks that the output is different, but it doesn't work there (maybe the model is too small?). With a larger model on the CLI it works fine.

fhieber:
> Are you ok proceeding without tests? I tired modifying the seq copy integration test and adding one that translates _n_ times with MC Dropout and checks that the output is different, but it doesn't work there (maybe the model is too small?). With a larger model on the CLI it works fine.

It'd be great to have a test. But maybe it would be sufficient to check in a test that `load_model` with enabled mc_dropout creates a `train` mode and that dropout is not removed.

PR #843, opened 2020-07-08
Diverse Beam Search 

saeub:
Diverse Beam Search with Hamming diversity as described by [Vijayakumar et al. (2016)](https://arxiv.org/abs/1610.02424). `TopK` is applied to each one of the `num_diverse_groups` beam groups separately. When a word is selected in one group, that word's score is penalized in all subsequent groups. Regular beam search is still the default (with `--num-diverse-groups 1`) and should still have the same result.

Note: I didn't quite understand why `batch_size` and `beam_size` are combined into a single dimension, but I thought it was easiest to reshape them for treating each diverse group separately. I'm new to MXNet, so I might be missing something.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -366,6 +366,84 @@ class TopK(mx.gluon.HybridBlock):
         return F.reshape(F.cast(indices, 'int32'), shape=(-1,)), F.reshape(values, shape=(-1, 1))
 
 
+class DiverseTopK(mx.gluon.Block):
+    """
+    Batch-wise topk operation for diverse beam search with Hamming diversity.
+    Applies TopK to each diverse group separately, adding penalty for words already selected in previous groups.
+    """
+
+    def __init__(self,
+                 num_groups: int,
+                 group_size: int,
+                 penalty_strength: float,
+                 context: Union[mx.Context, List[mx.Context]],
+                 dtype: str,
+                 **kwargs) -> None:
+        """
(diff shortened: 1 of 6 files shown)


sockeye/beam_search.py
```diff
@@ -670,7 +678,45 @@ def forward(self,
                 if t == 1:
                     scores *= first_step_mask
 
-                best_hyp_indices, best_word_indices, scores_accumulated = self._top(scores, offset)
+                vocab_size = scores.shape[1]
+
+                # Groupwise shape: (batch_size, num_diverse_groups, diverse_group_size[, vocab_size])
```
fhieber, line 683: Instead of having this entire block inside the beam search class, would it be possible to move it into a separate Block (I assume not HybridBlock as you are using numpy-style indexing) that replaces `TopK`? The reason is that the Python for loop and the advanced indexing operations are likely slow.
Separating this code into its own block would be a bit cleaner I think.

fhieber: Maybe something like `GroupedTopK` or `DiverseTopK`

saeub: Ok, thanks! Moved it to a separate `Block`.

I'm having a hard time figuring out how to avoid the imperative `for` loop and indexing, but if you have a hint or suggestion, I'll be happy to have a stab at it.

fhieber:
closing this due to inactivity and incompatibility with the main branch.

PR #845, opened 2020-07-10
Updated output options for `sockeye.rerank`

mjdenkowski:
This PR adds 2 simple output options for `sockeye.rerank`:

- `--output`: Specify output file (`sys.stdout` by default)

- `--output-reference-instead-of-blank`: When outputting only the best hypothesis (`--output-best`) and the best hypothesis is blank, output the reference line instead of a blank line and log a warning message.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/rerank.py
+++ b/sockeye/rerank.py
@@ -19,6 +19,7 @@ import argparse
 from functools import partial
 import json
 import logging
+import sys
 from typing import Any, Dict, List
 
 import numpy as np
@@ -26,6 +27,7 @@ import sacrebleu
 
 from . import arguments
 from . import constants as C
+from .data_io import smart_open
 from . import log
 from . import utils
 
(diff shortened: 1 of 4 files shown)


sockeye/rerank.py
```diff
@@ -90,6 +92,7 @@ def rerank(args: argparse.Namespace):
     :param args: Namespace object holding CLI arguments.
     """
     reranker = Reranker(args.metric, args.return_score)
+    output_stream = sys.stdout if args.output is None else smart_open(args.output, mode='w')
```
fhieber, line 95: Can we move the open into a with statement or use a context manager to ensure we close the stream?

mjdenkowski: Good catch.  For this PR, I added a low-tech solution of just closing the stream when it's not `sys.stdout`.

I like the idea of a context manager for "file or stdout", probably as part of the utils module with proper unit tests.  Right now low-tech is less code overall, but if we find ourselves using this pattern again, we can add the utility function.

PR #847, opened 2020-07-24
Add Support for Target Factors

fhieber:
Adds support for target-side factors throughout the framework. These are learned in a multi-task setting, i.e. each factor has its own cross-entropy loss. Only the primary target factor (i.e. the surface words that existed before) participates in beam search during inference. Other, secondary factors are greedily decoded, i.e. the argmax prediction is carried along during beam search as metadata, and fed as factored input into the next decoder step. We also do not store or use the 'score' of secondary target factor outputs, i.e. the cost/score of a translation is the same as before this change.

Previously, the codebase contained many patterns of `(source_{paths,vocabs}, target_{path,vocab})`, i.e. there were multiple source-side resources allowed, but only a single target resource. With this change, this is now made parallel (i.e. many areas in the code now use `target_{paths,vocabs}` to reflect the support of target-side factors.

This change is backwards incompatible, therefore bumping major version.

This also changes the naming of parameters slightly. What was before (source) 'factor0' is now 'factor1' to reflect the convention that the first factor (0th, as 0-indexed) is the primary source/target factor, i.e. the surface form.

Losses for secondary target factors do not use label smoothing.

Secondary target factors are shifted by 1 to the right (inserting another BOS symbol for the target sequence of secondary target factors).

A detailed list of changes (mostly by file):

#### tutorials/seqcopy/genseqcopy.py
 - the small script now creates source and target factor data for the toy problem of digit copying. source factor: low/high, target factor: odd/even

#### arguments.py
 - CLI arguments for training/preparing data are augmented to contain target factor arguments.

#### beam_search.py
 - the `_Inference` class handles the greedy decoding of target factors if available, both for single model and ensemble decoding
 - the top-1 target factors for every beam search iteration are stored in a single tensor that is sorted according to the TopK output in `SortNormalizedAndUpdateFinished`. The result is then concatenated to the `best_word_indices` tensor, which serves as the input to the model's decoder step in the next iteration
 - constrained decoding and vocabulary selection only operate on the primary target factor (i.e. the surface words). These features are not affected by this change.

#### constants.py
 - introducing short metric names for more brevity in logging (when there are multiple target factors). 'perplexity' -> 'ppl'
 - bump prepared data version to 4

#### data_io.py
 - Generalize all target-side data logic to lists of target-side data (in line with source factors)

#### encoder.py
 - The `Embedding` class now works the same for both source and target embeddings. Optional factor configs passed to the class lead to the creation of a joint embedding of primary and secondary factors, controlled by the SOURCE_FACTOR_COMBINE setting (summing, averaging, concatenating).

#### inference.py
 - Update most classes to support factored output.
 - `TranslatorOutput` has two optional fields: `other_factors` and `other_factor_tokens` (other_factors can be dumped to json)
 - `Translator` changed constructor signature (target_vocab -> target_vocabs). It also has a new property `num_target_factors`.
 - nbest translations currently do not contain factor information. This is left for future work if need arises.

#### loss.py
 - Loss classes now have a field `metric_prefix` to name their corresponding metrics. This leads to clearer logging output during training with multiple source factors: `ppl` (for the primary target output), `f1-ppl` (for the second target factor), etc.
 - Added support for short metric names (perplexity -> ppl, accuracy -> acc)
 - For training with target factor, there is a separate output layer & loss instance for each factor

#### model.py
 - `SockeyeModel` class has 3 new properties: `num_target_factors`, `target_factor_configs`, and `factor_output_layers`
 - `load_model(s)` now returns a list of target vocabs.

#### output_handler.py
 - Added a new output handler (`translation_with_factors`) that adds the predicted target factors to each target words, separated by `|`.

#### test_utils.py
 - Added utility function to create odd/even factors for system/integration tests.

until you can check this box.
- [x] Updated majorversion in `sockeye/__init__.py`. Major version bump if this is a backwards incompatible change.
- [x] Updated CHANGELOG.md


mjdenkowski:
It looks like models with target factors are running end-to-end.  Thanks for working on this!

The code appears to be mechanically correct in terms of feeding target factor data through the model similar to source factor data.  I also see that you added target factors to some of the sequence copy tests.  Have you checked the results to see if models can correctly predict odd/even?  It looks like this could be automated with the new `translation_with_factors` output handler if it isn’t already.

fhieber:
Adding an explicit check for target factors in the system tests is a good point, I will add that.

But yes, it correctly predicts odd/even.

sockeye/beam_search.py
```diff
@@ -72,11 +72,19 @@ def decode_step(self,
...
                     states: List,
                     vocab_slice_ids: Optional[mx.nd.NDArray] = None):
-        logits, states, _ = self._model.decode_step(step_input, states, vocab_slice_ids)
+        logits, states, target_factor_outputs = self._model.decode_step(step_input, states, vocab_slice_ids)
         if not self._skip_softmax:
             logits = logits.log_softmax(axis=-1)
         scores = -logits
-        return scores, states
+
+        target_factors = None  # type: Optional[mx.nd.NDArray]
```
tdomhan, line 80: not super important, but given that this is identical to the ensemble case would it make sense to combine both into a single method (say in _Inference?)

sockeye/beam_search.py
```diff
@@ -305,7 +326,8 @@ def __init__(self,
         self._scorer = scorer
 
     def hybrid_forward(self, F, best_hyp_indices, best_word_indices,
-                       finished, scores_accumulated, lengths, reference_lengths):
+                       finished, scores_accumulated, lengths, reference_lengths,
+                       secondary_factors=None):
```
tdomhan, line 330: nit: I guess the intention for calling this `secondary_factors` is that the target words are technically the main factor, but given we don't call something primary factor in the arguemnts I'd find just `factors` clearer.

sockeye/data_io.py
```diff
@@ -989,9 +1000,9 @@ def get_scoring_data_iters(sources: List[str],
     are truncated.
 
     :param sources: Path to source training data (with optional factor data paths).
-    :param target: Path to target training data.
+    :param targets: Path to source training data (with optional factor data paths).
```
tdomhan, line 1003: Path to `target`

sockeye/model.py
```diff
@@ -111,18 +111,17 @@ def __init__(self,
...
             self.source_embed_weight, self.target_embed_weight, self.output_weight = self._get_embedding_weights()
 
             self.embedding_source = encoder.Embedding(config.config_embed_source,
-                                                      prefix=self.prefix,
-                                                      is_source=True,
+                                                      prefix=self.prefix + C.SOURCE_EMBEDDING_PREFIX,
                                                       embed_weight=self.source_embed_weight)
             self.embedding_target = encoder.Embedding(config.config_embed_target,
-                                                      prefix=self.prefix,
-                                                      is_source=False,
```
tdomhan, line 125: It's nice we can get rid of this. :)

sockeye/beam_search.py
fhieber: ok

sockeye/data_io.py
fhieber: done

sockeye/beam_search.py
fhieber: They are ALMOST the same. The ensemble case uses argmin and the interpolation method, whereas the single model case uses argmax. I'll leave it as is for now.

sockeye/model.py
```diff
@@ -237,42 +250,48 @@ def decode_step(self, step_input, states, vocab_slice_ids=None):
...
         decoder_out, new_states = self.decoder(target_embed, states)
 
         # step_output: (batch_size, target_vocab_size or vocab_slice_ids)
         step_output = self.output_layer(decoder_out, vocab_slice_ids)
 
-        return step_output, new_states, step_additional_outputs
+        # Target factor outputs are currently stored in additional outputs.
+        target_factor_outputs = []
+        # TODO: consider a dictionary mapping as return value
+        for i, factor_output_layer in enumerate(self.factor_output_layers, 1):
```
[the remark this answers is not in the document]
fhieber: removed

sockeye/inference.py
```diff
@@ -403,6 +403,8 @@ class TranslatorOutput:
     :param nbest_translations: List of nbest translations as strings.
     :param nbest_tokens: List of nbest translations as lists of tokens.
     :param nbest_scores: List of nbest scores, one for each nbest translation.
+    :param other_factors: List of secondary factor outputs.
```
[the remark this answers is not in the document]
fhieber: done

sockeye/inference.py
```diff
@@ -403,6 +403,8 @@ class TranslatorOutput:
...
+    :param other_factor_tokens: List of list of secondary factor tokens.
```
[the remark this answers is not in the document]
fhieber: thats a good idea.

sockeye/checkpoint_decoder.py
```diff
@@ -154,36 +157,54 @@ def decode_and_evaluate(self,
 
         # 1. Translate
         trans_wall_time = 0.0
-        translations = []
-        with data_io.smart_open(output_name, 'w') as output:
-            handler = sockeye.output_handler.StringOutputHandler(output)
+        translations = []  # type: List[List[str]]
+        with ExitStack() as exit_stack:
+            outputs = [exit_stack.enter_context(data_io.smart_open(output_name.format(factor=idx), 'w'))
+                       if output_name is not None else None for idx in range(self.model.num_target_factors)]
```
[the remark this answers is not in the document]
fhieber: fixed

sockeye/inference.py
```diff
@@ -610,7 +621,7 @@ def _concat_translations(translations: List[Translation],
         if idx == len(translations) - 1:
             target_ids.extend(translation.target_ids)
         else:
-            if translation.target_ids[-1] in stop_ids:
+            if translation.target_ids[-1][0] in stop_ids:
```
tdomhan, line 624: What about target factors, wouldn't we need to concatenate them as well?

sockeye/inference.py
```diff
@@ -962,36 +1005,42 @@ def _make_result(self,
...
         :return: TranslatorOutput.
         """
-        target_ids = translation.target_ids
-        target_tokens = [self.vocab_target_inv[target_id] for target_id in target_ids]
-        target_string = C.TOKEN_SEPARATOR.join(data_io.ids2tokens(target_ids, self.vocab_target_inv, self.strip_ids))
+        primary_target_tokens, primary_translation, other_target_tokens, other_target_strings = \
+            self._get_translation_tokens_and_factors(translation.target_ids)
 
         if translation.nbest_translations is None:
+            primary_target_tokens, primary_translation, other_target_tokens, other_target_strings = \
```
tdomhan, line 1012: Isn't this a duplicate to the line before the `if` statement?

sockeye/train.py
```diff
@@ -282,37 +300,48 @@ def create_data_iters_and_vocabs(args: argparse.Namespace,
...
-        check_condition(all([combine in [C.SOURCE_FACTORS_COMBINE_SUM, C.SOURCE_FACTORS_COMBINE_AVERAGE]
+        check_condition(all([combine in [C.FACTORS_COMBINE_SUM, C.FACTORS_COMBINE_AVERAGE]
                              for combine in args.source_factors_combine])
                         or len(source_vocabs) == len(args.source_factors_num_embed) + 1,
                         "Data was prepared with %d source factors, but only provided %d source factor dimensions." % (
                             len(source_vocabs), len(args.source_factors_num_embed) + 1))
+        check_condition(all([combine in [C.FACTORS_COMBINE_SUM, C.FACTORS_COMBINE_AVERAGE]
+                             for combine in args.target_factors_combine])
+                        or len(target_vocabs) == len(args.target_factors_num_embed) + 1,
+                        "Data was prepared with %d source factors, but only provided %d source factor dimensions." % (
```
tdomhan, line 321: I think this is meant to say `target`.

sockeye/inference.py
fhieber: `target_ids` is multi-dimensional, i.e. a nested list. So we check on the primary factor whether its a stop id, but we extend correctly.

sockeye/train.py
fhieber: good catch. I had hoped to get through this without a copy&paste error, but seems like this is impossible :)

sockeye/inference.py
tdomhan: ohh ok, got it, makes sense!

sockeye/train.py
tdomhan: To be honest I was looking for source/target mismatches somewhere, but I'm quite impressed that I haven't been able to find any :)

sockeye/inference.py
fhieber: True its a bit redundant. I moved the above line into the lower else-block. Can you review this carefully? This is quite complicated unfortunately.

tdomhan: It could be moved outside of the if block even couldn't it?

sockeye/score.py
```diff
@@ -73,13 +73,18 @@ def score(args: argparse.Namespace):
 
         sources = [args.source] + args.source_factors
         sources = [str(os.path.abspath(source)) for source in sources]
-        target = os.path.abspath(args.target)
+        targets = [args.target] + args.target_factors
+        targets = [str(os.path.abspath(target)) for target in targets]
+
+        check_condition(len(targets) == model.num_target_factors,
```
tdomhan, line 79: The check needs to be moved up before creating the paths as currently one gets a `TypeError: expected str, bytes or os.PathLike object, not NoneType` when no target factors are specified.

sockeye/inference.py
fhieber: yes, will change

sockeye/train.py
```diff
@@ -497,20 +554,46 @@ def get_num_embed(args: argparse.Namespace) -> Tuple[int, int]:
                             "has to be even and a multiple of encoder attention heads (%d)" % (
                                 adjusted_transformer_encoder_model_size, args.transformer_attention_heads[0]))
 
+    if not num_embed_source:
+        num_embed_source = C.DEFAULT_NUM_EMBED
+
     if args.decoder == C.TRANSFORMER_TYPE:
```
tdomhan, line 560: mhhh btw how does this work with SSRU decoders?
This check seems to be a left over from Sockeye 1 where we branched on Transformers vs RNNs, right?
Would it make sense to remove it?

sockeye/train.py
```diff
@@ -497,20 +554,46 @@ def get_num_embed(args: argparse.Namespace) -> Tuple[int, int]:
...
             num_embed_target = transformer_model_size_target
         else:
             # Make sure that if the user sets num_embed it matches the Transformer model size
-            check_condition(args.transformer_model_size[1] == num_embed_target,
+            check_condition(args.transformer_model_size[1] == num_embed_target + total_target_factor_size,
                             "Target embedding size must match transformer model size: %s vs. %s"
-                            % (args.transformer_model_size[1], num_embed_target))
+                            % (args.transformer_model_size[1], num_embed_target + total_target_factor_size))
+
+        total_target_factor_size = 0
```
tdomhan, line 577: This block seems to be duplicate compared to lines 562 to 566, right? (maybe I'm missing something)

sockeye/inference.py
```diff
@@ -1100,22 +1146,43 @@ def _assemble_translation(sequence: np.ndarray,
...
         :param length: The length of the translated segment.
         :param seq_score: Array of length-normalized negative log-probs.
         :param estimated_reference_length: Estimated reference length (if any).
         :param beam_history: The optional beam histories for each sentence in the batch.
         :return: A Translation object.
         """
+        if unshift_target_factors:
+            sequence = _unshift_target_factors(sequence, fill_last_with=C.EOS_ID)
+        else:
+            sequence = sequence.tolist()
```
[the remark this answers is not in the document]
fhieber: Yes, but I wanted to pave the way for potentially allowing setting whether to shift or not in the future.

sockeye/train.py
fhieber: good catch. I'll remove the branch, also for encoders

sockeye/train.py
fhieber: good catch, must be a leftover from a previous merge. removed the second block.

sockeye/train.py
tdomhan: thanks!

sockeye/inference.py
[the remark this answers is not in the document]
tdomhan: makes sense!

CHANGELOG.md
```diff
@@ -11,17 +11,30 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.3.0]
```
tdomhan, line 14: Maybe we could add that source factors changed from concat to sum under a `Changed` section?

tdomhan: Maybe also worth pointing out that `load_models`'s return types have changed as this is kind of the inference API some may interact with.

tdomhan (approved):
I left some suggestions for the CHANGELOG. Otherwise LGTM! Thanks :)

CHANGELOG.md
fhieber: done!

PR #848, opened 2020-07-28
Parallelize token frequency calculation

Deseaus:
Make token frequency calculation parallel.

Currently relies on a couple of functions from [more-itertools](https://more-itertools.readthedocs.io/en/stable/index.html). We could also take a dependency on the package instead if that's desirable. It's even [endorsed](https://docs.python.org/3/library/itertools.html#itertools-recipes) by the official Python docs.

until you can check this box.
- [X] Updated CHANGELOG.md

### DIFF
--- a/sockeye/utils.py
+++ b/sockeye/utils.py
@@ -27,7 +27,8 @@ import random
 import sys
 import time
 from contextlib import contextmanager, ExitStack
-from functools import reduce
+from functools import partial, reduce
+from itertools import islice
 from typing import Any, List, Iterator, Iterable, Set, Tuple, Dict, Optional, Union, IO, TypeVar, cast
 
 import mxnet as mx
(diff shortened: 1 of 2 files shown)


tdomhan:
Thanks! What speed-ups are you seeing from this?

Deseaus:
You're right, I only got a small improvement with 4 processes, close to none with more, depending on how the chunking params are tweaked. This particular functionality does take a lot of time though, do we need a frequency-ranked vocabulary on the full data? Maybe we could take a different approach and offer to optionally calculate it on a sample of the data only?

tdomhan:
Thanks, that makes sense. I think basing this on a subset of the data does sound quite reasonable as a way to cut down the time. The other option from a Sockeye user perspective is to build vocabularies outside of Sockeye (say on a subsample) and then import those into Sockeye. We could of course support this directly in Sockeye without the additional step through a separate CLI.

Deseaus:
Closing for now, as this doesn't seem like a promising avenue to speed-up the vocabulary building. We'd need to subsample the data or move to something like Dask that can read a large file in chunks for parallel processing.

PR #850, opened 2020-08-06
Fix: Guard against null division for small batch sizes.

tdomhan:
For small batch sizes we could end up with batch sizes of 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.2.2]
+
+### Fixed
+ - Fix: Guard against null division for small batch sizes.
+
 ## [2.2.1]
 ## Fixed
 
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '2.2.1'
+__version__ = '2.2.2'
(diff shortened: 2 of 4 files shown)


mjdenkowski:
Good catch!

Instead of automatically increasing the batch size, how about raising an error like we do for sequence length ([data_io.py#L209-L211](https://github.com/awslabs/sockeye/blob/2cbad615d1f57ce3fd51063a998735eb6249ae91/sockeye/data_io.py#L209-L211))?

This would preserve the "always round down" property of `max-word` batching.  Since this error case is more likely for longer sequence lengths, requiring the user to actively increase the size of all batches would also avoid the issue of oversized batches for longer sequences.

tdomhan:
The batch_size would only be changed in the case of batch_size being 0 to ensure it's at least `min_batch_step = batch_sentences_multiple_of * batch_num_devices`. For word based batching we have similar logic to make sure to never go below `min_batch_step`: https://github.com/awslabs/sockeye/blob/2cbad615d1f57ce3fd51063a998735eb6249ae91/sockeye/data_io.py#L207

I'd say it's still reasonable to adjust the batch size to the minimum possible instead of raising an error in this edge cased.

fhieber:
Fair point, @mjdenkowski should we merge this as is?

mjdenkowski:
The case we want to avoid is silently changing users' settings and leaving them confused about why Sockeye is behaving in a particular way.  The `max_word` batching strategy is specifically intended to be as predictable as possible whereas `word` rounds up or down and increases batch sizes as needed.

I still recommend raising an error, but if @tdomhan and @fhieber prefer the current version of the PR, a middle ground would be adding a log message such as:
`The batch size for bucket (N, M) has been increased from X to Y to avoid rounding down to 0.  To avoid automatic per-bucket batch resizing, specify a larger value for --batch-size.`

tdomhan:
I changed to an assertion and rebase based on the current master.

fhieber (approved):

PR #851, opened 2020-08-12
Add SSRU layer and decoder

barbaradarques:
- Adds `layers.SSRU`, which implements a Simpler Simple Recurrent Unit as described by Kim et al, 2019.

- Adds `ssru_transformer` option to `--decoder`, which enables the usage of SSRUs as a replacement for the decoder-side self-attention layers.

- Reduces the number of arguments for `MultiHeadSelfAttention.hybrid_forward()`. 
 `previous_keys` and `previous_values` should now be input together as `previous_states`, a list containing two symbols.

until you can check this box.
- [x] Updated CHANGELOG.md

### Notes

* I have run a local seqcopy test of the regular transformer with and without the refactoring and verified that it reaches the same exact perplexity in both cases.


sockeye/__init__.py
```diff
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '2.1.16'
```
tdomhan, line 14: All changes are backwards compatible, meaning I could use a model trained with 2.1.x and it would still work with the latest code, right? If so we shouldn't bump the major version to not invalidate the already trained models of others if at all possible.

barbaradarques: Doesn't the change in the arguments of `MultiheadSelfAttention.hybrid_forward()` cause backwards incompatibility?

barbaradarques: Oh, got it. I had misunderstood what backwards compatibility meant here.

sockeye/decoder.py
```diff
@@ -202,14 +206,13 @@ def init_state_from_encoder(self,
...
-        self_att_key_value_dummies = [mx.nd.zeros((batch_size,
-                                                   self.config.attention_heads,
-                                                   1,
-                                                   self.config.model_size // self.config.attention_heads),
-                                                  ctx=encoder_outputs.context,
-                                                  dtype=encoder_outputs.dtype)] * self.config.num_layers * 2
-        states += self_att_key_value_dummies
+        autoregr_state_shape = self.layers[0].get_states_shape(batch_size)
+        dummy_autoregr_states = [mx.nd.zeros(autoregr_state_shape,
+                                                 ctx=encoder_outputs.context,
```
fhieber, line 211: indentation seems off

sockeye/decoder.py
```diff
@@ -311,18 +311,21 @@ def hybrid_forward(self, F, step_input, states):
...
-                                                           mask,
-                                                           source_encoded,
-                                                           source_mask,
-                                                           self_att_k, self_att_v,
-                                                           enc_att_k, enc_att_v)
-            new_self_att_kv += [new_self_att_k, new_self_att_v]
+        new_autoregr_states = []
+        for layer, layer_autoregr_state, (enc_att_k, enc_att_v) in zip(self.layers, autoregr_states, enc_att_kv):
+            target, new_layer_autoregr_state = layer(target,
+                                                      mask,
```
fhieber, line 317: indentation is off by one :)

sockeye/layers.py
```diff
@@ -22,6 +23,7 @@
 from . import constants as C
 from . import quantization
 from . import utils
+from .constants import SSRU_PREFIX
```
fhieber, line 26: [nit] I'd prefer using `C.SSRU_PREFIX` from the above import.

sockeye/layers.py
```diff
@@ -732,3 +796,126 @@ def hybrid_forward(self, F, data, steps, weight):  # pylint: disable=arguments-d
...
+
+            self.linear = quantization.QuantizableDense(in_units=model_size,
+                                                        units=model_size,
+                                                        use_bias=False,
+                                                        flatten=False,
+                                                        prefix="linear_",
+                                                        dtype=dtype)
+
+    @property
+    def prefix(self) -> str:
```
fhieber, line 853: this does overwrite the `HybridBlock.prefix` attribute, does it? Fine with me, especially because MXNet 2.0 won't have Block prefixes anymore.

sockeye/transformer.py
```diff
@@ -120,29 +122,41 @@ def hybrid_forward(self, F, data: mx.sym.Symbol, bias: mx.sym.Symbol) -> mx.sym.
...
     with pre/post process blocks in between.
     """
 
     def __init__(self,
                  config: TransformerConfig,
+                 inference_only: bool,
                  prefix: str,
                  dtype: str) -> None:
         super().__init__(prefix=prefix)
+        self.decoder_type = config.decoder_type
```
fhieber, line 135: do we need that attribute? Accessing `config` below just like for the other arguments seems fine to me.

test/integration/test_seq_copy_int.py
```diff
@@ -42,9 +42,9 @@
 _TEST_MAX_LENGTH = 20
 
 # tuple format: (train_params, translate_params, use_prepared_data, use_source_factors)
-ENCODER_DECODER_SETTINGS = [
+ENCODER_DECODER_SETTINGS_TEMPLATE = [
     # Basic transformer, nbest=2 decoding
-    ("--encoder transformer --decoder transformer"
+    ("--encoder transformer --decoder {decoder}"
```
fhieber, line 47: what is the additional runtime of integration tests with this cross-product of test x decoder types?

tdomhan:
> Would it make sense to add or change an existing system test to use SSRUs to make sure it converges well?

+1 would make sense.

sockeye/transformer.py
barbaradarques: Definitely, I think it's a leftover from some previous iteration. Good catch!

sockeye/layers.py
barbaradarques: Good point, it does overwrite that.

test/integration/test_seq_copy_int.py
barbaradarques: Unit/integration tests last 15-20s with both, so I don't think that it's significant. The system tests, however, take about 8 minutes for each decoder.

barbaradarques:
> Would it make sense to add or change an existing system test to use SSRUs to make sure it converges well?

I added an equivalent SSRU test for each regular transformer case, but they are all failing. Perplexity-wise they seem ok, but the BLEU scores are very off (e.g. 0.6). Am I missing something?

sockeye/layers.py
```diff
@@ -732,3 +795,126 @@ def hybrid_forward(self, F, data, steps, weight):  # pylint: disable=arguments-d
...
+        cell_state, last_step_state = F.contrib.foreach(_time_step_update,
+                                                        [weighted_inputs, forget_rates],
+                                                        previous_cell_state)
+
+        return F.transpose(cell_state, axes=(1, 0, 2)), last_step_state
+
+    @staticmethod
+    def _inference_cell_state_transform(F, previous_cell_state, weighted_inputs, forget_rates) -> Tuple:
+        """Update SSRU cell at inference time"""
+        new_step_state = forget_rates * previous_cell_state + weighted_inputs  # (batch, 1, input_depth)
```
barbaradarques, line 904: I had accidentally replaced a ```F.elemwise_add()`` with `*`, causing inference updates to be different than at training time. Hence the right perplexity at training and the bad BLEU at inference.

fhieber: Good catch!

fhieber (approved):
Thanks for iterating!

CHANGELOG.md
```diff
@@ -11,6 +11,19 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.1.17]
+
+### Added
+
+- Added `layers.SSRU`, which implements a Simpler Simple Recurrent Unit as described by Kim et al, 2019.
```
tdomhan, line 18: Maybe add the full paper title here and/or in the SSRU class?

sockeye/decoder.py
```diff
@@ -202,14 +206,13 @@ def init_state_from_encoder(self,
...
-                                                  ctx=encoder_outputs.context,
-                                                  dtype=encoder_outputs.dtype)] * self.config.num_layers * 2
-        states += self_att_key_value_dummies
+        autoregr_state_shape = self.layers[0].get_states_shape(batch_size)
+        dummy_autoregr_states = [mx.nd.zeros(autoregr_state_shape,
+                                             ctx=encoder_outputs.context,
+                                             dtype=encoder_outputs.dtype)
+                                 for _ in range(self.config.num_layers * self.layers[0].num_state_tensors)]
+
+        states += dummy_autoregr_states
```
tdomhan, line 215: I think we should either check that `num_state_tensors` is the same across layers or directly use individual layers instead of `layers[0]`:
```suggestion
        dummy_autoregr_states = [mx.nd.zeros(layer.get_states_shape(batch_size),
                                             ctx=encoder_outputs.context,
                                             dtype=encoder_outputs.dtype)
                                 for layer in self.layers for _ in range(0, layer.num_state_tensors)]

        states += dummy_autoregr_states
```

sockeye/decoder.py
```diff
@@ -269,39 +272,36 @@ def forward(self, step_input, states):
...
+            autoregr_states = other[self.config.num_layers * 2:]
         else:
-            mask = self.autoregressive_bias(step_input)  # mask: (1, length, length)
-
-            steps, source_encoded, source_mask, *other = states
-
-            self_att_kv = other
-            self_att_kv = [self_att_kv[i:i + 2] for i in range(0, len(self_att_kv), 2)]
-
+            if any(layer.needs_mask for layer in self.layers):
```
tdomhan, line 295: nice, thanks for generalizing this :)

sockeye/decoder.py
```diff
@@ -269,39 +272,36 @@ def forward(self, step_input, states):
...
+                mask = self.autoregressive_bias(step_input)  # mask: (1, length, length)
+            steps, source_encoded, source_mask, *autoregr_states = states
             enc_att_kv = [(None, None) for _ in range(self.config.num_layers)]
 
+        if any(layer.num_state_tensors != 1 for layer in self.layers):
```
tdomhan, line 300: Would it be cleaner to say `> 1`?

sockeye/decoder.py
```diff
@@ -311,18 +311,21 @@ def hybrid_forward(self, F, step_input, states):
...
+        for layer, layer_autoregr_state, (enc_att_k, enc_att_v) in zip(self.layers, autoregr_states, enc_att_kv):
+            target, new_layer_autoregr_state = layer(target,
+                                                     mask,
+                                                     source_encoded,
+                                                     source_mask,
+                                                     layer_autoregr_state,
+                                                     enc_att_k, enc_att_v)
+
+            new_autoregr_states += [*new_layer_autoregr_state]
+            # NOTE: the list expansion is needed in order to handle both a tuple (of Symbols) and a Symbol as a new state
```
tdomhan, line 324: Thanks for adding!

sockeye/layers.py
```diff
@@ -418,7 +419,45 @@ def _attend(self,
         return contexts
 
 
-class MultiHeadSelfAttention(MultiHeadAttentionBase):
+class AutoregressiveLayer(mx.gluon.HybridBlock):
```
tdomhan, line 422: Thanks for renaming!

tdomhan (commented):
Thanks for iterating!

sockeye/decoder.py
barbaradarques: Definitely, I missed that.

CHANGELOG.md
barbaradarques: Sure.

tdomhan (approved):

tdomhan:
Thanks! LGTM.

PR #878, opened 2020-09-11
Re-introduce softmax with length in DotAttention (see #772)

fhieber:
Re-introduces #772 which was reverted after discovering an incompatibility with `broadcast_like` in MXNet 1.6. This is fixed in MXNet 1.7, therefore, this PR is based on the `mx17` branch.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/encoder.py
+++ b/sockeye/encoder.py
@@ -308,9 +308,6 @@ class TransformerEncoder(Encoder, mx.gluon.HybridBlock):
                                                              prefix=C.SOURCE_POSITIONAL_EMBEDDING_PREFIX,
                                                              scale_up_input=True,
                                                              scale_down_positions=False)
-            self.valid_length_mask = transformer.TransformerValidLengthMask(num_heads=self.config.attention_heads,
-                                                                            fold_heads=True,
-                                                                            name="bias")
 
             self.layers = mx.gluon.nn.HybridSequential()
             for i in range(config.num_layers):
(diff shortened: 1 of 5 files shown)


sockeye/encoder.py
```diff
@@ -328,11 +325,11 @@ def hybrid_forward(self, F, data, valid_length):
         if self.config.dropout_prepost > 0.0:
             data = F.Dropout(data=data, p=self.config.dropout_prepost)
 
-        # (batch_size * heads, 1, seq_len)
-        bias = F.expand_dims(self.valid_length_mask(data, valid_length), axis=1)
```
barbaradarques, line 332: I like that the name is now consistent across classes. :)

sockeye/layers.py
```diff
@@ -329,23 +304,17 @@ def hybrid_forward(self, F, queries, keys, values, lengths=None, bias=None):
...
-            # (n, lq, lk)
-            logits = F.transpose(data=logits, axes=(1, 2, 0))
-
         if bias is not None:
+            assert lengths is None, "Cannot provide both bias and lengths"
             logits = F.broadcast_add(logits, bias)
 
-        probs = F.softmax(logits, axis=-1)
+        if lengths is not None:
+            assert bias is None, "Cannot provide both bias and lengths"
```
barbaradarques, line 312: Wouldn't it it be better to assert this only once and in the beginning of the method? e.g.
```python
assert not (bias and lengths), "Cannot provide both bias and lengths"
```

fhieber: Yeah, that would be a bit cleaner. Actually, there is no reason to have this assert really. Of course it doesn't really make sense to use both 'modes' usually, but why not allow it?

barbaradarques (approved):

PR #893, opened 2020-10-14
Use arange_like in AutoRegressiveBias block to reduce overall number of ops

fhieber:
Addition of `arange_like` since MXNet 1.6 allows us to simplify AutoRegressiveBias a little bit.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/transformer.py
+++ b/sockeye/transformer.py
@@ -372,7 +372,7 @@ class TransformerValidLengthMask(mx.gluon.HybridBlock):
 
 
 class AutoRegressiveBias(mx.gluon.HybridBlock):
-    def __init__(self, prefix: str = '',) -> None:
+    def __init__(self, prefix: str = '') -> None:
         super().__init__(prefix=prefix)
         self._dtype = 'float32'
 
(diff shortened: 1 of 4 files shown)


sockeye/transformer.py
```diff
@@ -381,15 +381,12 @@ def cast(self, dtype):
...
-        length_array = F.cast(F.contrib.index_array(x, axes=(1,)), dtype=self._dtype)
+        # Shape: (length, 1)
+        length_array = F.contrib.arange_like(x, axis=1)
         # matrix with lower triangle and main diagonal set to 0, upper triangle set to 1
         # Shape: (length, length)
-        bias = F.broadcast_greater(F.reshape(length_array, shape=(1, -1)),
-                                   length_array)
+        bias = F.broadcast_greater(F.expand_dims(length_array, axis=0),
+                                   F.expand_dims(length_array, axis=1))
         bias = bias * -C.LARGE_VALUES[self._dtype]
```
barbaradarques, line 390: [Curiosity only] Why is this necessary? I read the docs and saw that for np.float16 it will move it to some multiple of 32 and that for np.float32 it will move to 1.0e8. Is the goal to keep the np.float32 masked logits intact and round them up when using np.float16?

barbaradarques (approved):

sockeye/transformer.py
fhieber: If training in FP16 you want to avoid overflowing the numerical range, therefore the max value for FP16 is different than for FP32.

PR #894, opened 2020-10-14
Fix amp issues with using length argument for softmax in attention

fhieber:
Re-introduce (again and again) the use of the length argument in the softmax call in DotAttentionCell.
Previously this caused issues with Automatic Mixed Precision training as broadcast_like does not like different dtypes for its input arguments. While this has been fixed in MXNet master, it was not yet merged back to the 1.x branches, therefore it is not part of MXNet 1.6 or 1.7 (or 1.8).

However, we can move the creation of the length tensor for softmax to a different place which avoids the differing dtype problem. 

This change is confirmed to be working with AMP and giving a slight increase in training and inference speed as previously measured.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -295,15 +295,9 @@ class TransformerDecoder(Decoder, mx.gluon.HybridBlock):
             states_iter = iter(autoregr_states)
             autoregr_states = [list(islice(states_iter, 0, layer.num_state_tensors)) for layer in self.layers]
 
-        # prepare encoder attention valid length tensor
-        # (batch * heads,)
-        att_valid_length = F.repeat(att_valid_length, repeats=self.config.attention_heads, axis=0)
-        # (batch * heads, query_length)
-        att_valid_length = F.broadcast_like(F.expand_dims(att_valid_length, axis=1),
-                                            step_input,
-                                            lhs_axes=(1,),
-                                            rhs_axes=(1,))
-        att_valid_length = F.cast(att_valid_length, dtype='int32')
+        # (batch_size * heads, query_length)
+        att_valid_length = layers.prepare_softmax_lengths(F, att_valid_length, step_input,
+                                                          num_heads=self.config.attention_heads)
 
         # target: (batch_size, length, model_size)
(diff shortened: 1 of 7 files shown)


sockeye/decoder.py
```diff
@@ -286,23 +282,24 @@ def forward(self, step_input, states):
     def hybrid_forward(self, F, step_input, states):
         mask = None
         if self.inference_only:
-            steps, source_mask, *other = states
+            steps, att_valid_length, *other = states
```
ketranm, line 285: Is `att_valid_length` the valid length of source sequence. Maybe we can made it more explicit (e.g., `source_att_valid_length`)

sockeye/decoder.py
```diff
@@ -286,23 +282,24 @@ def forward(self, step_input, states):
...
 
         if any(layer.num_state_tensors > 1 for layer in self.layers):
             # separates autoregressive states by layer
             states_iter = iter(autoregr_states)
             autoregr_states = [list(islice(states_iter, 0, layer.num_state_tensors)) for layer in self.layers]
 
-        # Fold the heads of source_mask (batch_size, num_heads, seq_len) -> (batch_size * num_heads, 1, seq_len)
-        source_mask = F.expand_dims(F.reshape(source_mask, shape=(-3, -2)), axis=1)
+        # (batch_size * heads, query_length)
+        att_valid_length = layers.prepare_softmax_lengths(F, att_valid_length, step_input,
```
ketranm, line 301: I can guess what this function does, but its name is a bit ambiguous.

sockeye/decoder.py
fhieber: good point, renamed

sockeye/decoder.py
fhieber: Do you have a better suggestion? I can't think of a better alternative.

ketranm: I can only think of a boring name `prepare_source_valid_lengths` :)

fhieber: ok will do

ketranm (approved):

PR #896, opened 2020-10-14
Reduce calls to ndarray.shape and number of casts

fhieber:
Each call to `ndarray.shape` is a library call to MXNet. This PR reduces the number of calls in some of the critical code paths.

Also, return topk indices directly as int, avoiding a separate casting.

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -603,20 +603,20 @@ class BeamSearch(mx.gluon.Block):
                                      mode='constant', constant_values=self.eos_id)
             vocab_slice_ids = mx.nd.array(vocab_slice_ids, ctx=self.context, dtype='int32')
 
-            vocab_slice_ids_shape = vocab_slice_ids.shape
-            if vocab_slice_ids_shape[0] < self.beam_size + 1:
+            vocab_slice_ids_shape = vocab_slice_ids.shape[0]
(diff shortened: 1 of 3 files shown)


sockeye/beam_search.py
```diff
@@ -600,22 +600,23 @@ def forward(self,
...
+                n = self.beam_size - vocab_slice_ids_shape[0] + 1
                 vocab_slice_ids = mx.nd.concat(vocab_slice_ids,
                                                mx.nd.full((n,), val=self.eos_id, ctx=self.context, dtype='int32'),
                                                dim=0)
 
-            pad_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids.shape[0] - 1),
+            pad_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids_shape[0] - 1),
                                   val=np.inf, ctx=self.context)
-            eos_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids.shape[0]),
+            eos_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids_shape[0]),
```
ketranm, line 619: only `vocab_slice_ids_shape[0]` is needed, isn't it? Why not making it a scalar?

fhieber: good point!

ketranm (approved):

PR #900, opened 2020-10-29
Avoid edge case in _adjust_learning_rate.

tdomhan:

### DIFF
--- a/sockeye/training.py
+++ b/sockeye/training.py
@@ -527,7 +527,8 @@ class GluonEarlyStoppingTrainer:
                             self.state.best_checkpoint)
                 adjusted_lr = self.trainer.optimizer.lr_scheduler.lr
                 # trainer.load_states also reloads the parameters
-                self._load_trainer_states(self.best_optimizer_states_fname)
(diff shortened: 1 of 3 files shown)


sockeye/training.py
```diff
@@ -527,7 +527,8 @@ def _adjust_learning_rate(self, has_improved: bool):
                             self.state.best_checkpoint)
                 adjusted_lr = self.trainer.optimizer.lr_scheduler.lr
                 # trainer.load_states also reloads the parameters
-                self._load_trainer_states(self.best_optimizer_states_fname)
+                if os.path.exists(self.best_optimizer_states_fname):
+                    self._load_trainer_states(self.best_optimizer_states_fname)
```
tdomhan, line 531: `self.best_optimizer_states_fname` only gets written to if `has_improved` is true in `fit()`. For the edge case where has_improved has never been True before the first time the learning rate is reduced training would crash as the file is not found..

fhieber: nice catch

fhieber (approved):

PR #901, opened 2020-10-30
Log level for secondary workers.

tdomhan:
Allow specifying a different log level for secondary workers.

### DIFF
--- a/sockeye/log.py
+++ b/sockeye/log.py
@@ -130,8 +130,8 @@ def setup_main_logger(file_logging=True, console=True, path: Optional[str] = Non
     for _, handler_config in log_config['handlers'].items():  # type: ignore
         handler_config['level'] = level
 
-    if 'console' in log_config['handlers'] and console_level is not None:
-        log_config['handlers']['console']['level'] = console_level
+    if 'console' in log_config['handlers'] and console_level is not None:  # type: ignore
+        log_config['handlers']['console']['level'] = console_level  # type: ignore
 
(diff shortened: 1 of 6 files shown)


fhieber (approved):
thanks!

mjdenkowski:
Thanks for adding this functionality!

I recommend that the default log level for secondary workers is the same as for the primary worker.  This creates a lot of log lines, but it lets the user know that all processes are running as expected.  This is a useful default behavior for when users are first setting up Sockeye, Horovod, and MPI on one or more hosts.  After the user confirms everything is working, they can specify one of the options to reduce console logging.  This is the approach we use for the `--quiet-secondary-workers` option.

fhieber:
Looks like `test_arguments` requires an update for the checks to pass.

sockeye/log.py
```diff
@@ -128,6 +130,9 @@ def setup_main_logger(file_logging=True, console=True, path: Optional[str] = Non
     for _, handler_config in log_config['handlers'].items():  # type: ignore
         handler_config['level'] = level
 
+    if 'console' in log_config['handlers'] and console_level is not None:
```
fhieber, line 133: you might need a `# type: ignore` somewhere here for mypy to pass

fhieber (changes requested):

sockeye/log.py
tdomhan: Yeah, I did that but hadn't noticed that my push was rejected because I needed to pull the master merge :)

fhieber (approved):

PR #902, opened 2020-11-01
Disable LHUC+SSRU integration test

fhieber:
Fixes occasional integration test failure that uses LHUC + SSRUs. LHUC by default fixes all other parameters, making only lhuc params trainable. For the integration tests (which only train for a few updates), this sometimes leads to divergence. Disabled this integration test for now.

### DIFF
--- a/test/integration/test_seq_copy_int.py
+++ b/test/integration/test_seq_copy_int.py
@@ -80,7 +80,7 @@ ENCODER_DECODER_SETTINGS_TEMPLATE = [
      "--beam-size 2 --beam-search-stop first",
      True, 3),
     # Basic transformer with LHUC
-    ("--encoder transformer --decoder {decoder}"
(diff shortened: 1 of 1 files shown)


test/integration/test_seq_copy_int.py
```diff
@@ -80,7 +80,7 @@
      "--beam-size 2 --beam-search-stop first",
      True, 3),
     # Basic transformer with LHUC
-    ("--encoder transformer --decoder {decoder}"
+    ("--encoder transformer --decoder transformer"
```
tdomhan, line 83: Oh ok, so we'd just run it twice with Transformer.

fhieber: Yes, seems like the easiest solution for now

tdomhan (approved):

PR #906, opened 2020-11-05
Fix horovod and bleu/chrf as optimized metrics.

tdomhan:
With this PR we make sure to also create checkpoint decoders on secondary workers when Horovod is enabled as otherwise training will fail.

The reason is that otherwise secondary workers will not find the evaluation metric and fail here: https://github.com/awslabs/sockeye/blob/master/sockeye/training.py#L442

### DIFF
--- a/sockeye/train.py
+++ b/sockeye/train.py
@@ -204,16 +204,16 @@ def create_checkpoint_decoder(
     :return: A CheckpointDecoder if --decode-and-evaluate != 0, else None.
     """
     sample_size = args.decode_and_evaluate
-    if args.optimized_metric == C.BLEU and sample_size == 0:
-        logger.info("You chose BLEU as the optimized metric, will turn on BLEU monitoring during training. "
-                    "To control how many validation sentences are used for calculating bleu use "
-                    "the --decode-and-evaluate argument.")
+    if args.optimized_metric in C.METRICS_REQUIRING_DECODER and sample_size == 0:
+        logger.info(f"You chose {args.optimized_metric} as the optimized metric, will turn on BLEU monitoring during training. "
+                     "To control how many validation sentences are used for calculating bleu use "
+                     "the --decode-and-evaluate argument.")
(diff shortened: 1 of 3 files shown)


sockeye/train.py
```diff
@@ -204,16 +204,16 @@ def create_checkpoint_decoder(
     :return: A CheckpointDecoder if --decode-and-evaluate != 0, else None.
     """
     sample_size = args.decode_and_evaluate
-    if args.optimized_metric == C.BLEU and sample_size == 0:
-        logger.info("You chose BLEU as the optimized metric, will turn on BLEU monitoring during training. "
+    if args.optimized_metric in C.METRICS_REQUIRING_DECODER and sample_size == 0:
+        logger.info("You chose %s as the optimized metric, will turn on BLEU monitoring during training. "
```
fhieber, line 208: `[...] will turn on %s monitoring during training`?

tdomhan: ah yeah, ups!

mjdenkowski:
When using Horovod, the primary worker makes the authoritative check for metric improvement and broadcasts the result to other workers ([training.py#L479-L485](https://github.com/awslabs/sockeye/blob/master/sockeye/training.py#L479-L485)).

Having each worker run its own checkpoint decoder should be compatible with this as long as we're still populating `self.state.best_metric` and `self.state.best_metric_history`.

Can you confirm that this is the case?

tdomhan:
That's right. So basically when we have one checkpoint decoder per worker we make sure that the chrf/bleu metrics are populated in the list `val_metrics` passed to `_determine_improvement`. The actual logic that synchronizes across workers and that sets `best_metric` is actually identical to other metrics such as perplexity. The bleu/chrf metrics are added in `_evaluate` if the `checkpoint_decoder` object is provided.

fhieber (approved):

mjdenkowski (approved):
Thanks Tobi!

Issue #913, opened 2020-11-10
Upgrade to newest version of sacrebleu


bricksdont:
Sockeye currently pins the sacrebleu dependency to an older version, `sacrebleu==1.4.10`.

I would like to install the most recent sacebleu (1.4.14 at the moment) because it introduced TER, but this breaks the Sockeye checkpoint decoder code due to backward-incompatible changes to the sacrebleu API.

Adapting those calls in Sockeye would only be a minor change, and I would like to open a PR for this.

Is there another reason for pinning the sacrebleu version to 1.4.10?

tdomhan:
No, I think that's the only reason, namely that we want to avoid API backwards incompatabilities to be automatically pulled in. If you update and adjust the calls that sounds great!

PR #914, opened 2020-11-10
Upgrade sacrebleu version (fixes issue #913)

bricksdont:
Upgrading to the most recent release of sacrebleu, with minor changes to Sockeye (fixes https://github.com/awslabs/sockeye/issues/913).

The major addition in this version of sacrebleu is TER as a metric, but in my view, it is probably not worth adding this to the list of Sockeye metrics.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/evaluate.py
+++ b/sockeye/evaluate.py
@@ -54,8 +54,7 @@ def raw_corpus_chrf(hypotheses: Iterable[str], references: Iterable[str]) -> flo
     :param references: Reference stream.
     :return: chrF score as float between 0 and 1.
     """
-    return sacrebleu.corpus_chrf(hypotheses, references, order=sacrebleu.sacrebleu.CHRF_ORDER, beta=sacrebleu.sacrebleu.CHRF_BETA,
-                                 remove_whitespace=True).score
+    return sacrebleu.corpus_chrf(hypotheses, [references]).score
 
 
 def raw_corpus_rouge1(hypotheses: Iterable[str], references: Iterable[str]) -> float:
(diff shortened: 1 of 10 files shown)


sockeye/evaluate.py
```diff
@@ -54,8 +54,7 @@ def raw_corpus_chrf(hypotheses: Iterable[str], references: Iterable[str]) -> flo
     :param references: Reference stream.
     :return: chrF score as float between 0 and 1.
     """
-    return sacrebleu.corpus_chrf(hypotheses, references, order=sacrebleu.sacrebleu.CHRF_ORDER, beta=sacrebleu.sacrebleu.CHRF_BETA,
-                                 remove_whitespace=True).score
+    return sacrebleu.corpus_chrf(hypotheses, [references]).score
```
fhieber, line 57: I know we fix the version of sacrebleu in the requirements, but I think its still worth making the settings here explicit by passing the keyword arguments.

bricksdont: Sure, I can add them back in. I thought that is redundant since those values are taken from sacrebleu anyway.

fhieber: yep, but they can change it and maybe forget to update the version, or we at some point don't fix the version anymore :)

bricksdont: Then wouldn't it be better to use actual numbers instead of a new way to access those variables in sacrebleu? (e.g. `sacrebleu.metrics.CHRF.ORDER`)

fhieber: True, I was only looking at the `remove_whitespace` flag really. Let's leave as you have it for now.

fhieber (approved):

PR #922, opened 2020-12-09
WIP: Masked pre-training and online back-translation

tdomhan:
This PR adds support for monolingual data. Monolingual data can be used either for masking ([MASS](https://arxiv.org/abs/1905.02450), [BART](https://arxiv.org/abs/1910.13461), DAE) or for online back-translation.

The PR is still in quite a rough state with several thing still outstanding. For example (non-exhaustive):

- [ ] Support for target factors.
- [ ] I'd like to revamp the CLI interface for monolingual data
- [ ] Reduce duplicate code in `data_io.py` by sharing code between monolingual and bilingual classes/functions.
- [ ] Add unit/system/integration tests + doc strings.
- [ ] Revise the way loss prefixing is done in training.py/loss.py
- [ ] Correct training resumption with monolingual iterators
- [ ] Use of monolingual data without prepared data folders
- [ ] ...

Because the PR is still WIP at this point I'd mainly like to get comments/suggestions on the high level structure (e.g. masking in the data iterator vs masking in the trainer, a potential decoupling of different 'tasks' in the trainer brought up by Felix in the past) and the CLI interface.

until you can check this box.
- [ ] Updated CHANGELOG.md

### DIFF
--- a/sockeye/arguments.py
+++ b/sockeye/arguments.py
@@ -399,6 +399,56 @@ def add_logging_args(params):
                                 help='Console log level for secondary workers. Default: %(default)s.')
 
 
+def add_mono_data_args(params, required=False):
+    params.add_argument('--monolingual', '-mo',
+                        required=required,
+                        type=regular_file(),
+                        help='Monolingual training data.')
+    params.add_argument('--monolingual-factors', '-mf',
+                        required=False,
+                        nargs='+',
+                        type=regular_file(),
+                        default=[],
+                        help='File(s) containing additional token-parallel factors for monolingual data. Default: %(default)s.')
+    params.add_argument('--mono-factors-use-mono-vocab',
+                        required=False,
+                        nargs='+',
+                        type=bool_str(),
+                        default=[],
+                        help='List of bools signaling whether to use the mono data vocabulary for the factors. '
+                        'If empty (default) each factor has its own vocabulary.')
+    params.add_argument('--monolingual-language', '-ml',
+                        required=False,
+                        type=str,
+                        default=None,
+                        help='The (optional) language of the monolingual data.')
+
+    params.add_argument('--mono-masking-style',
+                        required=False,
+                        type=str,
+                        choices=["MASS", "BART"],
+                        default="BART",
+                        help='The type of monolingual masking to use.')
+
+    params.add_argument('--use-bt-steps',
+                        required=False,
+                        type=bool,
+                        default=False)
+
+
+def add_bidirectional_language_args(params):
+    params.add_argument('--source-language', '-sl',
+                        required=False,
+                        type=str,
+                        default=None,
+                        help='The language of the source data.')
+    params.add_argument('--target-language', '-tl',
+                        required=False,
+                        type=str,
+                        default=None,
+                        help='The language of the target data.')
+
+
 def add_training_data_args(params, required=False):
     params.add_argument(C.TRAINING_ARG_SOURCE, '-s',
                         required=required,
@@ -434,6 +484,8 @@ def add_training_data_args(params, required=False):
                         required=required,
                         type=regular_file(),
                         help='Target side of parallel training data.')
+    add_bidirectional_language_args(params)
+    add_mono_data_args(params, required=False)
 
 
 def add_validation_data_params(params):
@@ -465,7 +517,10 @@ def add_prepared_data_args(params):
     params.add_argument(C.TRAINING_ARG_PREPARED_DATA, '-d',
                         type=regular_folder(),
                         help='Prepared training data directory created through python -m sockeye.prepare_data.')
-
+    params.add_argument(C.TRAINING_ARG_PREPARED_MONO_DATA, '-dm',
+                        type=regular_folder(),
+                        nargs='+',
+                        help='Prepared training data directory created through python -m sockeye.prepare_mono_data.')
 
 def add_monitoring_args(params):
     params.add_argument('--monitor-pattern',
@@ -531,11 +586,7 @@ def add_bucketing_args(params):
                              'length is X+1). Use "x:x" to specify separate values for src&tgt. Default: %(default)s.')
 
 
-def add_prepare_data_cli_args(params):
-    add_training_data_args(params, required=True)
-    add_vocab_args(params)
-    add_bucketing_args(params)
-
+def add_prepare_data_args(params):
     params.add_argument('--num-samples-per-shard',
                         type=int_greater_or_equal(1),
                         default=10000000,
@@ -559,6 +610,22 @@ def add_prepare_data_cli_args(params):
                         type=int_greater_or_equal(1),
                         default=1,
                         help='Process the shards in parallel using max-processes processes.')
+
+
+def add_prepare_data_cli_args(params):
+    add_training_data_args(params, required=True)
+    add_vocab_args(params)
+    add_bucketing_args(params)
+    add_prepare_data_args(params)
+
+    add_logging_args(params)
+
+
+def add_prepare_mono_data_cli_args(params):
+    add_mono_data_args(params, required=True)
+    add_mono_vocab_args(params)
+    add_bucketing_args(params)
+    add_prepare_data_args(params)
 
     add_logging_args(params)
 
@@ -635,6 +702,34 @@ def add_vocab_args(params):
                         default=(1, 1),
                         help='Minimum frequency of words to be included in vocabularies. Default: %(default)s.')
     params.add_argument('--pad-vocab-to-multiple-of',
+                        type=int,
+                        default=None,
+                        help='Pad vocabulary to a multiple of this integer. Default: %(default)s.')
+
+
+def add_mono_vocab_args(params):
+    params.add_argument('--mono-vocab',
+                        required=False,
+                        default=None,
+                        help='Existing vocabulary (JSON).')
+    params.add_argument('--mono-factor-vocabs',
+                        required=False,
+                        nargs='+',
+                        type=regular_file(),
+                        default=[],
+                        help='Existing factor vocabulary (-ies) (JSON).')
+    params.add_argument('--mono-num-words',
+                        type=int,
+                        default=0,
+                        help='Maximum vocabulary size. '
+                             'A value of 0 indicates that the vocabulary unrestricted and determined from the data by '
+                             'creating an entry for all words that occur at least --word-min-count times.'
+                             'Default: %(default)s.')
+    params.add_argument('--mono-word-min-count',
+                        type=int,
+                        default=1,
+                        help='Minimum frequency of words to be included in vocabularies. Default: %(default)s.')
+    params.add_argument('--mono-pad-vocab-to-multiple-of',
                         type=int,
                         default=None,
                         help='Pad vocabulary to a multiple of this integer. Default: %(default)s.')
@@ -792,6 +887,14 @@ def add_model_parameters(params):
     model_params.add_argument('--amp-scale-interval', type=int, default=2000,
                               help='Attempt to increase loss scale after this many updates without overflow. '
                                    'Default: %(default)s.')
+    model_params.add_argument('--encoder-lang-specific-layers',
+                              type=str,
+                              default=None,
+                              help="Specify the layers that will get language specific weights. Default: %(default)s.")
+    model_params.add_argument('--decoder-lang-specific-layers',
+                              type=str,
+                              default=None,
+                              help="Specify the layers that will get language specific weights. Default: %(default)s.")
 
 def add_batch_args(params, default_batch_size=4096, default_batch_type=C.BATCH_TYPE_WORD):
     params.add_argument('--batch-size', '-b',
@@ -1105,6 +1208,7 @@ def add_translate_cli_args(params):
     add_device_args(params)
     add_logging_args(params)
     add_hybridization_arg(params)
+    add_bidirectional_language_args(params)
 
 
 def add_score_cli_args(params):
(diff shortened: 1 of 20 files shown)


tdomhan:
Regarding the CLI interface what is currently supported is to prepare monolingual data separately from parallel data (using the same vocabulary) and then to specify multiple prepared data folders at training time. For example:
```
python -m sockeye.prepare_mono_data -mo docs/tutorials/seqcopy/data/dev.source -o prepared_mono_fr -ml fr
python -m sockeye.prepare_mono_data -mo docs/tutorials/seqcopy/data/dev.target -o prepared_mono_en -ml en
python -m sockeye.prepare_parallel_data -s docs/tutorials/seqcopy/data/dev.source -t docs/tutorials/seqcopy/data/dev.target -o prepared_en_fr -sl en -tl fr
python -m sockeye.train -d prepared_en_fr -dm prepared_mono_fr prepared_mono_en ....
```

This interface I'm quite happy with but open for other suggestions. What's currently missing from the code is a command line interface for using monolingual data without previously preparing data. Here I'd like to hear opinions on an interface. One option would be something like:
```
python -m sockeye.train -s file1 file2 -t file1 file2 -m file1 file2 \
--mono-language en fr \
--source-language en fr \
--target-language fr en
```
where the `-s`/`-t` arguments would be 'parallel' with the `--source-language` and `--target-language` parameters. 
We could also consider specifying the language inline:
```
python -m sockeye.train -s  en=file1 pt=file2 -t fr=file1 zh=file2 -m en=file1 fr=file2 
```
Source/target would still need to be parallel, but we'd get languages inline.

Thoughts?

fhieber:
I find the existing interface more intuitive than the inlining, which is also a bit cumbersome to compose in scripts or upstream applications.

sockeye/vocab.py
```diff
@@ -336,53 +363,82 @@ def load_or_create_vocabs(source_paths: List[str],
...
 
-    return [vocab_source] + vocab_source_factors, [vocab_target] + vocab_target_factors
+
+def load_or_create_mono_vocab(data_paths: List[str],
+                              vocab_paths: List[Optional[str]],
+                              factor_vocab_same_as_main: List[bool],
+                              num_words: Optional[int], word_min_count: int,
+                              pad_to_multiple_of: Optional[int] = None) -> List[Vocab]:
+    """
+    Returns vocabularies for source files (including factors) and target.
```
fhieber, line 414: mono not source/target  right?

sockeye/translate.py
```diff
@@ -79,6 +79,8 @@ def run_translate(args: argparse.Namespace):
                                     exit_stack=exit_stack)[0]
         logger.info("Translate Device: %s", context)
 
+        # TODO: add a check whether the source/target language arguments are required by the model and if so fail if they aren't
```
fhieber, line 82: +1, source target language should be a property of the model maybe?

sockeye/training.py
```diff
@@ -162,6 +164,41 @@ def __setstate__(self, state):
...
 
 
+def get_loss_prefix(prefix, source_lang, target_lang):
+    if source_lang is None and target_lang is None:
+        return prefix
+    else:
+        return f"{prefix}-{source_lang}-{target_lang}"
+
+
+def merge_metrics(all_metrics: Dict[str, List[loss.LossMetric]]) -> List[Tuple[str,float]]:
```
fhieber, line 174: this function seems to return a tuple

sockeye/training.py
```diff
@@ -193,11 +230,17 @@ def __init__(self,
...
+        # With language specific layers some parameters may be stale depenending on the value of the update_interval
+        self.ignore_stale_grad = self.model.config.config_encoder.lang_specific_layers or self.model.config.config_decoder.lang_specific_layers
 
     def fit(self,
             train_iter: data_io.BaseParallelSampleIter,
+            mono_iters: Dict[Language, data_io.MonoSampleIter],
             validation_iter: data_io.BaseParallelSampleIter,
-            checkpoint_decoder: Optional[CheckpointDecoder] = None):
+            checkpoint_decoder: Optional[CheckpointDecoder] = None,
+            # TODO: consider changing the interface for this. Should we just accept a list of steps? (where each step can be 'mt', 'mass' or 'bt'. 'mt' is on by default and 'mass' is added when monolingual data is available?)
```
fhieber, line 241: +1 for a list of steps, seems like a straight forward option

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
                         self.config.max_updates,
                         self.config.max_checkpoints)
 
+        # TODO: support en-fr and fr-en steps
```
fhieber, line 278: source/target

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
...
 
-            did_grad_step = self._step(batch=train_iter.next())
-            checkpoint_up_to_date = checkpoint_up_to_date and not did_grad_step
+            step, step_variant = next(steps_iter)
+            if step == "MT":
+                loss_prefix = get_loss_prefix("MT", step_variant[0].name, step_variant[1].name)
+                if step_variant == (source_lang, target_lang):
+                    did_grad_step = self._mt_step(batch=train_iter.next(), loss_prefix=loss_prefix)
+                elif step_variant == (target_lang, source_lang):
+                    did_grad_step = self._mt_step(batch=train_iter.next().reverse(), loss_prefix=loss_prefix)
```
fhieber, line 363: What is the effect of `reverse()` here?

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
...
+                mt_steps = [
+                    ("MT", (source_lang, target_lang)),
+                    ("MT", (target_lang, source_lang))
+                ]
+            else:
+                mt_steps = [("MT", (source_lang, target_lang))]
+            steps = mt_steps
+        loss_prefixes = set()
+
+        random.shuffle(steps)
```
fhieber, line 319: so steps contains roughly how many items? Just a few, right? Are we shuffling this list again or is there a good reason to keep the permutation generated once?

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
...
+                elif step_variant == (target_lang, source_lang):
+                    did_grad_step = self._mt_step(batch=train_iter.next().reverse(), loss_prefix=loss_prefix)
+                else:
+                    assert False
+
+                # TODO: how to define epochs with the presence of the mono iters? For now we'll just keep epochs = parllel data epochs...
+                if not train_iter.iter_next():
+                    self.state.epoch += 1
+                    train_iter.reset()
+            elif step == "MASS":
```
fhieber, line 371: these strings should be constants I suppose

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
...
+                step_lang = step_variant
+                # TODO: support for monolingual data without a language!?
+                assert step_lang.name is not None
+                loss_prefix = f"MASS-{step_lang.name}"
+                mono_iter = mono_iters[step_lang]
+                did_grad_step = self._mass_step(batch=mono_iter.next(),
+                                                # TODO: remove the vocab (was just for print debugging...)
```
fhieber, line 378: TODO: remove the todo? :)

sockeye/training.py
```diff
@@ -228,40 +275,141 @@ def fit(self,
...
+                # bubble up/expose parameters
+                sampled_bt = False
+                tagged_bt = False
+
+                step_lang, bt_src_lang, bt_trg_lang = step_variant
+                # TODO: support for monolingual data without a language!?
+                assert step_lang.name is not None
+                loss_prefix = f"BT-{step_lang.name}"
+                # TODO: how to get mono dat here!?
+                mono_iter = mono_iters[step_lang]._iter
```
fhieber, line 393: why is this line different than for MASS above? (376)

sockeye/training.py
```diff
@@ -283,33 +431,55 @@ def fit(self,
...
         self.state.diverged = self._determine_divergence(val_metrics)
         self._adjust_learning_rate(has_improved)
         if has_improved:
             self._update_best_params()
             self._save_trainer_states(self.best_optimizer_states_fname)
-        self._write_and_log_metrics(train_metrics=train_metrics, val_metrics=val_metrics)
+        self._write_and_log_metrics(train_metrics=train_metrics, val_metrics=all_prefixed_val_metrics)
         for metric in train_metrics:
             metric.reset()
+        # TODO: add the monolingual iterators here as well!
```
fhieber, line 482: +1

sockeye/training.py
```diff
@@ -341,39 +511,187 @@ def _forward_backward(self, batch: data_io.Batch):
...
+            tagged = mx.nd.full(shape=(source.shape[0], 1, 1), val=C.BOS_ID, dtype=source.dtype)
+            source = mx.nd.concat(tagged, source, dim=1)
+
+        label = batch.data.squeeze().astype(np.int32)
+        target = label[:, :-1]
+        target = mx.nd.where(target == C.EOS_ID, mx.nd.zeros_like(target), target)  # replace other <eos>'s with <pad>
+        # target: (batch_size, seq_len, 1)
+        target = mx.nd.concat(mx.nd.full((target.shape[0], 1), val=C.BOS_ID, dtype=np.int32), target)
+
+        # TODO: full target factor support
```
fhieber, line 636: For my understanding: for bi-directional models that use both source and target factors and in case of backtranslation, the gold target factors should act as source factors for the backtranslation, correct?

sockeye/training.py
```diff
@@ -341,39 +511,187 @@ def _forward_backward(self, batch: data_io.Batch):
...
+        did_grad_step = self.gradient_step()
+
+        self.state.samples += batch.samples
+        for loss_func, (loss_value, num_samples) in zip(self.loss_functions, loss_outputs):
+            loss_func.prefixed_metric(loss_prefix).update(loss_value.asscalar(), num_samples.asscalar())
+        self._speedometer(self.state.epoch, self.state.batches,
+                          self.state.updates, batch.samples, batch.tokens, (metric for lf in self.loss_functions for metric in lf.metrics))
+        return did_grad_step
+
+    def _bt_step(self, batch: data_io.MonoBatch, translator: inference.Translator, bt_src_lang, bt_trg_lang, loss_prefix, tagged_bt):
```
fhieber, line 562: I think it would be helpful to document what all this step does

fhieber (commented):
Great addition, thanks!
Started from the bottom of the changed files, just first set of comments / questions.

fhieber:
I think we should consider converging to a single code path for data preparation to reduce code duplication related to creation of vocabularies, data iterators, statistics etc. How about folding data preparation (binarization and storage as numpy arrays on disk) into the `sockeye.train` process. That is, we always 'prepare' data and serialize it to disk (maybe as a subfolder in the model folder with an option to clean up afterwards). This would allow us to simplify quite a bit of logic in the `data_io` and `train` modules.

sockeye/arguments.py
```diff
@@ -640,6 +707,34 @@ def add_vocab_args(params):
                         help='Pad vocabulary to a multiple of this integer. Default: %(default)s.')
 
 
+def add_mono_vocab_args(params):
+    params.add_argument('--mono-vocab',
+                        required=False,
+                        default=None,
+                        help='Existing vocabulary (JSON).')
+    params.add_argument('--mono-factor-vocabs',
```
barbaradarques, line 715: What about we merge this with `--mono-factors-use-mono-vocab`? If a path is specified, use it; if "default" (or something to this effect), then the data`s vocab is used.

tdomhan:
I agree that there's quite some potential for simplification when it comes to data io and binarization. I'm not sure I understand the proposal though. Are you suggesting to remove data preparation and always perform data preparation during training?
To be honest I do like the fact that the trainer gets the data iterators which were either binarized before or on the fly. Also, I wouldn't remove the functionality to binarize separately from training in order to binarize once and then potentially train multiple models. Not sure if that is what you are suggesting though.

fhieber:
The proposal is to keep data preparation as is, but replace the code path for `sockeye.train` when 'raw' / non-binarized data is given with code that 1) binarizes the data (by calling prepare_data) and then 2) passing on that folder to further sockeye.train logic.
What we would get rid of, is the on-the-fly iterator creation from raw data without writing to disk, and replace it with a pipeline approach of first binarizing the data & writing to disk, followed by loading it again, all in the same single process.
`sockeye.prepare_data` would remain untouched.

The downsides of this proposed approach (which is also used by other toolkits afaik) is that 
- we have additional disk usage for every training (as we always need to store binarized data. We can create an option to delete afterwards)
- initial time until actual training starts might be a bit higher.

We can think of where to store the on-the-fly binarized data should be stored:
- inside the model directory (`<model output folder>/data`), or
- outside the model directory with a predefined, related name? i.e. `<model output folder>.data`

sockeye/training.py
tdomhan: It is because we do the masking inside the iterator and the `_iter` just gives the raw mono data without masking. In any case this is not nice and I'd like to refactor this. The reason I do the masking in the iterator is to be able to use threading there. I wanted to reevaluate whether this was actually necessary and if not pull out masking from data_io.

sockeye/training.py
tdomhan: Reverse takes a source -> target batch and reverses it to target -> source. This is for bi-directional model training.

sockeye/training.py
tdomhan: Yes, only a few items. It is shuffled only once. Would you prefer to shuffle it regularly?

fhieber: I have no good intuition about this, but I would probably shuffle it regularly to avoid any kind of consistent bias in the training algorithm towards grouping/not grouping monolingual and parallel steps if you train for a long time. Not sure, just speculating.

PR #923, opened 2020-12-22
Move to dataclasses & update version of pylint & mypiy

fhieber:
Python 3.7+ adds Dataclasses, which are convenient syntactic sugar to simplify `Config` classes: https://docs.python.org/3/library/dataclasses.html

There is also a Python 3.6 backport module that we now depend on (until we remove support for Python 3.6).

Also updated versions of pylint and mypy.

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/checkpoint_decoder.py
+++ b/sockeye/checkpoint_decoder.py
@@ -1,4 +1,4 @@
-# Copyright 2017--2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2017--2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not
 # use this file except in compliance with the License. A copy of the License
--- a/sockeye/checkpoint_decoder.py
+++ b/sockeye/checkpoint_decoder.py
@@ -122,7 +122,7 @@ class CheckpointDecoder:
         for factor_idx, factor in enumerate(self.targets_sentences):
             write_to_file(factor, os.path.join(model_folder, C.DECODE_REF_NAME.format(factor=factor_idx)))
 
-        self.inputs_sentences = list(zip(*self.inputs_sentences))  # type: List[List[str]]
+        self.inputs_sentences = list(zip(*self.inputs_sentences))  # type: ignore
 
         scorer = inference.CandidateScorer(
             length_penalty_alpha=length_penalty_alpha,
@@ -177,7 +177,7 @@ class CheckpointDecoder:
                     if output_file is not None:
                         print(output_string, file=output_file)
         avg_time = trans_wall_time / len(self.targets_sentences[0])
-        translations = list(zip(*translations))
+        translations = list(zip(*translations))  # type: ignore
 
         # 2. Evaluate
         metrics = {C.BLEU: evaluate.raw_corpus_bleu(hypotheses=translations[0],
--- a/sockeye/config.py
+++ b/sockeye/config.py
@@ -1,4 +1,4 @@
-# Copyright 2017--2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2017--2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not
 # use this file except in compliance with the License. A copy of the License
@@ -12,8 +12,8 @@
 # permissions and limitations under the License.
 
 import copy
-import inspect
 import logging
+from dataclasses import dataclass
 
 import yaml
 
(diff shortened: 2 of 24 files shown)


sockeye/checkpoint_decoder.py
```diff
@@ -122,7 +122,7 @@ def __init__(self,
         for factor_idx, factor in enumerate(self.targets_sentences):
             write_to_file(factor, os.path.join(model_folder, C.DECODE_REF_NAME.format(factor=factor_idx)))
 
-        self.inputs_sentences = list(zip(*self.inputs_sentences))  # type: List[List[str]]
+        self.inputs_sentences = list(zip(*self.inputs_sentences))  # type: ignore
```
tdomhan, line 125: Just out of curiosity: Why is it necessary to ignore the type?

sockeye/data_io.py
```diff
@@ -404,7 +405,7 @@ def shard_data(source_fnames: List[str],
                buckets: List[Tuple[int, int]],
                length_ratio_mean: float,
                length_ratio_std: float,
-               output_prefix: str) -> Tuple[List[Tuple[List[str], List[str], 'DataStatistics']], 'DataStatistics']:
+               output_prefix: str):
```
tdomhan, line 408: Why remove the return type?

sockeye/data_io.py
```diff
@@ -1061,54 +1062,31 @@ def get_scoring_data_iters(sources: List[str],
...
-        super().__init__()
-        self.num_sents = num_sents
-        self.length_ratio_mean = length_ratio_mean
-        self.length_ratio_std = length_ratio_std
+    num_sents: int
+    length_ratio_mean: float
+    length_ratio_std: float
 
 
+@dataclass
```
tdomhan, line 1072: Looks like a nice simplification of code :)

sockeye/checkpoint_decoder.py
fhieber: It's one of the very old mypy issues that we were always unable to fix.

sockeye/data_io.py
fhieber: same as above.

sockeye/config.py
```diff
@@ -29,44 +29,14 @@ def __init__(cls, name, bases, kwds):
...
-            raise AttributeError("Cannot set self as attribute")
-        object.__setattr__(self, key, value)
-
-    def __setstate__(self, state):
-        """Pickle protocol implementation."""
-        # We first take the serialized state:
-        self.__dict__.update(state)
-        # Then we take the constructors default values for missing arguments in order to stay backwards compatible
-        # This way we can add parameters to Config objects and still load old models.
-        init_signature = inspect.signature(self.__init__)
```
tdomhan, line 50: I see from the test `test_config_missing_attributes_filled_with_default` that this is now already supported out of the box, neat!

sockeye/checkpoint_decoder.py
```diff
@@ -177,7 +177,7 @@ def decode_and_evaluate(self, output_name: Optional[str] = None) -> Dict[str, fl
                     if output_file is not None:
                         print(output_string, file=output_file)
         avg_time = trans_wall_time / len(self.targets_sentences[0])
-        translations = list(zip(*translations))
+        translations = list(zip(*translations))  # type: ignore
```
tdomhan, line 180: nit: the copyright date of the file could also be updated.

sockeye/transformer.py
```diff
@@ -102,7 +85,7 @@ def __init__(self,
                                                    prefix="ff_post_",
                                                    num_hidden=config.model_size)
             self.lhuc = None
-            if config.use_lhuc:
+            if config.lhuc:
```
tdomhan, line 88: Renaming this wouldn't be backwards compatible, right? I would leave it at `use_lhuc`.

test/unit/test_config.py
```diff
@@ -12,29 +12,28 @@
...
-class ConfigTest(config.Config):
-    yaml_tag = "!ConfigTest"
-
-    def __init__(self, param, config=None):
-        super().__init__()
-        self.param = param
-        self.config = config
+@dataclass
+class ConfigTest(Config):
+    #yaml_tag = "!ConfigTest"
```
tdomhan, line 25: nit: commented out code.

test/unit/test_config.py
```diff
@@ -93,18 +85,17 @@ def test_config_copy():
...
+    new_attribute: str = "new_attribute"
 
 
 def test_config_missing_attributes_filled_with_default():
     # when we load a configuration object that does not contain all attributes as the current version of the
     # configuration object we expect the missing attributes to be filled with the default values taken from the
     # __init__ method.
 
-    config_obj = config.Config.load("test/data/config_with_missing_attributes.yaml")
+    config_obj = Config.load("test/data/config_with_missing_attributes.yaml")
```
tdomhan, line 99: The comment above isn't correct anymore (refers to the `__init__` method).

test/unit/test_config.py
```diff
@@ -51,13 +50,6 @@ def test_config_eq():
     assert c_nested != c_nested_c2
 
 
-def test_config_no_self_attribute():
```
tdomhan, line 54: What was the reason we had this constraint/test?

sockeye/checkpoint_decoder.py
tdomhan: ok. Yeah, for the CheckpointDecoder in particular I always get confused on whether we have a list of sentences with each having a list of factors or a list of factors which each have a list of sentences (the types don't help for that though).

sockeye/config.py
fhieber: yes, pretty cool!

sockeye/checkpoint_decoder.py
fhieber: sure thing

sockeye/transformer.py
fhieber: good catch, will revert

test/unit/test_config.py
fhieber: removed

test/unit/test_config.py
fhieber: changed

test/unit/test_config.py
fhieber: I think we wanted to avoid nested/circular config dependencies that refer to itself, because the str/repr methods would create infinite strings. I don't think we need that since it never really happened that we accidentally made a config a child of one of its child configs.

tdomhan (approved):

Issue #890, opened 2020-10-05
Continuing training from a partially trained model sometimes reverts learning-rate to previous value


tuglat:
Experienced in both 2.1.16 and 2.1.21.

Training a model using --max-checkpoints to periodically stop and then continue training. When continuing training, the learn-rate seems to have been set back to the original. I have included the log files for two sequential training runs that exhibit this behavior.

This does not always happen, but I don't know how to characterize the different cases.

[log.00.txt](https://github.com/awslabs/sockeye/files/5329983/log.00.txt)
[log.01.txt](https://github.com/awslabs/sockeye/files/5329984/log.01.txt)

tdomhan:
Thanks for reporting! Could you also share the full commands used for initially starting training and then to continue training?

tuglat:
Whenever doing continued training w/ Sockeye 2 (last tested on 2.8.2), the LearningRateSchedulerPlateauReduce that I specify does not seem to be recovered from the saved state. Instead it appears that a new one is created. The learning rate is then set to the MXNet optimizer base learning rate, which **is** recovered, and set to the initial learning rate specified in the parameters.

Could this be related to the message 

      learning rate from ``lr_scheduler`` has been overwritten by ``learning_rate`` in optimizer.

that appears at the bottom of the the stdout from running sockeye-train? This message comes from mxnet/optimizer/optimizer.py.

huache:
I encounter the same problem. Every time I continue the training, the learning rate will restart from 2e-4 ，and it will drastically reduce the performance of the model. I update the sockeye version from 2.1.13 to 2.1.16 , then 2.1.26 .  The problem still exists.

![image](https://user-images.githubusercontent.com/11642845/103450805-506e3e00-4cf6-11eb-8cd8-5a46fd09636f.png)

tuglat:
This has to do w/ the fact that the optimizer is not saved in _save_trainer_states in training.py. The lr_scheduler is held inside the optimizer.

I am working on a fix that saves the lr_scheduler separately.

I closed the issue because it was marked as a question, and getting no attention. I was going to open it up a new issue, and submit my fix as a PR. Got sidetracked by family.

Steve

huache:
Thank you so much !  Looking forward to your fix PR .

tuglat:
@huache PR has been accepted and merged.

tuglat:
fixed in PR's #920, #924, and #926

HossamAmer12:
[...]

PR #924, opened 2021-01-05
save and restore lr_scheduler as needed

tuglat:
This change fixes the problem in issue #890, where the lr_scheduler does not behave as expected when continuing training. The problem is that the lr_scheduler is kept as part of the optimizer, but the optimizer is not saved when saving state. Therefore, every time training is restarted, a new lr_scheduler is created with initial parameter settings.

- [ x] Changes are complete (if posting work-in-progress code, prefix your pull request title with '[WIP]'
until you can check this box.

        --- There are failures, but these are the same failures exhibited by the master branch
- [no] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?
         --- There are failures, but these are the same failures exhibited by the master branch
         --- There are failures, but these are the same failures exhibited by the master branch
- [no] You have considered writing a test
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/lr_scheduler.py
+++ b/sockeye/lr_scheduler.py
@@ -176,7 +176,13 @@ class LearningRateSchedulerPlateauReduce(AdaptiveLearningRateScheduler):
         return lr
 
     def __repr__(self):
-        return "LearningRateSchedulerPlateauReduce(reduce_factor=%.2f, reduce_num_not_improved=%d, num_not_improved=%d, base_lr=%s, lr=%s, warmup=%d, warmed_up=%s)" % (self.reduce_factor, self.reduce_num_not_improved, self.num_not_improved, self.base_lr, self.lr, self.warmup, self.warmed_up)
+        return (
+            "LearningRateSchedulerPlateauReduce(reduce_factor=%.2f, reduce_num_not_improved=%d, num_not_improved=%d,"
+            " base_lr=%s, lr=%s, warmup=%d, warmed_up=%s)"
+            %
+            (self.reduce_factor, self.reduce_num_not_improved,
+             self.num_not_improved, self.base_lr, self.lr, self.warmup, self.warmed_up)
+        )
 
 
 def get_lr_scheduler(scheduler_type: str,
--- a/sockeye/lr_scheduler.py
+++ b/sockeye/lr_scheduler.py
@@ -138,6 +138,7 @@ class LearningRateSchedulerPlateauReduce(AdaptiveLearningRateScheduler):
         self.lr = None  # type: Optional[float]
         self.t_last_log = -1
         self.warmed_up = not self.warmup > 0
+        
         logger.info("Will reduce the learning rate by a factor of %.2f whenever"
                     " the validation score doesn't improve %d times.",
                     reduce_factor, reduce_num_not_improved)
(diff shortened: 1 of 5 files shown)


sockeye/lr_scheduler.py
```diff
@@ -175,8 +176,7 @@ def __call__(self, t):
         return lr
 
     def __repr__(self):
-        return "LearningRateSchedulerPlateauReduce(reduce_factor=%.2f, " \
-               "reduce_num_not_improved=%d)" % (self.reduce_factor, self.reduce_num_not_improved)
+        return "LearningRateSchedulerPlateauReduce(reduce_factor=%.2f, reduce_num_not_improved=%d, num_not_improved=%d, base_lr=%s, lr=%s, warmup=%d, warmed_up=%s)" % (self.reduce_factor, self.reduce_num_not_improved, self.num_not_improved, self.base_lr, self.lr, self.warmup, self.warmed_up)
```
fhieber, line 179: Can you limit line length to 120 here?

sockeye/training.py
```diff
@@ -587,6 +588,18 @@ def _load_trainer_states(self, fname):
         self.trainer.load_states(fname)
         logger.info('Loaded optimizer states from "%s"', fname)
 
+    def _save_lr_scheduler(self, fname):
+        with open(fname, "wb") as fp:
+            pickle.dump(self.trainer.optimizer.lr_scheduler, fp)
+        logger.info('Saved lr_scheduler to "%s"', fname)
```
fhieber, line 594: can we combine / simplify to one line of logging instead?
```suggestion
        logger.info('Saved '%s' to "%s", self.trainer.optimizer.lr_scheduler, fname)
```

sockeye/training.py
```diff
@@ -587,6 +588,18 @@ def _load_trainer_states(self, fname):
...
+        logger.info('Saved lr_scheduler %s', self.trainer.optimizer.lr_scheduler.__repr__())
+
+    def _load_lr_scheduler(self, fname):
+        with open(fname, "rb") as fp:
+            self.trainer.optimizer.lr_scheduler = pickle.load(fp)
+        logger.info('Loaded lr_scheduler from "%s"', fname)
```
fhieber, line 600: same here

sockeye/training.py
```diff
@@ -587,6 +588,18 @@ def _load_trainer_states(self, fname):
         self.trainer.load_states(fname)
         logger.info('Loaded optimizer states from "%s"', fname)
 
+    def _save_lr_scheduler(self, fname):
+        with open(fname, "wb") as fp:
+            pickle.dump(self.trainer.optimizer.lr_scheduler, fp)
+        logger.info('Saved lr_scheduler to "%s"', fname)
+        logger.info('Saved lr_scheduler %s', self.trainer.optimizer.lr_scheduler.__repr__())
+
+    def _load_lr_scheduler(self, fname):
```
fhieber, line 597: can we make this method backwards compatible? Right now, it would be failing when trying to continue training an existing (2.3.5 and lower) model. You could check in the method for existence of the file under `fname` and only load if it exists.

fhieber (approved):
Thanks for fixing this!

fhieber (changes requested):
Sorry meant to click 'Request Changes'

sockeye/lr_scheduler.py
tuglat: ok

sockeye/training.py
tuglat: ok

sockeye/training.py
tuglat: ok

sockeye/training.py
tuglat: ok

fhieber (approved):
LGTM, thanks!

ketranm:
LGTM!

sockeye/training.py
```diff
@@ -587,6 +588,17 @@ def _load_trainer_states(self, fname):
         self.trainer.load_states(fname)
         logger.info('Loaded optimizer states from "%s"', fname)
 
+    def _save_lr_scheduler(self, fname):
+        with open(fname, "wb") as fp:
```
tdomhan, line 592: The scheduler can also be `None`. Could you add a check for that?

tdomhan (changes requested):

sockeye/training.py
tuglat: ok

tdomhan (approved):
Excellent, thanks for the contribution!

Issue #928, opened 2021-01-11
Bug: During training, if  --keep-last-params is too small, construction of model averages can be corrupted


tuglat:
if  --keep-last-params is set for 4, for example, and we want to construct an average of the 4 best parameters sets, it is very likely that the true 4 best parameters sets are not all saved in those kept around by --keep-last-params.

fhieber:
Hi @tuglat,

Thanks for bringing this up. I am not yet convinced that the above issue is really a bug, and not more a configuration issue by the user. I agree that the interaction between parameter averaging and `--keep-last-params` is a bit obscure and maybe it is worth adding a warning to parameter averaging if the set to select parameters from is very small.

tuglat:
One of the reasons for only keeping N of the past parameters sets is to save storage space. Keeping all of the parameters sets as we go can eat up quite a bit of disk space. Instead of a bug, then, this could be considered a performance improvement, where we can ensure that we only keep around as many of the N-best that we actually need.

tuglat:
fixed w/ pull request #929

PR #929, opened 2021-01-11
add facility to cache 'best' parameter sets to a sub-directory

tuglat:
(description of the change)

Addresses issue #928  

Modified code so that the N-best parameter sets are cached into a sub-directory, where they will be available to construct average models, even if these parameter sets would not be saved depending upon the value set for '--keep-last-params`.

until you can check this box.
- [no] Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?
  Modified test/unit/test_arguments.py::test_training_arg test to account for new optional arguments
- [x] Updated CHANGELOG.md


mjdenkowski:
Thanks for working on this!  Keeping only the N best checkpoints is a great way to reduce disk space usage during training.

Instead of copying params files to a separate directory, how about extending the `cleanup_params_files` function?
https://github.com/awslabs/sockeye/blob/b6d8d356b3d2b626a12cde767b2d995fadbb9751/sockeye/utils.py#L691

You could apply the same point-finding logic to make sure the current N best checkpoints are preserved each time `cleanup_params_files` is called.

tuglat:
I'll take a look at modifying cleanup_params_files as you suggest. The separate directory thing does make it easier to satisfy both the constraints imposed by `--keep-last-params` and `--cache-last-best-params`.

tuglat:
There is a conflict in CHANGELOG.md, probably due to changes that were committed while have been slowly working on this branch. __init__.py probably needs to be fixed as well. What is the procedure here? Should I make changes to this branch to make consistent w/ head, or is this done by reviewers?

fhieber:
Thanks, feel free to make changes to the branches to resolve the conflicts.

CHANGELOG.md
```diff
@@ -11,12 +11,21 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.3.9]
+
+### Fixed
+
+- Make sure that the top N best params files are cached in a separate
```
fhieber, line 18: This description seems no longer accurate (no more separate directory)

sockeye/arguments.py
```diff
@@ -1085,6 +1085,22 @@ def add_training_args(params):
...
 
+    
+    train_params.add_argument('--cache-last-best-params',
+                              type=int,
+                              default=4,
+                              help='Cache the last n best params files to a separate directory, use 0 or negative to disable. Default: %(default)s')
+
+    train_params.add_argument('--cache-strategy',
+                              type=str,
+                              default="best",
```
fhieber, line 1096: Can we use the string constants defined here instead: https://github.com/awslabs/sockeye/blob/master/sockeye/constants.py#L431

sockeye/arguments.py
```diff
@@ -1085,6 +1085,22 @@ def add_training_args(params):
                               action="store_true",
                               help='In addition to keeping the last n params files, also keep params from checkpoint 0.')
 
+    
+    train_params.add_argument('--cache-last-best-params',
+                              type=int,
+                              default=4,
```
fhieber, line 1091: What would this default need to be set to, to not change the default behavior of the training CLI as it is right now?

fhieber: I am probably missing something, but why do we need this argument if we already have `--keep-last-params'? And why use 'cache' instead of 'keep'?

sockeye/utils.py
```diff
@@ -688,7 +689,8 @@ def metric_value_is_better(new: float, old: float, metric: str) -> bool:
         return new < old
 
 
-def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int, best_checkpoint: int, keep_first: bool):
+def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int, best_checkpoint: int, keep_first: bool,
+                         max_params_files_to_cache: int, cache_metric: str, cache_strategy: str):
```
fhieber, line 693: can these arguments be made optional? If no strategy is defined for example we would not go into below if block and call the 'average' logic.

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+
+    #
+    # make sure we keep N best params files from .metrics file according to strategy.
+    #
+
+    top_n = []
+    metrics_path = os.path.join(output_folder, C.METRICS_NAME)
+    
+    if max_params_files_to_cache > 0 and os.path.exists(metrics_path):
+        metric = cache_metric
```
fhieber, line 714: why renaming the variable here?

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+        strategy = cache_strategy
```
fhieber, line 715: and here?

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+            
+        maximize = C.METRIC_MAXIMIZE[metric]
+        points = get_validation_metric_points(model_path=output_folder, metric=metric)
+
+        if strategy == "best":
```
fhieber, line 720: Can we use the string constants instead?

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+            # N best scoring points
+            top_n = average._strategy_best(points, max_params_files_to_cache, maximize)
```
fhieber, line 722: Could you refactor these methods to be non-private (i.e. rename to remove the '_')?

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
     """
     if max_to_keep <= 0:
         return
+
+    #
+    # make sure we keep N best params files from .metrics file according to strategy.
+    #
+
+    top_n = []
```
fhieber, line 710: could you add a type annotation?

sockeye/utils.py
```diff
@@ -700,12 +702,49 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+            top_n = average._strategy_last(points, max_params_files_to_cache, maximize)
+
+        elif strategy == "lifespan":
+            # Track lifespan of every "new best" point
+            # Points dominated by a previous better point have lifespan 0
+            top_n = average._strategy_lifespan(points, max_params_files_to_cache, maximize)
+        else:
+            raise RuntimeError("Unknown strategy, options: [best, last, lifespan]")
+
+    top_n = [x[1] for x in top_n]
```
fhieber, line 735: could this be made a set for faster membership checking?

fhieber (changes requested):

sockeye/arguments.py
tuglat: The default value would be anything <= the value of `--keep-last-params`. 

The reason for adding the arg `--cache-last-best-params` in addition to `--keep-last-params' is because the semantics are different. `--keep-last-params'  does just what it says, makes sure that we keep only the last N params files we saved. However, these last N params files in time are not necessarily the last N **best** params files, and very often are not. If we don't make the effort to keep the N **best** params files, then if we want to build an averaged params file, we will not be able to use the true N best. 

Now, we could just set `--keep-last-params' to zero, and keep every params file we save off. But this can eat an awful lot of disk space. This change makes sure we keep the information we need to build an accurate averaged parameter set, and to save disk space while so doing.

There is no great reason for using 'cache' instead of 'keep'. It seemed more evocative of what this is trying to do, and to distinguish of `--keep-last-params`.

sockeye/utils.py
tuglat: These could be made optional.

sockeye/utils.py
tuglat: No good reason. Will address.

sockeye/utils.py
tuglat: No good reason. Will address.

sockeye/utils.py
tuglat: yes. Will address.

sockeye/utils.py
tuglat: sure. Will address.

sockeye/utils.py
tuglat: Sure. Will address.

CHANGELOG.md
tuglat: fixed

sockeye/arguments.py
tuglat: yes, will address

sockeye/utils.py
tuglat: sure. Will address

sockeye/arguments.py
fhieber: Got you, thanks for the explanation! Let's set the default value to match previous behavior. How about adding a small note in the help string of this new argument that in the 'worst case' this would keep up to x additional param files on top of the y files from `--keep-last-params`.

fhieber: Just saw your recent commit that you already added a modified help string, thanks!

sockeye/constants.py
```diff
@@ -341,7 +341,7 @@
 AVG_TIME = "avg-sec-per-sent"
 DECODING_TIME = "decode-walltime"
 METRICS = [PERPLEXITY, ACCURACY, LENRATIO_MSE, BLEU, CHRF, ROUGE1]
-METRIC_MAXIMIZE = {ACCURACY: True, BLEU: True, CHRF: True, ROUGE1: True, PERPLEXITY: False}
+METRIC_MAXIMIZE = {ACCURACY: True, BLEU: True, CHRF: True, ROUGE1: True, PERPLEXITY: False, LENRATIO_MSE: False}
```
fhieber, line 344: thanks!

sockeye/utils.py
```diff
@@ -700,12 +702,46 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
     """
     if max_to_keep <= 0:
         return
+
+    #
+    # make sure we keep N best params files from .metrics file according to strategy.
+    #
+
+    top_n : Set[int] = set()
```
fhieber, line 710: I didn't know this syntax worked, thanks!

sockeye/utils.py
```diff
@@ -700,12 +702,46 @@ def cleanup_params_files(output_folder: str, max_to_keep: int, checkpoint: int,
...
+            top = average.strategy_last(points, max_params_files_to_cache, maximize)
+
+        elif cache_strategy == C.AVERAGE_LIFESPAN:
+            # Track lifespan of every "new best" point
+            # Points dominated by a previous better point have lifespan 0
+            top = average.strategy_lifespan(points, max_params_files_to_cache, maximize)
+        else:
+            raise RuntimeError("Unknown strategy, options are: %s" % (C.AVERAGE_CHOICES))
+
+        top_n = set([x[1] for x in top])
```
fhieber, line 732: I think you can get rid of the `[]`s and do `top_n = set(x[1] for x in top)`

fhieber (approved):
LGTM, the failing tests seem to be unrelated (mypy).

fhieber:
Thank you for the contribution!

PR #936, opened 2021-03-09
Adding an option to avoid generate <unk> during decoding

ketranm:
Avoid generate `<unk>` during beam search by reset the probablity of generating `<unk>` token to 0. This change has minmal to zero effect to the search space.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/beam_search.py
```diff
@@ -677,7 +679,12 @@ def forward(self,
             target_dists, model_states, target_factors = self._inference.decode_step(best_word_indices,
                                                                                      model_states,
                                                                                      vocab_slice_ids)
-
+            # Do not generate <unk> token if prevent_unk is specified
+            if self.prevent_unk:
+                if target_dists.ndim == 1:
+                    target_dists[C.UNK_ID] = float('inf')
```
fhieber, line 685: in what cases would `ndim` be 1?

sockeye/beam_search.py
```diff
@@ -677,7 +679,12 @@ def forward(self,
             target_dists, model_states, target_factors = self._inference.decode_step(best_word_indices,
                                                                                      model_states,
                                                                                      vocab_slice_ids)
-
+            # Do not generate <unk> token if prevent_unk is specified
```
fhieber, line 682: can we move this block into the `_Inference`  classes maybe? It would be good to keep the core beam search loop clean and high-level and move logic into sub-blocks. One reason is that it may make it easier in the future to convert these blocks into compiled computation graphs as much as possible.

sockeye/beam_search.py
ketranm: when `beam_size =1` and `batch_size = 1`. The [unit test for beam_search ](https://github.com/awslabs/sockeye/blob/master/test/unit/test_beam_search.py#L307)use `beam_size=1` and `batch_size =1` so I add this case to make sure it passes the test.

sockeye/beam_search.py
ketranm: updated the code. I move the block into  `__Inference`

sockeye/beam_search.py
ketranm: tested with `float('inf')` and `np.inf`, and also with using an `adding_mask` on GPU. Updated the code with adding mask since it's faster.

sockeye/beam_search.py
```diff
@@ -178,7 +178,10 @@ def __init__(self, **kwargs):
     def hybrid_forward(self, F,
                        target_dists, finished, inactive,
                        scores_accumulated, lengths, max_lengths,
-                       pad_dist, eos_dist):
+                       unk_dist, pad_dist, eos_dist):
+        # make sure to avoid generating <unk> if unk_dist is specified
+        if unk_dist is not None:
```
fhieber, line 183: nice, thanks!

sockeye/beam_search.py
```diff
@@ -604,6 +609,11 @@ def forward(self,
         eos_dist = mx.nd.full((batch_size * self.beam_size, self.output_vocab_size), val=np.inf,
                               ctx=self.context, dtype=self.dtype)
         eos_dist[:, C.EOS_ID] = 0
+        unk_dist = None
+        if self.prevent_unk:
+            unk_dist = mx.nd.full((batch_size * self.beam_size, self.output_vocab_size), val=0,
```
fhieber, line 614: you could simplify `mx.nd.zeros_like(eos_dist)`, context is already taken from the argument of `zeros_like()`

sockeye/beam_search.py
```diff
@@ -652,6 +662,10 @@ def forward(self,
             eos_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids_shape),
                                   val=np.inf, ctx=self.context, dtype=self.dtype)
             eos_dist[:, C.EOS_ID] = 0
+            if unk_dist is not None:
+                unk_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids_shape),
```
fhieber, line 666: same here

test/unit/test_beam_search.py
```diff
@@ -227,6 +227,8 @@ def test_update_scores():
     pad_dist = mx.nd.full((batch_beam_size, vocab_size - 1), val=np.inf, dtype='float32')
     eos_dist = mx.nd.full((batch_beam_size, vocab_size), val=np.inf, dtype='float32')
     eos_dist[:, C.EOS_ID] = 0
+    unk_dist = mx.nd.full((batch_beam_size, vocab_size), val=0, dtype='float32')
+    unk_dist[:, C.UNK_ID] = np.inf
```
fhieber, line 231: could you add a test to check whether the output of UpdateScores is as expected when prevent_unk is enabled? I think this default test should use `unk_dist = None`.

fhieber (commented):
Thanks for iterating!
Would it make sense to have at least 1 integration or system test use this new flag?

sockeye/beam_search.py
ketranm: `mx_nd.zeros_like` seems cleaner but it causes pylint error 
```
E1137: 'unk_dist' does not support item assignment (unsupported-assignment-operation)
```
because of the assignment `unk_dist[:, C.UNK_ID] = np.inf`. Is there any solution for this?

fhieber: Pylint isn't always right. You can try to add a type annotation in line 612: `unk_dist = None. # type: Optional[mx.nd.NDArray]` and see if that fixes it. Otherwise you could ignore the specific pylint error in that line.

ketranm: add `unk_dist = None # type: Optional[mx.nd.NDArray]` but it doesn't help. Ignoring pylint error for now

test/unit/test_beam_search.py
ketranm: added a new test.

sockeye/beam_search.py
ketranm: it seems that I have to manually add `pylint --rcfile=pylintrc test -E --disable= E1137` to the code

fhieber: you add a comment to line 614: `#. pylint: ignore: E1137` or something like this.

fhieber (approved):
LGTM

sockeye/beam_search.py
```diff
@@ -604,6 +609,10 @@ def forward(self,
         eos_dist = mx.nd.full((batch_size * self.beam_size, self.output_vocab_size), val=np.inf,
                               ctx=self.context, dtype=self.dtype)
         eos_dist[:, C.EOS_ID] = 0
+        unk_dist = None
+        if self.prevent_unk:
+            unk_dist = mx.nd.zeros_like(eos_dist)
```
tdomhan, line 614: Should we not use the `self.context` here?

tdomhan (changes requested):

sockeye/beam_search.py
```diff
@@ -652,6 +661,9 @@ def forward(self,
             eos_dist = mx.nd.full((batch_size * self.beam_size, vocab_slice_ids_shape),
                                   val=np.inf, ctx=self.context, dtype=self.dtype)
             eos_dist[:, C.EOS_ID] = 0
+            if unk_dist is not None:
+                unk_dist = mx.nd.zeros_like(eos_dist)
```
tdomhan, line 665: same here.

sockeye/beam_search.py
fhieber: storage type, context, and data type are taken from the input array of `zeros_like` so `unk_dist` should be on `self.context`.

fhieber:
```python
a = mx.nd.ones((1,), ctx=mx.cpu(1))
a2 = mx.nd.ones((1,), ctx=mx.cpu(0))
mx.nd.zeros_like(a).context  # cpu(1)
mx.nd.zeros_like(a2).context  # cpu(0)
```

tdomhan: ohh ok, excellent!

tdomhan (approved):

PR #945, opened 2021-03-21
Update to MXNet 1.8 & Python 3.7

fhieber:
updates to mxnet 1.8 and Python 3.7.

MXNet 1.8 in pypi no longer publishes wheels for mac Python 3.6, we therefore bump our minimum python version from 3.6 to 3.7.

until you can check this box.
- [x] Updated CHANGELOG.md


tdomhan (approved):

tdomhan:
Huh, that's actually surprising that Python 3.6 was removed. Is there a way to allow a different set of Python versions for Linux? EOL for Python 3.6 is end of this year and I believe it is still widely used. So if possible I think we should keep 3.6 support till then.

tdomhan:
I guess the publishing version isn't important necessarily, but it currently uses `3.6`: https://github.com/awslabs/sockeye/blob/master/.github/workflows/python-publish.yml#L20

fhieber:
We depend on mxnet and if mxnet decides to drop 3.6 support we are kind of forced to go along.
Good catch, we should update the publish workflow as well.

tdomhan:
yeah, absolutely. I meant that MXNet dropping 3.6 is something Mac OS specific so that we could still support Linux + 3.6.

fhieber:
We need to find out then how to encode that in the Github actions file (not sure how to do this).

PR #948, opened 2021-04-06
Prevent empty references in test data when running sacrebleu

fhieber:
Recent versions of sacrebleu protect against empty references. This changes the temporary test data for `_test_other_cli` to not create empty test data target strings.
It also fixes the reading of the test data hypotheses (extract translation from json output).

This should fix the failing nightly builds (after #947 is merged).

until you can check this box.

### DIFF
--- a/test/integration/test_seq_copy_int.py
+++ b/test/integration/test_seq_copy_int.py
@@ -178,7 +178,7 @@ def test_other_clis(train_params: str, translate_params: str):
                             dev_line_count=_DEV_LINE_COUNT,
                             dev_max_length=_LINE_MAX_LENGTH,
                             test_line_count=_TEST_LINE_COUNT,
-                            test_line_count_empty=_TEST_LINE_COUNT_EMPTY,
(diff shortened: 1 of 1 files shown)


test/integration/test_seq_copy_int.py
```diff
@@ -37,7 +37,7 @@
 _TRAIN_LINE_COUNT_EMPTY = 1
 _DEV_LINE_COUNT = 5
 _TEST_LINE_COUNT = 5
-_TEST_LINE_COUNT_EMPTY = 2
+_TEST_LINE_COUNT_EMPTY = 0
```
tdomhan, line 40: Isn't this to test that Sockeye handles empty inputs? If so would it be an option to filter sentences before passing them to Sacrebleu instead?

fhieber: true, I'll just pass 0 to the `test_other_clis` invocation then.

tdomhan (approved):

PR #953, opened 2021-06-09
Update Sockeye to MXNet/Gluon 2.x

fhieber:
This updates Sockeye to the APIs introduced in MXNet 2 (i.e. the mxnet/master branch).

#### Major changes with MX2:
- `*Output` operators have been removed with MXNet v2. Therefore we are required to use our secondary loss implementation without `SoftmaxOutput`.
- MX2 introduces a new interface for HybridBlocks. Users do not have to implement `hybrid_forward`, but only `forward`, which uses Parameter class members directly (i.e. `self.weight.data()`), and the `mx.nd` namespace (no more '`F`'). One  consequence of this is that we cannot use our pattern of implementing both `forward` and `hybrid_forward` for HybridBlocks to have shape-inspecting logic/computation in HybridBlocks. We resolve this by introducing the same pattern using `__call__` and `forward` :)
- MX numpy arrays do not implement `__hash__`, therefore we cannot use the lru_cache trick in `OutputLayer`. I worked around this by writing a hacky minimal 'cache', hasing on the string representation of the array.
- Simplify state caching in inference as numpy arrays can have size 0 in dimensions, therefore we do not require 'dummy states of size'. -> this can have a small positive effect on inference speed for self-attentional decoders, since we can save a slice op.

#### Known/Open Issues:
- Sparse gradient updates for embeddings do not work for mx numpy arrays. This is disabled for now along with a logger warning. MXNet team is looking into building a workaround for this.

#### Changes:
- [x] Fixed a bug with predicted output lengths in beam search (missing a repeat op to scale to beam size) -- this bug is also in master currently.
- [x] Fixing imports due to changed locations of modules
- [x] removing prefix logic in all blocks for SockeyeModel (not done for loss.py yet)
- [x] hybrid_forward -> forward for HybridBlocks for SockeyeModel
- [x] simplified parameter sharing: one can simply assign parameters from another block.
- [x] Removed the PrintValue custom op
- [x] removed `with self.name_scope()` everywhere.
- [x] Changes to inference (hybrid blocks for beam search)
- [x] Switch to `mx.np` -- the new Numpy Array interface
- [x] Numpy indexing/broadcasting requires `(x, 1)` shapes for various auxiliary arrays in beam search (`finished, inactive` etc.) 
- [x] Changed saving/loading logic of prepared data to conform with numpy standards. 
- [ ] Write more detailed changelog for 2.4.0.
- [x] check if np.copy is still needed in encoders -- it is

#### Potential future work
- Explore optimizing when and how to convert data back from MXnet numpy arrays to lists in inference (and whether that should be done on the inference device, GPUs, or on the CPU
- Explore optimizing writing to bucket arrays in chunks instead of for loops for multi-factored data
- Lexical constrained decoding could probably be sped up/vectorized with all the availble operators/indexing from numpy now

#### Related links:
- MXNet update to v2 graph api for default blocks: https://github.com/apache/incubator-mxnet/pull/20262 (also updates to npx namespace, which this PR does not do yet)
- https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/block.py#L1620 (to see how the v2 Graph API works).

Note: This PR changes the Github action for pull requests to install a nightly build of MX2.

until you can check this box.
- [ ] Updated CHANGELOG.md


fhieber:
I am running into a few problems with the `ssru_transformer` since this commit: f68c251.
The system test covering the ssru implementation complains about a shape error:
```
mxnet.base.MXNetError: MXNetError: Error in operator node_121: [08:48:56] ../src/operator/numpy/../tensor/elemwise_binary_broadcast_op.h:68: Check failed: l == 1 || r == 1: operands could not be broadcast together with shapes [0,32] [18,32]
```

I haven't been able to track this down to a specific line because without hybridization it works. Moving back the SSRU class to the v1 Graph API does not fix the problem.

.github/workflows/push_pr.yml
```diff
@@ -32,6 +32,8 @@ jobs:
       run: pip install -r requirements/requirements.txt
     - name: Development requirements
       run: pip install -r requirements/requirements.dev.txt
+    - name: Install MX2
+      run: pip install --pre -f https://dist.mxnet.io/python mxnet==2.0.0b20210707 --upgrade
```
fhieber, line 36: this works for macos builds, but for some reason the linux nightly build for that date hasn't been published. will monitor this.

fhieber:
unit & integration tests passing now

PR #960, opened 2021-08-31
Update README.md with a new publication

bricksdont:
Dear all

Here is another publication that uses Sockeye. In my biased opinion, MBR for NMT decoding is also quite interesting :-).


fhieber (approved):
Thanks Mathias!

PR #962, opened 2021-09-01
Parallelized prepare_data

celinadongye:
Parallelized prepare_data.py with multiprocessing pool.

- [x] Updated `sockeye/train.py`
until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.3.21]
+### Added
+
+- Added parallelization for prepare_data
+
 ## [2.3.20]
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Each version section may have have subsections for: _Added_, _Changed_, _Removed
 ## [2.3.21]
 ### Added
 
-- Added parallelization for prepare_data
+- Extended parallelization of data preparation to vocabulary and statistics creation while minimizing the overhead of sharding.
 
 ## [2.3.20]
 ### Added
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -404,7 +404,7 @@ class DataStatisticsAccumulator:
 def create_shards(source_fnames: List[str],
                   target_fnames: List[str],
                   num_shards: int,
-                  output_prefix: str) -> List[Tuple[Tuple[str], Tuple[str]]]:
+                  output_prefix: str) -> Tuple[List[Tuple[Tuple[str], Tuple[str]]], bool]:
     """
     Assign source/target sentence pairs to shards at random.
 
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -301,17 +301,18 @@ def are_none(sequences: Sequence[Sized]) -> bool:
     return all(s is None for s in sequences)
 
 
-def are_token_parallel(sequences: Sequence[Sized], binary: bool = False) -> bool:
+def are_token_parallel(sequences: Sequence[Sized]) -> bool:
     """
     Returns True if all sequences in the list have the same length.
     """
     if not sequences or len(sequences) == 1:
             return True
     else:
-        if not binary:
-            return all(len(s) == len(sequences[0]) for s in sequences)
-        else:
-            return all(len(s.split()) == len(sequences[0].split()) for s in sequences)
+        # if not binary:
+        return all(len(s) == len(sequences[0]) for s in sequences)
+        # else:
+        #     print('ARE TOKENS PARALLEL BINARY MODE')
+        #     return all(len(s.split()) == len(sequences[0].split()) for s in sequences)
 
 
 class DataStatisticsAccumulator:
@@ -433,7 +434,7 @@ def create_shards(source_fnames: List[str],
         target_readers = [exit_stack.enter_context(smart_open(f, mode="rb")) for f in target_fnames]
 
         random_shard_iter = iter(lambda: random.randrange(num_shards), None)
-        for (sources, targets), random_shard_index in zip(parallel_iter(source_readers, target_readers, True, True), random_shard_iter):
+        for (sources, targets), random_shard_index in zip(parallel_iter(source_readers, target_readers, True, False), random_shard_iter):
             random_shard_index = cast(int, random_shard_index)
             for i, line in enumerate(sources):
                 file = sources_shards[i][random_shard_index]
@@ -603,8 +604,7 @@ def save_shard(shard_idx: int,
                                                        length_ratio_mean, length_ratio_std)
 
     # Shards contain the raw sentences. Need to map to integers using the vocabs and add BOS/EOS
-    sources_sentences = [SequenceReader(source, vocab, add_eos=True) for source, vocab in zip(shard_sources, source_vocabs)]
-    targets_sentences = [SequenceReader(target, vocab, add_bos=True) for target, vocab in zip(shard_targets, target_vocabs)]
+    sources_sentences, targets_sentences = create_sequence_readers(shard_sources, shard_targets, source_vocabs, target_vocabs)
 
     for sources, targets in parallel_iter(sources_sentences, targets_sentences):
         source_len = len(sources[0])
@@ -680,7 +680,7 @@ def prepare_data(source_fnames: List[str],
                                            pad_id=C.PAD_ID)
 
 
-    # Process shards in parallel using max_processes process
+    # Process shards in parallel
     args = ((shard_idx, data_loader, shard_sources, shard_targets, source_vocabs,
              target_vocabs, length_statistics.length_ratio_mean, length_statistics.length_ratio_std,
              buckets, output_prefix, keep_tmp_shard_files) for shard_idx, (shard_sources, shard_targets) in enumerate(shards))
@@ -1286,7 +1286,7 @@ def create_sequence_readers(sources: List[str], targets: List[str],
                             vocab_sources: List[vocab.Vocab],
                             vocab_targets: List[vocab.Vocab]) -> Tuple[List[SequenceReader], List[SequenceReader]]:
     """
-    Create source readers with EOS and target readers with BOS if vocabularies are passed.
+    Create source readers with EOS and target readers with BOS.
 
     :param sources: The file names of source data and factors.
     :param targets: The file name of the target data and factors.
@@ -1294,10 +1294,6 @@ def create_sequence_readers(sources: List[str], targets: List[str],
     :param vocab_targets: The target vocabularies.
     :return: The source sequence readers and the target reader.
     """
-    # if not vocab_sources or not vocab_targets:
-    #     source_sequence_readers = [SequenceReader(source, None, add_eos=False) for source in sources]
-    #     target_sequence_readers = [SequenceReader(target, None, add_bos=False) for target in targets]
-    # else:
     source_sequence_readers = [SequenceReader(source, vocab, add_eos=True) for source, vocab in
                                 zip(sources, vocab_sources)]
     target_sequence_readers = [SequenceReader(target, vocab, add_bos=True) for target, vocab in
@@ -1308,7 +1304,7 @@ def create_sequence_readers(sources: List[str], targets: List[str],
 def parallel_iter(source_iterables: Sequence[Iterable[Optional[Any]]],
                   target_iterables: Sequence[Iterable[Optional[Any]]],
                   skip_blanks: bool = True,
-                  binary: bool = False):
+                  check_token_parallel: bool = True):
     """
     Creates iterators over parallel iterables by calling iter() on the iterables
     and chaining to parallel_iterate(). The purpose of the separation is to allow
@@ -1317,18 +1313,18 @@ def parallel_iter(source_iterables: Sequence[Iterable[Optional[Any]]],
     :param source_iterables: A list of source iterables.
     :param target_iterables: A target iterable.
     :param skip_blanks: Whether to skip empty target lines.
-    :param binary: Whether the sequence is in binary.
+    :param check_token_parallel: Whether to check if the tokens are parallel or not.
     :return: Iterators over sources and target.
     """
     source_iterators = [iter(s) for s in source_iterables]
     target_iterators = [iter(t) for t in target_iterables]
-    return parallel_iterate(source_iterators, target_iterators, skip_blanks, binary)
+    return parallel_iterate(source_iterators, target_iterators, skip_blanks, check_token_parallel)
 
 
 def parallel_iterate(source_iterators: Sequence[Iterator[Optional[Any]]],
                      target_iterators: Sequence[Iterator[Optional[Any]]],
                      skip_blanks: bool = True,
-                     binary: bool = False):
+                     check_token_parallel: bool = True):
     """
     Yields parallel source(s), target sequences from iterables.
     Checks for token parallelism in source sequences.
@@ -1339,7 +1335,7 @@ def parallel_iterate(source_iterators: Sequence[Iterator[Optional[Any]]],
     :param source_iterators: A list of source iterators.
     :param target_iterators: A list of source iterators.
     :param skip_blanks: Whether to skip empty target lines.
-    :param binary: Whether the sequence is in binary.
+    :param check_token_parallel: Whether to check if the tokens are parallel or not.
     :return: Iterators over sources and target.
     """
     num_skipped = 0
@@ -1352,10 +1348,11 @@ def parallel_iterate(source_iterators: Sequence[Iterator[Optional[Any]]],
         if skip_blanks and (any((s is None for s in sources)) or any((t is None for t in targets))):
             num_skipped += 1
             continue
-        check_condition(are_none(sources) or are_token_parallel(sources, binary),
-                        "Source sequences are not token-parallel: %s" % (str(sources)))
-        check_condition(are_none(targets) or are_token_parallel(targets, binary),
-                        "Target sequences are not token-parallel: %s" % (str(targets)))
+        if check_token_parallel:
+            check_condition(are_none(sources) or are_token_parallel(sources),
+                            "Source sequences are not token-parallel: %s" % (str(sources)))
+            check_condition(are_none(targets) or are_token_parallel(targets),
+                            "Target sequences are not token-parallel: %s" % (str(targets)))
         yield sources, targets
 
     if num_skipped > 0:
--- a/sockeye/data_io.py
+++ b/sockeye/data_io.py
@@ -400,7 +400,7 @@ class DataStatisticsAccumulator:
 def create_shards(source_fnames: List[str],
                   target_fnames: List[str],
                   num_shards: int,
-                  output_prefix: str) -> Tuple[List[Tuple[Tuple[str], Tuple[str]]], bool]:
+                  output_prefix: str) -> Tuple[List[Tuple[Tuple[str, ...], Tuple[str, ...]]], bool]:
     """
     Assign source/target sentence pairs to shards at random.
 
@@ -640,7 +640,7 @@ def prepare_data(source_fnames: List[str],
                  bucket_scaling: bool = True,
                  keep_tmp_shard_files: bool = False,
                  pool: multiprocessing.pool.Pool = None, 
-                 shards: List[Tuple[Tuple[str], Tuple[str]]] = None):
+                 shards: List[Tuple[Tuple[str, ...], Tuple[str, ...]]] = None):
     """
     :param shards: List of num_shards shards of parallel source and target tuples which in turn contain tuples to shard data factor file paths.
     """
@@ -685,7 +685,7 @@ def prepare_data(source_fnames: List[str],
     shard_average_len = [shard_stats.average_len_target_per_bucket for shard_stats in per_shard_statistics]
     shard_num_sents = [shard_stats.num_sents_per_bucket for shard_stats in per_shard_statistics]
     num_sents_per_bucket = [sum(n) for n in zip(*shard_num_sents)]
-    average_len_target_per_bucket = []
+    average_len_target_per_bucket = [] # type: List[Optional[float]]
     for num_sents_bucket, average_len_bucket in zip(zip(*shard_num_sents), zip(*shard_average_len)):
         if all(avg is None for avg in average_len_bucket):
             average_len_target_per_bucket.append(None)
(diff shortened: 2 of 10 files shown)


sockeye/data_io.py
```diff
@@ -396,79 +396,47 @@ def statistics(self):
...
 
-    data_stats_accumulator = DataStatisticsAccumulator(buckets, source_vocabs[0], target_vocabs[0],
-                                                       length_ratio_mean, length_ratio_std)
-    per_shard_stat_accumulators = [
-        DataStatisticsAccumulator(buckets, source_vocabs[0], target_vocabs[0], length_ratio_mean,
-                                  length_ratio_std) for shard_idx in range(num_shards)]
-
     with ExitStack() as exit_stack:
-        sources_shards = [[exit_stack.enter_context(smart_open(f, mode="wt")) for f in sources_shard_fnames[i]] for i in
+        sources_shards = [[exit_stack.enter_context(smart_open(f, mode="wb")) for f in sources_shard_fnames[i]] for i in
```
tdomhan, line 419: When we read and write in binary mode we could potentially speed up the splitting. In the function here I see that we open the output files as binary, but do read/parse the files as unicode by using the sequence readers. Would this still speed things up compared to `wt`?

I would either:
1. Keep `wt` (unless there is indeed a speed advantage)
2. or try a version of split that reads/writes lines in binary without using the sequence reader to keep splitting minimal.

What do you think?

sockeye/data_io.py
```diff
@@ -1243,10 +1272,12 @@ def __init__(self,
 
     def __iter__(self):
         for tokens in read_content(self.path, self.limit):
+            # TODO: if read_content in bytes, tokens will be in bytes. This might be a problem for tokens2ids
             if self.vocab is not None:
                 sequence = tokens2ids(tokens, self.vocab)
             else:
-                sequence = strids2ids(tokens)
+                sequence = tokens # split input data into shards
+                # sequence = strids2ids(tokens)
```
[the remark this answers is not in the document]
tdomhan: Currently the unit tests for SequenceReader are also failing. If we go back to the original implementation this should it though.

tdomhan (changes requested):
Thanks Celina!

The code looks good to me.
We should still:
- [ ] decide whether we should support the `max_processes=1` use case (for all parts where the `pool` is used).
- [ ] Make sure the existing CLIs/use cases work and go through the tests to make sure they pass.

sockeye/data_io.py
celinadongye: There was a slight speedup with `wb` compared to `wt` in my experiments with the smaller dataset. I can run both versions on the larger dataset and if there is no significant speedup then I'll keep `wt`, otherwise I'll try (2).

sockeye/data_io.py
```diff
@@ -396,79 +396,47 @@ def statistics(self):
...
-            data_stats_accumulator.sequence_pair(sources[0], targets[0], buck_idx)
-            per_shard_stat_accumulators[random_shard_index].sequence_pair(sources[0], targets[0], buck_idx)
-
-            if buck is None:
-                continue
-
             for i, line in enumerate(sources):
-                print(ids2strids(line), file=sources_shards[i][random_shard_index])
+                file = sources_shards[i][random_shard_index]
+                file.write((ids2strids(line)+"\n").encode(encoding="utf-8"))
```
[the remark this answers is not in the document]
tdomhan: If we make the sharding 'minimal' as in writing and reading input lines (in binary format) without using SequenceReaders etc. then we wouldn't need to evaluate different string formats. So we should maybe first decide on the type of sharding (minimal vs using SeqReader) before this :)

sockeye/data_io.py
```diff
@@ -631,20 +625,25 @@ def prepare_data(source_fnames: List[str],
...
     vocab.save_source_vocabs(source_vocabs, output_prefix)
     vocab.save_target_vocabs(target_vocabs, output_prefix)
 
     # Pass 1: get target/source length ratios.
-    length_statistics = analyze_sequence_lengths(source_fnames, target_fnames, source_vocabs, target_vocabs,
-                                                 max_seq_len_source, max_seq_len_target)
+    stats_args = ((source_path, target_path, source_vocabs, target_vocabs, max_seq_len_source, max_seq_len_target) for source_path, target_path in shards)
+    length_stats = pool.starmap(analyze_sequence_lengths, stats_args)
+    num_sents = sum([stat.num_sents for stat in length_stats])
+    length_ratio_mean = sum([stat.num_sents * stat.length_ratio_mean for stat in length_stats]) / num_sents
```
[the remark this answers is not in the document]
celinadongye: The num_sents for each shard, since the sentences are currently being assigned to shards randomly, in an extreme case a shard could be empty. But the final num_sents would only be 0 if the input file is empty.

sockeye/data_io.py
```diff
@@ -619,9 +611,11 @@ def save_shard(shard_idx: int, data_loader: RawParallelDatasetLoader,
         for f in chain(shard_sources, shard_targets):
             os.remove(f)
 
+    return shard_stat_accumulator.statistics
+
 
-def prepare_data(source_fnames: List[str],
-                 target_fnames: List[str],
+def prepare_data(source_fnames: Tuple[Tuple[str]],
```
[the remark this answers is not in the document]
celinadongye: My bad, that's still List[str], the original input source and target filenames. They're used for the data info and data config. But yes, I'll add the docstring to describe the `shards` parameter.

sockeye/prepare_data.py
```diff
@@ -69,19 +69,37 @@ def prepare_data(args: argparse.Namespace):
...
+    logger.info("%d samples will be split into %d shard(s) (requested samples/shard=%d, min_num_shards=%d)."
+                % (num_sents, num_shards, samples_per_shard, minimum_num_shards))
+    shards = data_io.create_shards(source_fnames=source_paths,
+                                   target_fnames=target_paths,
+                                   num_shards=num_shards,
+                                   output_prefix=output_folder)
+    shard_source_paths, shard_target_paths = [paths for paths in zip(*shards)]
+
+    # Process shards in parallel using max_processes process    
+    pool = multiprocessing.pool.Pool(processes=args.max_processes)
```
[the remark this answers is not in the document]
celinadongye: I agree, it would get quite complicated...
What was the main reason for having a variant for the single process? Was it because execution was slower with a pool of a single worker process?

sockeye/prepare_data.py
```diff
@@ -94,12 +112,15 @@ def prepare_data(args: argparse.Namespace):
                          max_seq_len_target=max_seq_len_target,
                          bucketing=bucketing,
                          bucket_width=bucket_width,
-                         samples_per_shard=samples_per_shard,
-                         min_num_shards=minimum_num_shards,
+                         num_shards=num_shards,
                          output_prefix=output_folder,
                          bucket_scaling=bucket_scaling,
-                         max_processes=args.max_processes)
+                         max_processes=args.max_processes,
```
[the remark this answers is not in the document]
celinadongye: You're right, not needed :)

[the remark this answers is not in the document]
celinadongye: You're right, not needed :) , then the case for processing shards sequentially can be removed too.

sockeye/data_io.py
```diff
@@ -656,50 +655,79 @@ def prepare_data(source_fnames: List[str],
...
-        pool = multiprocessing.pool.Pool(processes=max_processes)
-        for shard_idx, (shard_sources, shard_targets, shard_stats) in enumerate(shards):
-            args = (shard_idx, data_loader, shard_sources, shard_targets,
-                    shard_stats, output_prefix, keep_tmp_shard_files)
+        per_shard_statistics = [None] * num_shards
+        for shard_idx, (shard_sources, shard_targets) in enumerate(shards):
+            args = (shard_idx, data_loader, shard_sources, shard_targets, source_vocabs,
+                    target_vocabs, length_statistics.length_ratio_mean, length_statistics.length_ratio_std,
+                    buckets, output_prefix, keep_tmp_shard_files)
             result = pool.apply_async(save_shard, args=args)
```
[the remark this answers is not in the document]
celinadongye: Yes, i tried with `map` as well and execution time was the same but the code is neater, so I will switch to the `map`.

sockeye/data_io.py
```diff
@@ -656,50 +655,79 @@ def prepare_data(source_fnames: List[str],
...
+            per_shard_statistics[shard_idx] = result.get()
+    
+
+    # Combine per shard statistics to obtain global statistics
+    shard_average_len = [shard_stats.average_len_target_per_bucket for shard_stats in per_shard_statistics]
+    shard_num_sents = [shard_stats.num_sents_per_bucket for shard_stats in per_shard_statistics]
+    num_sents_per_bucket = [sum(n) for n in zip(*shard_num_sents)]
+    average_len_target_per_bucket = []
+    for num_sents_bucket, average_len_bucket in zip(zip(*shard_num_sents), zip(*shard_average_len)):
+        if None in average_len_bucket:
```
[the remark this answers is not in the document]
celinadongye: Yes, but when I was checking against the results of the current implementation, when the bucket was empty it was returning a tuple `(None, None)` rather than `(0, None)`.

sockeye/prepare_data.py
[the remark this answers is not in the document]
tdomhan: I think mainly to not have the overhead of `pool` when there is only a single process/single shard. If we wanted to support single process without pool the easiest would probably to define a dummy `Pool` that implements all relevant pool methods but uses Python `map` etc. instead.

sockeye/data_io.py
```diff
@@ -396,79 +396,47 @@ def statistics(self):
...
     :param num_shards: The total number of shards.
-    :param buckets: Bucket list.
-    :param length_ratio_mean: Mean length ratio.
-    :param length_ratio_std: Standard deviation of length ratios.
     :param output_prefix: The prefix under which the shard files will be created.
-    :return: Tuple of source (and source factor) file names, target (and target factor) file names
-             and statistics for each shard, as well as global statistics.
+    :return: Tuple of source (and source factor) file names and target (and target factor) file names for each shard.
     """
     os.makedirs(output_prefix, exist_ok=True)
```
tdomhan, line 412: Could we support the special case of `num_shards=1` by returning the paths of the original files?

sockeye/utils.py
```diff
@@ -233,9 +233,47 @@ def smart_open(filename: str, mode: str = "rt", ftype: str = "auto", errors: str
...
+                return gzip.open(filename, mode=mode, encoding='utf-8', errors=errors)
     else:
-        return open(filename, mode=mode, encoding='utf-8', errors=errors)
+        if mode == "rb" or mode == "wb":
+            return open(filename, mode=mode)
+        else:
+            return open(filename, mode=mode, encoding='utf-8', errors=errors)
+
+
+def combine_means(means: List[float], num_sents: List[int]) -> float:
```
tdomhan, line 247: Thanks for extracting this logic into separate functions!

Given that floats could be `None` should the type be `List[Optional[float]]`?

Could you add a unit test for combine_means and combine_stds?

sockeye/train.py
```diff
@@ -1026,9 +1026,7 @@ def train(args: argparse.Namespace, custom_metrics_logger: Optional[Callable] =
                                            max_seq_len_target=max_seq_len_target,
                                            config_data=config_data)
 
-        training_model = model.SockeyeModel(
-            model_config,
-            train_decoder_only=args.fixed_param_strategy == C.FIXED_PARAM_STRATEGY_ALL_EXCEPT_DECODER)
+        training_model = model.SockeyeModel(model_config)
```
tdomhan, line 1029: It seems like there are some recent changes that show as 'deleted'. Could you merge from master to make sure no recent changes get reverted?

sockeye/data_io.py
tdomhan: Thanks!

test/unit/test_utils.py
```diff
@@ -238,6 +238,35 @@ def test_get_tokens(line, expected_tokens):
     tokens = list(utils.get_tokens(line))
     assert tokens == expected_tokens
 
+def test_combine_means():
+    error_tolerance = 1e-05
+    n_lists = np.random.randint(1, 10)
```
tdomhan, line 243: Thanks for adding tests! I think for tests we should avoid randomness to make results deterministic. We could either fix a seed or simply specify a list of fixed data. I'd prefer the latter as it's more explicit and wouldn't fail if the underlying random number generator would change in the future. 

Given that the inputs can also be `None`, we could use `@pytest.mark.parametrize` (see above) to specify some lists where some entries are None to test this works as expected..

sockeye/data_io.py
```diff
@@ -656,50 +674,57 @@ def prepare_data(source_fnames: List[str],
...
-            results.append(result)
-        pool.close()
-        pool.join()
-
-        for result in results:
-            if not result.successful():
-                logger.error("Process ended in error.")
-                raise RuntimeError("Shard processing failed.")
+
+    # Process shards in parallel using max_processes process
```
tdomhan, line 683: nit: I'd remove using max_processes process
```suggestion
    # Process shards in parallel
```

sockeye/data_io.py
```diff
@@ -1261,24 +1286,29 @@ def create_sequence_readers(sources: List[str], targets: List[str],
                             vocab_sources: List[vocab.Vocab],
                             vocab_targets: List[vocab.Vocab]) -> Tuple[List[SequenceReader], List[SequenceReader]]:
     """
-    Create source readers with EOS and target readers with BOS.
+    Create source readers with EOS and target readers with BOS if vocabularies are passed.
```
tdomhan, line 1289: The comment should be updated to match the code.
```suggestion
    Create source readers with EOS and target readers with BOS.
```

sockeye/data_io.py
```diff
@@ -1261,24 +1286,29 @@ def create_sequence_readers(sources: List[str], targets: List[str],
...
 
     :param sources: The file names of source data and factors.
     :param targets: The file name of the target data and factors.
     :param vocab_sources: The source vocabularies.
     :param vocab_targets: The target vocabularies.
     :return: The source sequence readers and the target reader.
     """
+    # if not vocab_sources or not vocab_targets:
```
tdomhan, line 1297: Dead code.

sockeye/data_io.py
```diff
@@ -592,23 +568,54 @@ def get_num_shards(num_samples: int, samples_per_shard: int, min_num_shards: int
...
     """
-    sources_sentences = [SequenceReader(s) for s in shard_sources]
-    targets_sentences = [SequenceReader(s) for s in shard_targets]
+
+    # Compute shard statistics and bucketing
+    shard_stat_accumulator = DataStatisticsAccumulator(buckets, source_vocabs[0], target_vocabs[0],
+                                                       length_ratio_mean, length_ratio_std)
+
+    # Shards contain the raw sentences. Need to map to integers using the vocabs and add BOS/EOS
+    sources_sentences = [SequenceReader(source, vocab, add_eos=True) for source, vocab in zip(shard_sources, source_vocabs)]
```
tdomhan, line 606: Could you call `create_sequence_readers`?

```suggestion
    sources_sentences, target_sentences = create_sequence_readers(shard_sources, shard_targets, source_vocabs, target_vocabs)
```

sockeye/data_io.py
```diff
@@ -1261,24 +1286,29 @@ def create_sequence_readers(sources: List[str], targets: List[str],
...
-                               zip(targets, vocab_targets)]
+                                zip(targets, vocab_targets)]
     return source_sequence_readers, target_sequence_readers
 
 
 def parallel_iter(source_iterables: Sequence[Iterable[Optional[Any]]],
                   target_iterables: Sequence[Iterable[Optional[Any]]],
-                  skip_blanks: bool = True):
+                  skip_blanks: bool = True,
+                  binary: bool = False):
```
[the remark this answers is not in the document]
tdomhan: I'd still rename. I'd even consider adding a flag `check_token_parallel` which we could set to False for sharding and to True later when we load the shard (in the spirit of keeping the sharding logic minimal).

sockeye/utils.py
tdomhan:
```suggestion
def combine_means(means: List[Optional[float]], num_sents: List[int]) -> float:
```

sockeye/utils.py
```diff
@@ -233,9 +235,47 @@ def smart_open(filename: str, mode: str = "rt", ftype: str = "auto", errors: str
...
+    :param num_sents: A list with the number of sentences used to compute each mean value.
+    :return: The combined mean of the list of means.
+    """
+    if not means or not num_sents:
+        raise ValueError("Invalid input list.")
+    check_condition(len(means) == len(num_sents), "List lengths do not match")
+    return sum(num_sent * mean for num_sent, mean in zip(num_sents, means) if mean is not None) / sum(num_sents)
+
+
+def combine_stds(stds: List[float], means: List[float], num_sents: List[int]) -> float:
```
tdomhan, line 263:
```suggestion
def combine_stds(stds: List[Optional[float]], means: List[Optional[float]], num_sents: List[int]) -> float:
```

sockeye/vocab.py
```diff
@@ -32,23 +32,42 @@
 InverseVocab = Dict[int, str]
 
 
-def build_from_paths(paths: List[str], num_words: Optional[int] = None, min_count: int = 1,
-                     pad_to_multiple_of: Optional[int] = None) -> Vocab:
+def build_from_paths(paths: List[str]) -> Counter:
```
tdomhan, line 35: It seems it can either be a single value or a list, so I think we should use the `Union` type from `typing` which specifies this:
```suggestion
def build_from_paths(paths: Union[List[str], str]) -> Counter:
```

tdomhan: This should also then bubbled up if it applies to callers.

sockeye/vocab.py
```diff
@@ -253,36 +263,38 @@ def load_target_vocabs(folder: str) -> List[Vocab]:
...
+                          pad_to_multiple_of: Optional[int] = None,
+                          mapper: Callable = map) -> Tuple[List[Vocab], List[Vocab]]:
     """
     Returns vocabularies for source files (including factors) and target files (including factors.
     If the respective vocabulary paths are not None, the vocabulary is read from the path and returned.
     Otherwise, it is built from the support and saved to the path.
 
-    :param source_paths: The path to the source text (and optional token-parallel factor files).
-    :param target_paths: The path to the target text (and optional token-parallel factor files).
+    :param shard_source_paths: List of shards of list paths to the source text (and optional token-parallel factor files).
```
tdomhan, line 296: The first index is the shard and the second index is the factor, correct? so list[shard][factor]?

sockeye/vocab.py
```diff
@@ -320,65 +332,68 @@ def load_or_create_vocabs(source_paths: List[str],
...
-            source_factor_vocab_same_as_source = source_factor_vocab_same_as_source * len(source_factor_paths)
+            source_factor_vocab_same_as_source = source_factor_vocab_same_as_source * len(shard_source_factor_paths)
         else:
-            source_factor_vocab_same_as_source = [False] * len(source_factor_paths)
+            source_factor_vocab_same_as_source = [False] * len(shard_source_factor_paths)
 
-    for factor_path, factor_vocab_path, share_source_vocab in zip(source_factor_paths,
-                                                                  source_factor_vocab_paths,
-                                                                  source_factor_vocab_same_as_source):
+    for shard_factor_path, factor_vocab_path, share_source_vocab in zip(shard_source_factor_paths,
```
tdomhan, line 365: shard_source_factor_paths is a list of shard_source_factor_paths[factor][shard], right?
Could you add this as a comment and `make shard_factor_path` plural, so `shard_factor_paths` just to make it 'clear' that it is a list.

sockeye/data_io.py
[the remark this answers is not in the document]
celinadongye: Yes, I added that flag because when the file was being read in binary mode the check whether the tokens were parallel failed because the length of binary data did not match.

With the flag for check_token_parallel set to False when sharding then we wouldn't ensure that the tokens are parallel though. I guess it's fine since parallelism is checked later on. If we don't want to check token parallelism when sharding then the `binary` flag I added can be removed.

sockeye/vocab.py
celinadongye: Yes

sockeye/data_io.py
```diff
@@ -306,8 +306,13 @@ def are_token_parallel(sequences: Sequence[Sized]) -> bool:
...
     """
     if not sequences or len(sequences) == 1:
-        return True
-    return all(len(s) == len(sequences[0]) for s in sequences)
+            return True
+    else:
+        # if not binary:
+        return all(len(s) == len(sequences[0]) for s in sequences)
+        # else:
+        #     print('ARE TOKENS PARALLEL BINARY MODE')
```
tdomhan, line 314: dead code :)

tdomhan:
Thanks for iterating! Could you increment the minor version in `sockeye/__init__.py` and add a corresponding entry in the Changelog.md?

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [2.3.21]
+### Added
+
+- Added parallelization for prepare_data
```
tdomhan, line 17:
```suggestion
- Extended parallelization of data preparation to vocabulary and statistics creation while minimizing the overhead of sharding.
```

sockeye/vocab.py
```diff
@@ -32,23 +32,41 @@
...
+    """
+    with ExitStack() as stack:
+        logger.info("Building vocabulary from dataset(s): %s", paths)
+        if isinstance(paths, List):
+            files = (stack.enter_context(utils.smart_open(path, mode='rt')) for path in paths)
+        else:
+            files = stack.enter_context(utils.smart_open(paths, mode='rt'))
+        return count_tokens(chain(*files))
+
+def build_from_shards(paths: Iterable[Union[Tuple[str], str]], num_words: Optional[int] = None, min_count: int = 1,
```
tdomhan, line 48: I'd maybe change from `Tuple` to `Sequence` here and above (which we'd need to import still). For Tuple we'd need the ellipsis.

```suggestion
def build_from_shards(paths: Iterable[Union[Sequence[str], str]], num_words: Optional[int] = None, min_count: int = 1,
```

sockeye/vocab.py
```diff
@@ -32,23 +32,41 @@
 InverseVocab = Dict[int, str]
 
 
-def build_from_paths(paths: List[str], num_words: Optional[int] = None, min_count: int = 1,
-                     pad_to_multiple_of: Optional[int] = None) -> Vocab:
+def build_from_paths(paths: Union[List[str], str]) -> Counter:
```
tdomhan, line 35:
```suggestion
def build_from_paths(paths: Union[Sequence[str], str]) -> Counter:
```

tdomhan (approved):
LGTM. Thanks for the change!

PR #977, opened 2021-11-08
Sockeye 3 - Fast Neural Machine Translation with PyTorch

fhieber:
## [3.0.0] Sockeye 3: Fast Neural Machine Translation with PyTorch

Sockeye is now based on PyTorch. We maintain backwards compatibility with MXNet models in version 2.3.x until 3.1.0. If MXNet 2.x is installed, Sockeye can run both with PyTorch or MXNet but MXNet is no longer strictly required.

### Added

- Added model converter CLI `sockeye.mx_to_pt` that converts MXNet models to PyTorch models.
- Added `--apex-amp` training argument that runs entire model in FP16 mode, replaces `--dtype float16` (requires [Apex](https://github.com/NVIDIA/apex)).
- Training automatically uses Apex fused optimizers if available (requires [Apex](https://github.com/NVIDIA/apex)).
- Added training argument `--label-smoothing-impl` to choose label smoothing implementation (default of `mxnet` uses the same logic as MXNet Sockeye 2).

### Changed

- CLI names point to the PyTorch code base (e.g. `sockeye-train` etc.).
- MXNet-based CLIs are now accessible via `sockeye-<name>-mx`.
- MXNet code requires MXNet >= 2.0 since we adopted the new numpy interface.
- `sockeye-train` now uses PyTorch's distributed data-parallel mode for multi-process (multi-GPU) training. Launch with: `torchrun --no_python --nproc_per_node N sockeye-train --dist ...`
- Updated the [quickstart tutorial](docs/tutorials/wmt_large.md) to cover multi-device training with PyTorch Sockeye.
- Changed `--device-ids` argument (plural) to `--device-id` (singular). For multi-GPU training, see distributed mode noted above.
- Updated default value: `--pad-vocab-to-multiple-of 8`
- `sockeye-translate` and `sockeye-score` now automatically use the CPU device if CUDA is not available. CPU device can still be set 'manually' via `--use-cpu` on a CUDA-enabled machine.

### Removed

- Removed support for constrained decoding (both positive and negative lexical constraints)
- Removed support for beam histories
- Removed `--amp-scale-interval` argument.
- Removed `--kvstore` argument.
- Removed arguments: `--weight-init`, `--weight-init-scale` `--weight-init-xavier-factor-type`, `--weight-init-xavier-rand-type`
- Removed `--decode-and-evaluate-device-id` argument.
- Removed arguments: `--monitor-pattern'`, `--monitor-stat-func`
- Removed CUDA-specific requirements files in `requirements/`
- - Removed `--horovod` argument used with `horovodrun` (use `--dist` with `torchrun`).
- Removed `--optimizer-params` argument (use `--optimizer-betas`, `--optimizer-eps`).
- Removed `--no-hybridization` argument (use `PYTORCH_JIT=0`, see [Disable JIT for Debugging](https://pytorch.org/docs/stable/jit.html#disable-jit-for-debugging)).
- Removed `--omp-num-threads` argument (use `--env=OMP_NUM_THREADS=N`).

## Technical Notes
Push/PR workflow file in this PR is set up to install nightly build of MXNet 2.x from July to also execute equivalence tests between MXNet and PyTorch. Simply using the latest nightly build for MXNet 2.x does not work because linux installs a version from November 2021, while macos installs a version from July 2021. Behavior does not match between these two versions, making it difficult to fix the tests/linters for both cases.

until you can check this box.
- [x] Updated CHANGELOG.md


mjdenkowski (approved):
Sockeye 3 now provides greater functionality than Sockeye 2, except where unused/deprecated features are specifically removed.

Existing and new tests cover both the PyTorch and MXNet implementations, confirming their equivalence with each other and Sockeye 2 where applicable.  The code passes inspection by various automatic tools (Pylint, Mypy, etc.).

The Sockeye 3 PyTorch code is also cleaner than the Sockeye 2 MXNet code it replaces.  Nothing stands out as an obvious workaround (in fact, we replace workarounds from Sockeye 2 with simpler and more direct code).

The code is well documented and the tutorials are updated.

Looks good to merge!

PR #990, opened 2021-12-13
[DONOTMERGE] try windows CI

fhieber:
(description of the change)

until you can check this box.
- [ ] Updated CHANGELOG.md


RamoramaInteractive:
I tested the first tutorial under Windows 10 again (command shell and jupyter notebook), but I only receive

```
NOTE: Redirects are currently not supported in Windows or MacOs.
Mismatch in arguments for training continuation.
Differing arguments: use_cpu.
```
Is there a possible way, that redirects will be supported in Windows? On my Windows 10 PC, PyTorch recognizes my GPU.

mjdenkowski:
The note about redirects is a PyTorch message that shouldn't affect training.

It looks like Sockeye found an existing model that used different arguments: the previous training run used `--use-cpu` and the current doesn't, or the other way around.  By default, Sockeye attempts to continue training when a model directory already exists.  If the arguments aren't compatible, Sockeye exits with that error message.

You can resolve the issue by deleting/moving the previous model directory or specifying a different output location.

RamoramaInteractive:
I did, but the error remains the same.

mjdenkowski:
The args compatibility check is only reachable if the output model directory exists and contains a partial training state ([train_pt.py#L147-L162](https://github.com/awslabs/sockeye/blob/main/sockeye/train_pt.py#L147-L162)).

The safest way to resolve this is to manually move or delete the existing output model directory.  If you're certain that you want to overwrite anything that already exists in the directory, you can pass the `--overwrite-output` arg to `sockeye-train`.

RamoramaInteractive:
Now it's working! Thanks! 😃

RamoramaInteractive:
It's working! Thank you! 😃

PR #996, opened 2021-12-17
jit.script OutputLayer and trace source/target embedding during inference

fhieber:
until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/encoder_pt.py
+++ b/sockeye/encoder_pt.py
@@ -90,14 +90,17 @@ class PyTorchEmbedding(PyTorchEncoder):
     def __init__(self, config: EmbeddingConfig, embedding: Optional[pt.nn.Embedding] = None) -> None:
         super().__init__()
         self.config = config
-
         if embedding is not None:
             self.embedding = embedding
         else:
             self.embedding = pt.nn.Embedding(self.config.vocab_size, self.config.num_embed,
                                              sparse=self.config.allow_sparse_grad)
 
+        self.num_factors = self.config.num_factors
+        if self.num_factors > 1:
+            assert len(self.config.factor_configs) == self.num_factors - 1
         self.factor_embeds = pt.nn.ModuleList()
+        self.factor_combinations = []  # type: List[str]
         if self.config.factor_configs is not None:
             for i, fc in enumerate(self.config.factor_configs, 1):
                 if fc.share_embedding:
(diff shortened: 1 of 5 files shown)


sockeye/layers_pt.py
```diff
@@ -101,17 +102,16 @@ def __init__(self,
             self.weight = weight
         self.bias = pt.nn.Parameter(pt.Tensor(vocab_size))
 
-        self.previous_slice_ids = None  # type: Optional[pt.Tensor]
-        self.reduced_weight_bias = None  # type: Optional[Tuple[pt.Tensor, pt.Tensor]]
+        self.previous_slice_ids: Optional[pt.Tensor] = pt.empty(0)
+        self.reduced_weight: Optional[pt.Tensor] = pt.empty(0)
+        self.reduced_bias: Optional[pt.Tensor] = pt.empty(0)
```
mjdenkowski, line 107: Do these still need to be `Optional` (or have type annotations at all) now that they're created as `Tensor`s?

mjdenkowski (commented):
It looks like this still needs a version/CHANGELOG update.

Other than that and the minor typing comment, everything should be ready to merge.

sockeye/layers_pt.py
fhieber: Good catch, we can remove the type annotation.

sockeye/encoder_pt.py
```diff
@@ -90,14 +90,17 @@ class PyTorchEmbedding(PyTorchEncoder):
...
-
         if embedding is not None:
             self.embedding = embedding
         else:
             self.embedding = pt.nn.Embedding(self.config.vocab_size, self.config.num_embed,
                                              sparse=self.config.allow_sparse_grad)
 
+        self.num_factors = self.config.num_factors
+        if self.num_factors > 1:
+            assert len(self.config.factor_configs) == self.num_factors - 1
```
tdomhan, line 101: Do we expect this to be violated?
If we want to keep it should it be converted to `check_condition` + a message?

mjdenkowski (approved):

sockeye/encoder_pt.py
fhieber: No, I think its safe to assume this is always true. I will remove the assertion.

PR #1021, opened 2022-02-11
Allow torch==1.10.2, rename unit test file

fhieber:
Previous `pip install sockeye` would uninstall `torch==1.10.2` and install `torch==1.10.0`. We can safely depend on `torch>=1.10.0` (1.10 and later).

Also rename a unit test module that was missed in the previous update.

until you can check this box.
- [x] Updated CHANGELOG.md


requirements/requirements.txt
```diff
@@ -1,4 +1,4 @@
-torch==1.10.0
+torch>=1.10.0
```
mjdenkowski, line 1: Do we want to use `torch>=1.10.0,<1.11.0`?

I'm not aware of any (intended) backward-incompatible changes in PyTorch 1.11, but using a min/max convention can guard against unexpected breaks until we have a chance to test new major versions.

fhieber: yeah good point.

mjdenkowski (approved):

PR #1024, opened 2022-02-19
support JSON source prefix

hoangcuong2011:
This version makes Sockeye able to support JSON input format with special source tokens as follows: 

  1. Special source tokens can be added (e.g. verbosity token, formality token, multilingual token) in Sockeye's JSON input format
  2. Long input sequence is chunked in the way that the special source tokens are added to each chunk.

Examples of JSON input object that can be used during inference are as below (noticed of the keyword: `source_prefix` and `source_prefix_factors`):
```
{ "text": "(@@ A@@", "source_prefix": "a@@"}
{ "text": "(@@ A@@", "source_prefix": "a@@ a@@"} 
{ "text": "(@@ A@@", "factors": ["(@@ A@@"], "source_prefix": "a@@ a@@", "source_prefix_factors": ["(@@ A@@"]}

```
until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/inference.py
+++ b/sockeye/inference.py
@@ -130,7 +130,7 @@ class TranslatorInput:
     sentence_id: SentenceId
     tokens: Tokens
     factors: Optional[List[Tokens]] = None
-    source_prefix: Optional[Tokens] = None
+    source_prefix_tokens: Optional[Tokens] = None
     source_prefix_factors: Optional[List[Tokens]] = None
     restrict_lexicon: Optional[lexicon.TopKLexicon] = None
     constraints: Optional[List[Tokens]] = None
@@ -138,8 +138,8 @@ class TranslatorInput:
     pass_through_dict: Optional[Dict] = None
 
     def __str__(self):
-        return 'TranslatorInput(%s, %s, factors=%s, source_prefix=%s, source_prefix_factors=%s, constraints=%s, avoid=%s)' \
-            % (self.sentence_id, self.tokens, self.factors, self.source_prefix, self.source_prefix_factors, self.constraints, self.avoid_list)
+        return 'TranslatorInput(%s, %s, factors=%s, source_prefix_tokens=%s, source_prefix_factors=%s, constraints=%s, avoid=%s)' \
+            % (self.sentence_id, self.tokens, self.factors, self.source_prefix_tokens, self.source_prefix_factors, self.constraints, self.avoid_list)
 
     def __len__(self):
         return len(self.tokens)
@@ -150,6 +150,18 @@ class TranslatorInput:
         Returns the number of factors of this instance.
         """
         return 1 + (0 if not self.factors else len(self.factors))
+
+    def get_source_prefix_tokens(self) -> List[str]:
+        """
+        Returns the source prefix tokens of this instance.
+        """
+        return self.source_prefix_tokens if self.source_prefix_tokens is not None else []
+
+    def num_source_prefix_tokens(self) -> int:
+        """
+        Returns the number of source prefix tokens of this instance.
+        """
+        return len(self.get_source_prefix_tokens())
 
     def chunks(self, chunk_size: int) -> Generator['TranslatorInput', None, None]:
         """
@@ -176,7 +188,7 @@ class TranslatorInput:
             yield TranslatorInput(sentence_id=self.sentence_id,
                                   tokens=self.tokens[i:i + chunk_size],
                                   factors=factors,
-                                  source_prefix=self.source_prefix,
+                                  source_prefix_tokens=self.source_prefix_tokens,
                                   source_prefix_factors=self.source_prefix_factors,
                                   restrict_lexicon=self.restrict_lexicon,
                                   constraints=constraints,
@@ -191,7 +203,7 @@ class TranslatorInput:
                                tokens=self.tokens + [C.EOS_SYMBOL],
                                factors=[factor + [C.EOS_SYMBOL] for factor in
                                         self.factors] if self.factors is not None else None,
-                               source_prefix=self.source_prefix,
+                               source_prefix_tokens=self.source_prefix_tokens,
                                source_prefix_factors=self.source_prefix_factors,
                                restrict_lexicon=self.restrict_lexicon,
                                constraints=self.constraints,
@@ -261,9 +273,19 @@ def make_input_from_dict(sentence_id: SentenceId,
         tokens = input_dict[C.JSON_TEXT_KEY]
         tokens = list(utils.get_tokens(tokens))
         factors = input_dict.get(C.JSON_FACTORS_KEY)
-        source_prefix = input_dict.get(C.JSON_SOURCE_PREFIX_KEY)
+        source_prefix_tokens = input_dict.get(C.JSON_SOURCE_PREFIX_KEY)
+        source_prefix_tokens = list(utils.get_tokens(source_prefix_tokens)) if source_prefix_tokens else None
         source_prefix_factors = input_dict.get(C.JSON_SOURCE_PREFIX_FACTORS_KEY)
-        source_prefix = source_prefix.strip() if source_prefix else None
+        if source_prefix_factors is not None and not source_prefix_tokens:
+            logger.error("Source prefix factors cannot be specified when source prefix is not specified")
+            return _bad_input(sentence_id, reason=str(input_dict))
+        if source_prefix_factors is not None and not factors:
+            logger.error("Source prefix factors cannot be specified when source factors are not specified")
+            return _bad_input(sentence_id, reason=str(input_dict))
+        if source_prefix_tokens is not None and (factors is not None and not source_prefix_factors):
+            logger.error("Source prefix factors need to be also specified together with source factors")
+            return _bad_input(sentence_id, reason=str(input_dict))
+
         if isinstance(factors, list):
             factors = [list(utils.get_tokens(factor)) for factor in factors]
             lengths = [len(f) for f in factors]
@@ -273,24 +295,13 @@ def make_input_from_dict(sentence_id: SentenceId,
 
         if isinstance(source_prefix_factors, list):
             source_prefix_factors = [list(utils.get_tokens(source_prefix_factor)) for source_prefix_factor in source_prefix_factors]
-
-        if source_prefix is not None and \
-            (factors is not None and not source_prefix_factors):
-                logger.error("Source prefix factors need to be also specified together with source factors")
+            lengths = [len(source_prefix_factor) for source_prefix_factor in source_prefix_factors]
+            if not all(len(source_prefix_tokens) == length for length in lengths):
+                logger.error("Source prefix has %d tokens but there are %s prefix factors", len(source_prefix_tokens), str(lengths))
                 return _bad_input(sentence_id, reason=str(input_dict))
-        if source_prefix is not None and source_prefix_factors is not None:
-            if not factors:
-                logger.error("Source prefix factors cannot be specified when source factors are not specified")
+            if len(source_prefix_factors) != len(factors):
+                logger.error("There is mismatch in source factors %d and prefix factors %d", len(factors), len(source_prefix_factors))
                 return _bad_input(sentence_id, reason=str(input_dict))
-            else:
-                if len(factors) != len(source_prefix_factors):
-                    logger.error("There is mismatch in source factors %d and prefix factors %d", len(factors), len(source_prefix_factors))
-                    return _bad_input(sentence_id, reason=str(input_dict))
-            lengths = [len(source_prefix_factors[i]) for i in range(len(source_prefix_factors))]
-            if not all(len(source_prefix.split()) == length for length in lengths):
-                logger.error("Source prefix has %d tokens but there are %s prefix factors", len(source_prefix.split()), str(lengths))
-                return _bad_input(sentence_id, reason=str(input_dict))
-
 
         # Lexicon for vocabulary selection/restriction:
         # This is only populated when using multiple lexicons, in which case the
@@ -331,7 +342,7 @@ def make_input_from_dict(sentence_id: SentenceId,
             constraints = [list(utils.get_tokens(constraint)) for constraint in constraints]
 
         return TranslatorInput(sentence_id=sentence_id, tokens=tokens, factors=factors,
-                               source_prefix=source_prefix,
+                               source_prefix_tokens=source_prefix_tokens,
                                source_prefix_factors=source_prefix_factors,
                                restrict_lexicon=restrict_lexicon, constraints=constraints,
                                avoid_list=avoid_list, pass_through_dict=input_dict)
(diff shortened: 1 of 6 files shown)


sockeye/inference.py
```diff
@@ -255,13 +261,37 @@ def make_input_from_dict(sentence_id: SentenceId,
         tokens = input_dict[C.JSON_TEXT_KEY]
         tokens = list(utils.get_tokens(tokens))
         factors = input_dict.get(C.JSON_FACTORS_KEY)
+        source_prefix = input_dict.get(C.JSON_SOURCE_PREFIX_KEY)
```
fhieber, line 264: You define `source_prefix` in `TranslatorInput` as an optional List of Tokens (Tokens are List[str]). I think you should call `utils.get_tokens()` here as well, like for the tokens. It will also save you a number of `split()` calls below.

sockeye/inference.py
```diff
@@ -130,14 +130,16 @@ class TranslatorInput:
...
     pass_through_dict: Optional[Dict] = None
 
     def __str__(self):
-        return 'TranslatorInput(%s, %s, factors=%s, constraints=%s, avoid=%s)' \
-            % (self.sentence_id, self.tokens, self.factors, self.constraints, self.avoid_list)
+        return 'TranslatorInput(%s, %s, factors=%s, source_prefix=%s, source_prefix_factors=%s, constraints=%s, avoid=%s)' \
+            % (self.sentence_id, self.tokens, self.factors, self.source_prefix, self.source_prefix_factors, self.constraints, self.avoid_list)
 
     def __len__(self):
         return len(self.tokens)
```
fhieber, line 145: Should this be reconsidered and take into account the prefix length?

sockeye/inference.py
```diff
@@ -759,16 +791,19 @@ def translate(self, trans_inputs: List[TranslatorInput], fill_up_batches: bool =
                 translated_chunks.append(IndexedTranslation(input_idx=trans_input_idx, chunk_idx=0,
                                                             translation=empty_translation(add_nbest=(self.nbest_size > 1))))
             else:
-                if len(trans_input.tokens) > self.max_input_length:
+                max_input_length = self.max_input_length
+                #take length of source prefix into account while chunking
+                max_input_length = max_input_length if not trans_input.source_prefix else max_input_length - len(trans_input.source_prefix.split())
```
fhieber, line 796: like said above, I think we should make `source_prefix` actually `List[Tokens]` so you don't have to split here and in other places.

sockeye/inference.py
```diff
@@ -862,17 +897,27 @@ def _get_inference_input(self,
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)  # includes eos
             max_output_lengths.append(self._get_max_output_length(num_tokens))
-            source[j, :num_tokens, 0] = tokens2ids(trans_input.tokens, self.source_vocabs[0])
+            num_tokens = num_tokens if not trans_input.source_prefix else num_tokens + len(trans_input.source_prefix.split())
+            if not trans_input.source_prefix:
+                source[j, :num_tokens, 0] = tokens2ids(trans_input.tokens, self.source_vocabs[0])
+            else:
+                #adding source prefix to every chunk
```
fhieber, line 904: [nit] add space before comments

sockeye/inference.py
```diff
@@ -759,16 +791,19 @@ def translate(self, trans_inputs: List[TranslatorInput], fill_up_batches: bool =
                 translated_chunks.append(IndexedTranslation(input_idx=trans_input_idx, chunk_idx=0,
                                                             translation=empty_translation(add_nbest=(self.nbest_size > 1))))
             else:
-                if len(trans_input.tokens) > self.max_input_length:
+                max_input_length = self.max_input_length
+                #take length of source prefix into account while chunking
```
fhieber, line 795: [nit] add space before comment

sockeye/inference.py
```diff
@@ -862,17 +897,27 @@ def _get_inference_input(self,
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)  # includes eos
             max_output_lengths.append(self._get_max_output_length(num_tokens))
-            source[j, :num_tokens, 0] = tokens2ids(trans_input.tokens, self.source_vocabs[0])
+            num_tokens = num_tokens if not trans_input.source_prefix else num_tokens + len(trans_input.source_prefix.split())
+            if not trans_input.source_prefix:
+                source[j, :num_tokens, 0] = tokens2ids(trans_input.tokens, self.source_vocabs[0])
+            else:
+                #adding source prefix to every chunk
+                source[j, :num_tokens, 0] = tokens2ids(trans_input.source_prefix.split() + \
```
fhieber, line 905: maybe we can also add a property to `TranslatorInput` that returns the `source_prefix` or an empty list if the field is `None`, then you can do this in a one-liner without the if:
` source[j, :num_tokens, 0] = tokens2ids(trans_input.source_prefix + trans_input.tokens, self.source_vocabs[0])`

sockeye/inference.py
```diff
@@ -862,17 +897,27 @@ def _get_inference_input(self,
...
             factors = trans_input.factors if trans_input.factors is not None else []
             num_factors = 1 + len(factors)
             if num_factors != self.num_source_factors:
                 logger.warning("Input %d factors, but model(s) expect %d", num_factors,
                                self.num_source_factors)
-            for i, factor in enumerate(factors[:self.num_source_factors - 1], start=1):
-                # fill in as many factors as there are tokens
-
-                source[j, :num_tokens, i] = tokens2ids(factor, self.source_vocabs[i])[:num_tokens]
+            if not trans_input.source_prefix or not trans_input.source_prefix_factors: #no source prefix during inference
```
fhieber, line 913: [nit] space before comment

sockeye/inference.py
```diff
@@ -862,17 +897,27 @@ def _get_inference_input(self,
...
+                for i, factor in enumerate(factors[:self.num_source_factors - 1], start=1):
+                    # fill in as many factors as there are tokens
+                    source[j, :num_tokens, i] = tokens2ids(factor, self.source_vocabs[i])[:num_tokens]
+            else:
```
fhieber, line 917: instead of the if-else, could you maybe use`itertools.chain` ?

fhieber (changes requested):
Very useful feature, thanks for adding!

A few comments/suggestions:
- `source_prefix` should actually be a list of tokens, not a string
- I think we can simplify the code greatly by adding properties to the TranslatorInput dataclass.
- please add unit-test cases for TranslatorInput that cover `chunk`, `with_eos` methods 
- please add unit tests for computing max length

fhieber:
Regarding style checks: are you using the mypy & pylint versions defined in `requirements/requirements.txt`?

sockeye/inference.py
hoangcuong2011: I changed from `source_prefix` to `source_prefix_tokens` to make it align with the use of `tokens` variable. `get_tokens`() is also called here as you suggested. 
Once I added `get_source_prefix_tokens`() to `TranslatorInput` as property as you suggested, `split`() is avoided to be call several times indeed. 
Meanwhile IMO it is actually more intuitive to use `source_prefix_tokens` as a list[str] (i.e. `Tokens`), which is actually the same to tokens. Please take a look at the new revision and let me know if you still prefer a list of Tokens instead of Tokens.

sockeye/inference.py
hoangcuong2011: I also thought about that. However it is actually more convenient to not take into account of prefix length here. There are two reasons:
1. Please correct me: I think the length of input is used to compute the maximum length of translation output for the beam search (i.e. via the `get_max_input_output_length` function). Prefix tokens (e.g. <formal> <toGerman> <short>), however, should not be count as ``real" content tokens for that purpose in my opinion.
2. I found keeping the prefix length separately make it very convenient for chunking: We don't change anything regarding to the logic of chunking as we still chunk the input as is. What we need to change is simply the chunk size, which is the normal chunk size minus the prefix length.

Let me know if you have a different opinion.

sockeye/inference.py
hoangcuong2011: Now I use `Tokens`, which is a `List[Str]`. Yeah no need to do `split`() here and in other places.

sockeye/inference.py
hoangcuong2011: Thanks!

sockeye/inference.py
hoangcuong2011: Thanks

sockeye/inference.py
hoangcuong2011: Yeah I followed your suggestion in the revision version.

sockeye/inference.py
hoangcuong2011: Thanks

sockeye/inference.py
hoangcuong2011: I found using `intertools.chain` a bit complicate here because we need a merge of two separate List: `factor` and `source_prefix_factor` in an one by one mapping (for `intertools.chain` it is about concatenating the twos, which is not what we need here).

hoangcuong2011:
> Very useful feature, thanks for adding!
> 
> A few comments/suggestions:
> 
> * `source_prefix` should actually be a list of tokens, not a string
> * I think we can simplify the code greatly by adding properties to the TranslatorInput dataclass.
> * please add unit-test cases for TranslatorInput that cover `chunk`, `with_eos` methods
> * please add unit tests for computing max length



> Regarding style checks: are you using the mypy & pylint versions defined in `requirements/requirements.txt`?

Following your suggestion I recheck their versions and it works OK now.

hoangcuong2011:
> Very useful feature, thanks for adding!
> 
> A few comments/suggestions:
> 
> * `source_prefix` should actually be a list of tokens, not a string
> * I think we can simplify the code greatly by adding properties to the TranslatorInput dataclass.
> * please add unit-test cases for TranslatorInput that cover `chunk`, `with_eos` methods
> * please add unit tests for computing max length

I updated the code based on your suggestion. Please take a look. Meanwhile I have a question regarding adding unit-test cases with_eos method and unit tests for computing max length. Since I don't really change those parts, not really sure how to write unit tests for them properly. Please advise me on that.

sockeye/inference.py
fhieber: Yes you are right, what I mean was `Tokens` not `List[Tokens`. Thanks for the change!

sockeye/inference.py
fhieber: Makes sense, thanks for elaborating on it!

sockeye/inference.py
```diff
@@ -862,17 +912,22 @@ def _get_inference_input(self,
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)  # includes eos
             max_output_lengths.append(self._get_max_output_length(num_tokens))
```
fhieber, line 914: I am not sure yet about whether we really want to exclude the prefix tokens from the maximum output length calculation (which is based on the length ratio estimated on the training data). The reason is that if your model can expect special tokens at the beginning, these tokens are have likely also been added to the training data, effectively impacting the source length statistics of the model. 
So maybe we should simply change line 913 to `num_tokens = len(trans_input) + trans_input.num_source_prefix_tokens()`?

sockeye/inference.py
fhieber: Yes you are right

sockeye/inference.py
```diff
@@ -862,17 +912,22 @@ def _get_inference_input(self,
...
-
-                source[j, :num_tokens, i] = tokens2ids(factor, self.source_vocabs[i])[:num_tokens]
+            if not trans_input.source_prefix_factors: # no source prefix during inference
+                for i, factor in enumerate(factors[:self.num_source_factors - 1], start=1):
+                    # fill in as many factors as there are tokens
+                    source[j, :num_tokens, i] = tokens2ids(factor, self.source_vocabs[i])[:num_tokens]
+            else:
+                for i, zip_of_factor_and_prefix_factor in enumerate(zip(factors[:self.num_source_factors - 1], trans_input.source_prefix_factors[:self.num_source_factors - 1]), start=1):
+                    factor, source_prefix_factor = zip_of_factor_and_prefix_factor
+                    source[j, :num_tokens, i] = tokens2ids(source_prefix_factor + factor, self.source_vocabs[i])[:num_tokens]
```
fhieber, line 930: we could avoid the explicit construction of the concatenated list with `chain(source_prefix_factor, factor)`

sockeye/inference.py
```diff
@@ -862,17 +912,22 @@ def _get_inference_input(self,
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)  # includes eos
             max_output_lengths.append(self._get_max_output_length(num_tokens))
-            source[j, :num_tokens, 0] = tokens2ids(trans_input.tokens, self.source_vocabs[0])
-
+            num_tokens = num_tokens + trans_input.num_source_prefix_tokens()
+            source[j, :num_tokens, 0] = tokens2ids(trans_input.get_source_prefix_tokens() \
```
fhieber, line 916: we could avoid the explicit construction of the concatenated list with `chain(trans_input.get_source_prefix_tokens(), trans_input.tokens)`

fhieber (commented):
Thanks for the changes! Unit tests look fine to me!

Would you be able to document this feature in `https://github.com/awslabs/sockeye/blob/main/docs/inference.md#input-and-output-with-json`?

To get this ready, please also bump the minor version (`sockeye/__init__.py`) and create a corresponding entry in `CHANGELOG`.

sockeye/inference.py
mjdenkowski: One thing to think about is that prefixes are added differently during training and inference.

For training, users can add prefix tokens directly to the data because there's no chunking (sequences over max length are filtered out).  In this case, prefix tokens will be counted when computing length ratios.

During inference, the model treats each chunk as a separate sequence.  Including prefix tokens when computing the length of each chunk would match the training conditions.

hoangcuong2011: Thanks for suggestions. Since prefix tokens are counted when computing length ratios during training, let us keep it be counted at test time by adding prefix length here. Yeah I agreed.

sockeye/inference.py
hoangcuong2011: Yes I agreed we should take prefix tokens into account indeed.

sockeye/inference.py
hoangcuong2011: done!

sockeye/inference.py
```diff
@@ -149,6 +151,18 @@ def num_factors(self) -> int:
         """
         return 1 + (0 if not self.factors else len(self.factors))
 
+    def get_source_prefix_tokens(self) -> List[str]:
+        """
+        Returns the source prefix tokens of this instance.
+        """
+        return self.source_prefix_tokens if self.source_prefix_tokens is not None else []
+
+    def num_source_prefix_tokens(self) -> int:
```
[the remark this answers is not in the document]
hoangcuong2011: Not really sure about this. Could you elaborate a bit more?

hoangcuong2011:
> Thanks for the changes! Unit tests look fine to me!
> 
> Would you be able to document this feature in `https://github.com/awslabs/sockeye/blob/main/docs/inference.md#input-and-output-with-json`?
> 
> To get this ready, please also bump the minor version (`sockeye/__init__.py`) and create a corresponding entry in `CHANGELOG`.

I followed your suggestions and modified the code. Please take a look!

sockeye/inference.py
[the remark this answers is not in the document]
fhieber: Nevermind. I have a personal preference of using `@property` on such functions to avoid the parentheses in code, but it doesn't really matter here.

fhieber (approved):
Thanks for adding this!

PR #1028, opened 2022-03-11
Code formatting & cleanup for beam_search.py and inference.py

fhieber:
- reformatted a few lines to better adhere to max line length
- refactored `utils.gen_prefix_masking` to simplify logic in beam search
- rename numpy import from `onp` to `np`.

until you can check this box.

### DIFF
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -18,7 +18,7 @@ from abc import abstractmethod, ABC
 from dataclasses import dataclass
 from typing import Optional, Tuple, List, Union
 
-import numpy as onp
+import numpy as np
 import torch as pt
 
 import sockeye.constants as C
@@ -96,7 +96,7 @@ class _SingleModelInference(_Inference):
                     tf_logits = pt.log_softmax(tf_logits, dim=-1)
                 tf_scores = -tf_logits
                 if target_prefix_factor_mask is not None:
-                    tf_scores += target_prefix_factor_mask[:,:,i-1].reshape(-1, factor_vocab_size)
+                    tf_scores += target_prefix_factor_mask[:, :, i-1].reshape(-1, factor_vocab_size)
                 # target factors are greedily chosen, and score and index are collected via torch.min.
                 # Shape per factor: (batch*beam, 1, 2), where last dimension holds values and indices.
                 tf_prediction = pt.cat(tf_scores.min(dim=-1, keepdim=True), dim=1).unsqueeze(1)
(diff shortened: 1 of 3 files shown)


mjdenkowski (approved):

sockeye/beam_search.py
```diff
@@ -622,21 +624,24 @@ def forward(self,
...
         model_states, _ = self._inference.encode_and_initialize(source, source_length)
         # TODO: check for disabled predicted output length
 
         # Prefix masks, where scores are infinity for all other vocabulary items except target_prefix ids
-        prefix_masks, prefix_masks_length = (utils.gen_prefix_masking(target_prefix, self.output_vocab_size, self.dtype), target_prefix.size(1)) \
-                                            if target_prefix is not None else (None, None)  # type: ignore
-        if prefix_masks is not None and vocab_slice_ids is not None:
-            prefix_masks = pt.index_select(prefix_masks, -1, vocab_slice_ids)
+        prefix_masks, prefix_masks_length = None, 0
+        if prefix_masks is not None:
```
hoangcuong2011, line 636: Hi, something off here I think (prefix_masks is assigned to None in 635).

hoangcuong2011: Should be: if target_prefix is not None

sockeye/beam_search.py
```diff
@@ -622,21 +624,24 @@ def forward(self,
...
+            prefix_masks, prefix_masks_length = utils.gen_prefix_masking(target_prefix, self.output_vocab_size, self.dtype)
+            if vocab_slice_ids is not None:
+                prefix_masks = pt.index_select(prefix_masks, -1, vocab_slice_ids)
         # Prefix factor masks, where scores are also infinity for all other factor items except target_prefix_factor ids
-        target_prefix_factor_masks, target_prefix_factor_length = (utils.gen_prefix_masking(target_prefix_factors, self.output_factor_vocab_size, self.dtype), target_prefix_factors.size(1)) \
-                                                                  if self.num_target_factors > 1 and target_prefix_factors is not None \
-                                                                  else (None, None)  # type: ignore
+        target_prefix_factor_masks, target_prefix_factor_length = None, 0
+        if target_prefix_factors is not None:
```
hoangcuong2011, line 642: self.num_target_factors > 1 is used for if to handle the case where prefix factors are added (by mistake) but the trained model does not actually has factors. 

Perhaps if we ignore  self.num_target_factors > 1 here, it is worthy to check and assign target_prefix_factors to None if factors is not used in inference.py  in the first place? (somewhere in the function target_prefix_np)

hoangcuong2011 (commented):
Thank you Felix for making everything much cleaner!

sockeye/beam_search.py
fhieber: good cathch, thanks!

sockeye/beam_search.py
fhieber: Yes, you are right. I somehow thought we check for this already in `make_input_from_dict`. I changed the code to make sure we have the same number of target factor prefixes as there are model-supported target factors. I also removed the warning for an empty target factor because otherwise the user wouldn't be able to supply a prefix for the 3rd factor, but not for the 2nd factor.

PR #1033, opened 2022-03-23
Clarify use of Translator.batch_size in code

fhieber:
until you can check this box.
- [ ] Updated CHANGELOG.md


sockeye/inference.py
```diff
@@ -1200,7 +1200,7 @@ def _get_best_translations(self, result: SearchResult) -> List[Translation]:
         batch_size = best_hyp_indices.shape[0] // self.beam_size
         nbest_translations = []  # type: List[List[Translation]]
         reference_lengths = estimated_reference_lengths \
-            if estimated_reference_lengths is not None else np.zeros((self.batch_size * self.beam_size, 1))
+            if estimated_reference_lengths is not None else np.zeros((batch_size * self.beam_size, 1))
```
mjdenkowski, line 1203: Does `self.batch_size == batch_size` in all cases?

fhieber: Yes, that should be the case. This line is a fallback for cases where no estimated reference lengths are returned by the model (the default case). The zero tensor is then 'zipped' together with results from the previous beam search run from which `batch_size` is determined. Even if `self.batch_size` would be not equal to `batch_size` it would be strictly larger. In that case the zip would anyway exit early as it only zips until the shorted of its iterables are consumed.

sockeye/translate.py
```diff
@@ -218,9 +218,9 @@ def read_and_translate(translator: inference.Translator,
             chunk_size = C.CHUNK_SIZE_NO_BATCHING
         else:
             # Get a constant number of batches per call to Translator.translate.
-            chunk_size = C.CHUNK_SIZE_PER_BATCH_SEGMENT * translator.max_batch_size
+            chunk_size = C.CHUNK_SIZE_PER_BATCH_SEGMENT * batch_size
     else:
-        if chunk_size < translator.max_batch_size:
+        if chunk_size < batch_size:
```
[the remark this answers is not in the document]
fhieber: yes, saves one line

PR #1035, opened 2022-03-31
Optimized creation of repeated_state by using expand instead of repeat_interleave.

hoangcuong2011:
Replacing Pytorch's repeat_interleave function with a customized function repeat_interleave_with_expand reduces 48% latency on GPU (K80) and 41% latency on CPU on the creation of repeated_state. This could be worthy to consider given that `RepeatStates` is called numerous times during inference.

Comparison Code:
```
import timeit
set_up_line = """
import torch as pt
pt.manual_seed(1)
beam_size = 5
batch_size = 32
a = pt.rand(7, 2, 1024)
def repeat_interleave_with_expand(state, num_of_repeat, dim):
    state_size = list(state.size())
    expand_state_size = [-1 for _ in range(len(state_size) + 1)]
    state_size[dim] = state_size[dim] * num_of_repeat
    expand_state_size[dim +1] = num_of_repeat
    return state.unsqueeze(dim + 1).expand(expand_state_size).reshape(state_size)
"""
executive_line = """
a1 = repeat_interleave_with_expand(a, beam_size, 0)
a2 = repeat_interleave_with_expand(a, beam_size, 1)
"""
timeit.timeit(executive_line,setup=set_up_line, number=10000)
#gpu: 0.5507124290015781
#cpu: 0.5074962430226151

executive_line = """
a1 = a.repeat_interleave(beam_size, dim=0)
a2 = a.repeat_interleave(beam_size, dim=1)
"""
timeit.timeit(executive_line,setup=set_up_line, number=10000)
#gpu: 1.048011115999543
#cpu: 0.8592731970129535
```
<del>Side note: I still don't know yet why I have to change the code (decoder.py) from 
```source_mask_view = source_mask.view(batch * heads, target_max_len, source_max_len)```
to
```source_mask_view = source_mask.reshape(batch * heads, target_max_len, source_max_len)```
in order to pass `test/integration/test_backwards_compatibility.py` during running `pytest`.</del>

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -251,7 +251,7 @@ class TransformerDecoder(Decoder):
             autoregr_states = [list(islice(states_iter, 0, layer.num_state_tensors)) for layer in self.layers]  # type: ignore
 
         batch, heads, target_max_len, source_max_len = source_mask.size()
-        source_mask_view = source_mask.view(batch * heads, target_max_len, source_max_len)
+        source_mask_view = source_mask.reshape(batch * heads, target_max_len, source_max_len)
 
         # target: (batch_size, length, model_size)
         target = self.pos_embedding(step_input, steps)
--- a/sockeye/decoder.py
+++ b/sockeye/decoder.py
@@ -251,7 +251,7 @@ class TransformerDecoder(Decoder):
             autoregr_states = [list(islice(states_iter, 0, layer.num_state_tensors)) for layer in self.layers]  # type: ignore
 
         batch, heads, target_max_len, source_max_len = source_mask.size()
-        source_mask_view = source_mask.reshape(batch * heads, target_max_len, source_max_len)
+        source_mask_reshape = source_mask.reshape(batch * heads, target_max_len, source_max_len)
 
         # target: (batch_size, length, model_size)
         target = self.pos_embedding(step_input, steps)
@@ -266,7 +266,7 @@ class TransformerDecoder(Decoder):
             target, new_layer_autoregr_state = layer(target=target,
                                                      target_mask=target_mask,
                                                      source=source_encoded,
-                                                     source_mask=source_mask_view,
+                                                     source_mask=source_mask_reshape,
                                                      autoregr_states=layer_autoregr_state,
                                                      enc_att_kv=layer_enc_att_kv)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [3.1.8]
+
+### Changed
+
+- Optimized creation of repeated_state by using `expand` instead of `repeat_interleave`.
+
 ## [3.1.7]
 
 ### Changed
--- a/sockeye/__init__.py
+++ b/sockeye/__init__.py
@@ -11,4 +11,4 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '3.1.7'
+__version__ = '3.1.8'
--- a/sockeye/beam_search.py
+++ b/sockeye/beam_search.py
@@ -493,7 +493,7 @@ class RepeatStates(pt.nn.Module):
                 repeat_axis = 1
             else:
                 raise ValueError("Provided state format %s not recognized." % state_format)
-            repeated_state = state.repeat_interleave(repeats=self.beam_size, dim=repeat_axis)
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: sockeye/decoder.py [hunk 1]
```diff
@@ -251,7 +251,7 @@ def forward(self, step_input: pt.Tensor, states: List[pt.Tensor]) -> Tuple[pt.Te
             autoregr_states = [list(islice(states_iter, 0, layer.num_state_tensors)) for layer in self.layers]  # type: ignore
 
         batch, heads, target_max_len, source_max_len = source_mask.size()
-        source_mask_view = source_mask.view(batch * heads, target_max_len, source_max_len)
+        source_mask_view = source_mask.reshape(batch * heads, target_max_len, source_max_len)
```
mjdenkowski, line 254: Have you measured the effect of using `reshape()` instead of `view()` on CPU/GPU latency?

If we make this change, it would also make sense to rename the variable to `source_mask_reshape`.

mjdenkowski (commented):
Thanks for working on this!

sockeye/decoder.py
hoangcuong2011: Thanks Michael! I just changed the variable to `source_mask_reshape` as suggested.

Regarding speed, using reshape instead of view increases the latency to `4%` on GPU (K80) and `11%` on CPU.
Note that this reshaping is called only `1` time for each batch. Meanwhile repeat_interleave_with_expand is called a lot more than that (I noticed it is around `50` times for a batch I tried).
Code comparison:

```
import timeit
set_up_line = """
import torch as pt
pt.manual_seed(1)
heads = 8
batch = 20
target_max_len = source_max_len = 50
a = pt.rand(20, 8, 50, 50)
"""
executive_line = """
b = a.reshape(batch * heads, target_max_len, source_max_len)
"""
timeit.timeit(executive_line,setup=set_up_line, number=10000)
#cpu: 0.031618487962987274
#gpu: 0.013913217990193516

executive_line = """
b = a.view(batch * heads, target_max_len, source_max_len)
"""
timeit.timeit(executive_line,setup=set_up_line, number=10000)
#cpu: 0.028475572005845606
#gpu: 0.013381681987084448
```

hoangcuong2011:
Side note: 

`Source_mask` is one of the states (i.e. `steps, source_mask, *other = states` in `TransformerDecoder`). It is thus not possible to run `view` of `source_mask` when it is created by using `expand` instead of `repeat_interleave`. As a result we need to use `reshape` instead of `view` when reshaping this `source_mask` tensor in `TransformerDecoder` class (Luckily this is called only 1 time for each batch).

A way to make it possible to run `view` of `source_mask` is to return a clone of expanding tensor:

`return state.unsqueeze(dim + 1).expand(expand_state_size).clone().reshape(state_size)
`

instead of

`return state.unsqueeze(dim + 1).expand(expand_state_size).clone().reshape(state_size)
`

However, having additional `clone()` increases the latency of the customized function `repeat_interleave_with_expand` by 10% on GPU and 16% on CPU every time this function is called. Given that the customized function is called many more times (e.g. 4x, 5x) than reshaping a `source_mask` tensor, it is better to not include `clone()` in the customized function and use `reshape` instead of `view` of `source_mask` in the `TransformerDecoder` class.

hoangcuong2011:
Thanks Michael! The pull request is closed now do to a lack of evidence whether this improves latency for running `sockeye-translate` with full scale models. 

I still wonder why it does not improve latency because it shows improvement with examples I mentioned above. I suspected it is because of `jit trace`: with `jit trace` the difference between using two functions are minor regarding latency. 

To test this, I simplified the code even further with this [commit](https://github.com/hoangcuong2011/sockeye/commit/6b4dafd83cb0757131db9eb593d9ccd397a66969): https://github.com/hoangcuong2011/sockeye/commit/6b4dafd83cb0757131db9eb593d9ccd397a66969 With this commit, there is no need for `reshape` of `source_mask` in `TransformerDecoder`, as well as there is no need to have explicit additional lists in `repeat_interleave_with_expand` function, i.e. using:

```
state_size = state.size()
return state.unsqueeze(dim + 1).expand(*state_size[:dim + 1], repeats, *state_size[dim + 1:])\
           .clone().view(*state_size[:dim], state.size(dim) * repeats, *state_size[dim + 1:])
```

instead of

```
repeat_state_size = list(state.size())  # type: List[int]
repeat_state_size[dim] = repeat_state_size[dim] * repeats
expand_size = [-1 for _ in range(len(repeat_state_size) + 1)]  # type: List[int]
expand_size[dim + 1] = repeats
return state.unsqueeze(dim + 1).expand(expand_size).reshape(repeat_state_size)
```

@mjdenkowski : Could you later check whether with this commit, it has any latency improvements for running sockeye-translate with full scale models or not? That could be really helpful. Thanks!

PR #1048, opened 2022-05-27
Nbest reranking options using isometric translation criteria.

surafelml:
Added Nbest list reranking options using isometric translation criteria as proposed in an ICASSP 2021 paper https://arxiv.org/abs/2110.03847. 

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/rerank.py
```diff
@@ -41,36 +41,69 @@ class Reranker:
...
+        elif self.metric.startswith(C.RERANK_ISOMETRIC):
+            self.scoring_function = utils.compute_isometric_score  # type: ignore
         else:
             raise utils.SockeyeError("Scoring metric '%s' unknown. Choices are: %s" % (metric, C.RERANK_METRICS))
 
-        self.return_score = return_score
 
-    def rerank(self, hypotheses: Dict[str, Any], reference: str) -> Dict[str, Any]:
+    def rerank(self, hypotheses: Dict[str, Any], reference: str, source: str,
+               isometric_alpha: float = 0.5) -> Dict[str, Any]:
```
fhieber, line 62: would it make sense to move this (optinal) alpha parameter into the constructor of `Reranker` to keep the signature of the `rerank()` method the same?

sockeye/rerank.py
```diff
@@ -41,36 +41,69 @@ class Reranker:
...
+
+        if self.metric == C.RERANK_BLEU or self.metric == C.RERANK_CHRF:
+            scores = [self.scoring_function(hypothesis, [reference]).score for
+                      hypothesis in hypotheses['translations']]
+            # BLEU,CHRF - the higher, the better
+            ranking = self._get_ranking_indices(scores, kind='mergesort', descending=True)
+
+        if self.metric.startswith(C.RERANK_ISOMETRIC):
+            scores = [self.scoring_function(hypothesis, hypothesis_score[0], source,
+                                            self.metric, isometric_alpha)
```
fhieber, line 82: if you add `isometric_alpha` to the constructor, you could create a partial `scoring_function` with constant arguments, alpha & metric, in the constructor, and only pass hypothesis, score, source here.

sockeye/rerank.py
```diff
@@ -41,36 +41,69 @@ class Reranker:
...
+                      for hypothesis, hypothesis_score in
+                      zip(hypotheses['translations'], hypotheses['scores'])]
+
+            if self.metric == C.RERANK_ISOMETRIC_LC:
+                # isometric-lc - the smaller, the better
+                ranking = self._get_ranking_indices(scores, kind='mergesort', descending=False)
```
fhieber, line 88: same here, you could create a `partial` to bake in the sort order already in the constructor

sockeye/rerank.py
```diff
@@ -102,18 +136,36 @@ def rerank(args: argparse.Namespace):
...
+            source = hypotheses['text']
 
             if not num_hypotheses > 1:
                 logger.info("Line %d contains %d hypotheses. Nothing to rerank.", i, num_hypotheses)
                 reranked_hypotheses = hypotheses
             else:
-                reranked_hypotheses = reranker.rerank(hypotheses, reference)
+                reranked_hypotheses = reranker.rerank(hypotheses,
+                                                          reference,
+                                                          source,
```
fhieber, line 147: if `source` is a field in `hypotheses`, why change the signature of this function?

sockeye/rerank.py
```diff
@@ -102,18 +136,36 @@ def rerank(args: argparse.Namespace):
...
                 if not best_hypothesis and args.output_reference_instead_of_blank:
                     logger.warning('Line %d: replacing blank hypothesis with reference.', i)
                     best_hypothesis = reference
+
+                # get best non-blank hypothesis, when reference is not used
+                if not best_hypothesis and args.output_best_non_blank and num_hypotheses > 1:
+                    for h in range(num_hypotheses):
+                        best_hypothesis = reranked_hypotheses['translations'][h]
+
+                        if best_hypothesis == '':
```
fhieber, line 162: `if not best_hypothesis`?

sockeye/rerank.py
```diff
@@ -141,4 +194,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
fhieber, line 197: missing newline

sockeye/utils.py
```diff
@@ -615,3 +615,52 @@ def all_gather_object(obj: T) -> List[T]:
     obj_list = [None] * torch.distributed.get_world_size()  # type: List[T]
     torch.distributed.all_gather_object(obj_list, obj)
     return obj_list
+
+
+def count_seq_len(sample: str, count_type: str = 'char', replace_tokens: list = []) -> int:
```
fhieber, line 620: `list` -> `List`

sockeye/utils.py
```diff
@@ -615,3 +615,52 @@ def all_gather_object(obj: T) -> List[T]:
...
+    """Count sequence length (char, tokens) with optional replace tokens.
```
fhieber, line 621: [nit] I would format as:
```
"""
[...]
"""
```

sockeye/utils.py
```diff
@@ -615,3 +615,52 @@ def all_gather_object(obj: T) -> List[T]:
...
+    If count_type not char, count tokens."""
+    if len(replace_tokens) >= 1:
+        for r in replace_tokens:
+            sample = sample.replace(r, '')
+    if count_type != 'char':
```
fhieber, line 626: this is confusing to me. If the type is not `char`, return the number of characters?

sockeye/utils.py
```diff
@@ -615,3 +615,52 @@ def all_gather_object(obj: T) -> List[T]:
...
+    """
+    count_type = 'char' # constant for isometric scoring
+    replace_tokens = C.TOKEN_SEGMENTATION_MARKERS
+
+    hypothesis_len = count_seq_len(hypothesis, count_type, replace_tokens)
+    source_len = count_seq_len(source, count_type, replace_tokens)
+
+    if isometric_metric == C.RERANK_ISOMETRIC_LC:
+        abs_len_diff = abs(hypothesis_len - source_len)
+        isometric_score = (abs_len_diff*100) / source_len
```
fhieber, line 649: can `source_len` be 0 ?

sockeye/rerank.py
surafelml: thanks! i have made the update according to your suggestion.

sockeye/rerank.py
surafelml: only isometric metric (isometric-lc) criterion requires a different ordering, the partial func is moved to the constructor as if..else.

sockeye/utils.py
surafelml: thanks for catching this, updated to handle both character and token counts.

sockeye/utils.py
surafelml: updated to handle zero division error

sockeye/rerank.py
```diff
@@ -39,21 +39,32 @@ class Reranker:
 
     :param metric: Sentence-level metric such as smoothed BLEU.
     :param return_score: If True, also return the sentence-level score.
+    :param isometric_alpha: Factor for reranking with isometric criteria.
     """
 
-    def __init__(self, metric: str,
-                 return_score: bool = False) -> None:
-        if metric == C.RERANK_BLEU:
+    def __init__(self, metric: str, isometric_alpha: float, return_score: bool = False) -> None:
```
fhieber, line 45: you probably want to make isometric_alpha optional and the last argument? In Line 124 you pass the alpha as the last argument.

sockeye/rerank.py
```diff
@@ -90,8 +121,9 @@ def rerank(args: argparse.Namespace):
 
     :param args: Namespace object holding CLI arguments.
     """
-    reranker = Reranker(args.metric, args.return_score)
+    reranker = Reranker(args.metric, args.return_score, args.isometric_alpha)
```
fhieber, line 124: argument order seems swapped

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
     obj_list = [None] * torch.distributed.get_world_size()  # type: List[T]
     torch.distributed.all_gather_object(obj_list, obj)
     return obj_list
+
+
+def count_seq_len(sample: str, count_type: str = 'char', replace_tokens: List = []) -> int:
```
fhieber, line 620: I think mutable types as default arguments is discouraged. Maybe make it `Optional[List]` and set the default to `None`.

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
...
+    """
+    Count sequence length, after replacing word segmentation token/s.
+    """
+    if len(replace_tokens) >= 1:
```
fhieber, line 624: Then you can change this to `if replace_tokens is not None`

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
...
+        for r in replace_tokens:
+            sample = sample.replace(r, '')
+    if count_type == 'char':
+        return len(sample.replace(' ', ''))
+    else:
+        return len(sample.split(' '))
```
fhieber, line 630: there is a Sockeye-wide constant called TOKEN_SEPARATOR that maps to ' ' which you could use here.

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
...
+
+def count_seq_len(sample: str, count_type: str = 'char', replace_tokens: List = []) -> int:
+    """
+    Count sequence length, after replacing word segmentation token/s.
+    """
+    if len(replace_tokens) >= 1:
+        for r in replace_tokens:
+            sample = sample.replace(r, '')
+    if count_type == 'char':
+        return len(sample.replace(' ', ''))
```
fhieber, line 628: same as comment below

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
...
+                            isometric_metric: str = 'isometric-ratio', isometric_alpha: float = 0.5) -> float:
+    """
+    Compute hypothesis to source isometric score using sample char length
+    and isometric metric (ratio, diff, lc).
+        - isometric-diff: https://aclanthology.org/W19-5210.pdf
+        - isometric-ratio/lc: https://arxiv.org/pdf/2110.03847.pdf
+    :return isometric score
+    """
+    count_type = 'char'  # for isometric scoring, count length in characters
+    replace_tokens = C.TOKEN_SEGMENTATION_MARKERS
```
fhieber, line 643: if this is where you hardcode the segmentation markers, why not make the replace_tokens argument in `count_seq_len` default directly to C.TOKEN_SEGMENTATION_MARKERS?

surafelml: this will make `count_seq_len` less flexible, if called from another func to handle different tokens for replacement. i added Sockeye wide constants `SEQ_LEN_IN_*` to avoid hard coding `char`, and assuming common sequence length counting types (char, tokens, and words).

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
     obj_list = [None] * torch.distributed.get_world_size()  # type: List[T]
     torch.distributed.all_gather_object(obj_list, obj)
     return obj_list
+
+
+def count_seq_len(sample: str, count_type: str = 'char', replace_tokens: Optional[List] = None) -> int:
```
fhieber, line 620: I think some unit tests would be useful for this.

sockeye/utils.py
```diff
@@ -615,3 +615,53 @@ def all_gather_object(obj: T) -> List[T]:
...
+    if replace_tokens is not None:
+        for tokens in replace_tokens:
+            sample = sample.replace(tokens, '')
+    if count_type == C.SEQ_LEN_IN_CHARACTERS:
+        return len(sample.replace(C.TOKEN_SEPARATOR, ''))
+    if count_type == C.SEQ_LEN_IN_TOKENS:
+        return len(sample.split(C.TOKEN_SEPARATOR))
+
+
+def compute_isometric_score(hypothesis: str, hypothesis_score: float, source: str,
```
fhieber, line 633: I think some unit tests would be useful for this.

fhieber (commented):
Thanks for iterating, just 2 remaining comments.

surafelml:
> Thanks for iterating, just 2 remaining comments.

I have added the tests. Thanks for the reviews and comments!

fhieber (approved):
Looks good!
Only thing remaining is to resolve the conflict with main in CHANGELOG.md

surafelml:
style-check seems to throw an error on the inference script, which is not related with reranking module
`sockeye/inference.py:1215: error: "_assemble_translation" of "Translator" gets multiple values for keyword argument "unshift_target_factors"
Found 1 error in 1 file (checked 31 source files)`

surafelml:
i have checked it several times, after downgrading `pylint 2.13.9` to the gh CI version `pylint 2.6.0` i keep getting the same error.

fhieber:
The failure in CI is not related to the issues you see locally. Sometimes it is difficult to replicate the environment of CI exactly locally. 
The problem seems to be the linter mixing the argument structure of the case then `scoring_function` is chrf, and when it is your partial of `compute_isometric_score`, because they have different signatures. If you want just add an ignore to it.

fhieber:
`# type: ignore` afaik only works for mypi. If you want to ignore a specific pytlint warning you should use `# pylint: disable=<error/warning type>`.
The last commit seems to introduce another error.

surafelml:
local tests didn't catch mypy (same version) error as in CI for Literal. first option was to use set `kind: Literal['quicksort', 'mergesort', 'heapsort', 'stable']`, this seems to require `py3.8`, instead i have set type ignore for argsort.

PR #1057, opened 2022-08-07
[Review #1051 First] Refactoring Ahead of Adding DeepSpeed Support

mjdenkowski:
This PR includes #1051. Reviewing/merging it first is recommended.

This PR is the first stage of adding [DeepSpeed](https://github.com/microsoft/DeepSpeed) support to Sockeye's `main` branch. It ports several changes from the `deepspeed` branch to improve code modularity and compatibility with PyTorch/DeepSpeed APIs. No DeepSpeed support is added yet.

CLI-level behavior is unchanged except for removing one unused option. Training/inference outputs are identical before and after the changes.

Changes include:
- Cleaning up GPU and CPU memory used during training initialization before starting the main training loop.
- Moving logic for flagging interleaved key-value parameters from layers.py to model.py.
- Refactoring the LearningRateScheduler API to be compatible with PyTorch/DeepSpeed scheduler APIs.
- Refactoring the optimizer and learning rate scheduler creation code to improve modularity.
- Introducing the ModelWithLoss object, which wraps a Sockeye model and its losses in a single callable module.
- Refactoring primary and secondary worker logic to reduce redundant calculations.
- Refactoring code for saving/loading training states.
- Adding utility code for managing model/training configurations.
- Removing unused training option `--learning-rate-t-scale`.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/arguments.py
```diff
@@ -206,18 +208,20 @@ def _parse(value: str):
                 return True
             if value.lower() == "false":
                 return False
-            if "." in value or "e" in value:
+            if value.isdigit():
```
tdomhan, line 211: I guess we haven't needed nor tested this, but this wouldn't work for negative numbers, right?

tdomhan:
Thanks Michael! I approved #1051.

sockeye/model.py
```diff
@@ -422,9 +426,12 @@ def load_parameters(self,
             utils.check_condition(not missing, f"missing keys: {missing}")
         if not ignore_extra:
             utils.check_condition(not unexpected, f"extra keys: {unexpected}")
-        # Models are saved with interleaved key-value params. If the current
-        # model is in training mode, separate the loaded params to match the
-        # format used during training.
+        # Models are saved with interleaved key-value params.
+        for module in self.modules():
+            if hasattr(module, 'kv_interleaved') and isinstance(module.kv_interleaved, bool):
```
tdomhan, line 431: What is the check for bool for? Would we expect this member to ever be not bool?

sockeye/lr_scheduler.py
```diff
@@ -13,7 +13,10 @@
 
 import logging
 from math import sqrt
-from typing import Optional
+from typing import Any, Dict, List, Optional, Tuple, Type
+
+import torch
```
tdomhan, line 18: Elsewhere we import as `pt`, while non-standard, we could maybe apply the same import of `pt` here.

sockeye/train.py
```diff
@@ -1008,7 +1001,18 @@ def train(args: argparse.Namespace, custom_metrics_logger: Optional[Callable] =
...
+    optimizer_class, optimizer_kwargs, zero_grad_kwargs = optimizers.get_optimizer(optimizer_config)
+    optimizer = optimizer_class(sockeye_model.parameters(), **optimizer_kwargs)
+
+    lr_scheduler_class, lr_scheduler_kwargs = lr_scheduler.get_lr_scheduler(args.learning_rate_scheduler_type,
+                                                                            args.initial_learning_rate,
+                                                                            args.learning_rate_reduce_factor,
+                                                                            args.learning_rate_reduce_num_not_improved,
+                                                                            args.learning_rate_warmup,
+                                                                            args.max_updates)
+    _lr_scheduler = lr_scheduler_class(optimizer, **lr_scheduler_kwargs) if lr_scheduler_class is not None else None
```
tdomhan, line 1015: What's the reason to split this into the class and the kwargs rather than passing the optimizer to `get_lr_scheduler`?
(is this related to upcoming changes for DeepSpeed?)

tdomhan (commented):
Thanks Michael. Just left some small comments. Will take a closer look later.

sockeye/arguments.py
mjdenkowski: Good catch!

sockeye/model.py
mjdenkowski: Right now, `kv_interleaved` is always `bool` if present. The check is for extra type safety and forward compatibility.

sockeye/lr_scheduler.py
mjdenkowski: This is an open question. Right now some parts of our code use `torch` and others use `pt`. Do we have a compelling reason to use `pt` instead of the standard `torch`?

sockeye/train.py
mjdenkowski: Yes. The idea is that our functions for creating the optimizer and learning rate scheduler return a class and the kwargs for instantiating that class. We can instantiate the class directly or pass the class/kwargs to DeepSpeed. For example, DeepSpeed matches the optimizer name and instantiates its own implementation (DeepSpeed Adam instead of PyTorch Adam) using the same kwargs: https://github.com/awslabs/sockeye/blob/deepspeed/sockeye/train.py#L822-L824

sockeye/arguments.py
mjdenkowski: Commit 9f77fb6 adds support for negative integers.

sockeye/lr_scheduler.py
fhieber: When we wrote the initial PyTorch version we preferred `pt` over `torch` because its shorter, requiring less line breaks, and follows the 'convention' of abbreviating math/ml libraries with 2 letters (mxnet -> mx, numpy -> np). I'd be in favor of using `pt`.

mjdenkowski: I'm generally in favor of using libraries' chosen names. In their own documentation, [NumPy prefers `np`](https://numpy.org/doc/stable/user/quickstart.html) while [PyTorch prefers `torch`](https://pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html). This also avoids the need to change all instances of `torch` to `pt` during development (for example, when pasting snippets from both Sockeye and the PyTorch docs into the Python interpreter for interactive testing).

I agree with you about finding good ways to shorten lines. When we're using a line width of 120 characters, I'm not sure that +/- 3 makes a big difference.

fhieber: Depending on the statement, the difference might be more than 3 characters, but I see your point about pasting code snippets from PyTorch documentation.
If we want to make the change from `pt` to `torch`, we should do it consistently throughout the codebase. Maybe its worth doing this in one step in another PR?

tdomhan: I don't have a strong preference either way.  I agree that we should do a pass to do this consistently in the codebase.

mjdenkowski: I agree with changing everything over in a single, separate PR.

This PR uses the existing name in each source file (`pt` or `torch`). For the special case where the import needs to be added for `lr_scheduler.py`, it uses `torch` to match the training code that calls it.

sockeye/lr_scheduler.py
```diff
@@ -212,21 +254,17 @@ def get_lr_scheduler(scheduler_type: str,
...
+        return LearningRateSchedulerInvSqrtDecay, {'base_lr': base_learning_rate, 'warmup': learning_rate_warmup}
     if scheduler_type == C.LR_SCHEDULER_LINEAR_DECAY:
-        check_condition(max_updates is not None,
-                        "The total number of training updates (--max-updates) must be specified when using the linear "
-                        "decay learning rate scheduler.")
-        return LearningRateSchedulerLinearDecay(base_lr=base_learning_rate,
-                                                total_steps=max_updates,
-                                                warmup=learning_rate_warmup,
-                                                t_scale=learning_rate_t_scale)
+        assert max_updates is not None, "The total number of training updates (--max-updates) must be specified when " \
```
fhieber, line 264: why not keep using check_condition (which returns a nicer looking error than `assert`)?

sockeye/model.py
fhieber: this is a nice refactor, thanks

sockeye/training.py
```diff
@@ -316,26 +345,26 @@ def _forward_backward(self, batch: data_io.Batch, is_update_batch: bool = True):
         """
         batch = batch.load(device=self.device)
         with torch.cuda.amp.autocast(cache_enabled=False) if self.using_amp else utils.no_context():  # type: ignore
-            # Forward
-            outputs = self.training_model(batch.source, batch.source_length, batch.target, batch.target_length)
-            # Loss (scaled by update interval)
-            loss_outputs = [loss_function(outputs, batch.labels) for loss_function in self.loss_functions]
-            # TODO(mdenkows): We currently give 1/N weight to every batch in the
```
fhieber, line 323: Do we still need this comment? It is not part of the ModelWithLoss object.

sockeye/training.py
```diff
@@ -401,30 +427,29 @@ def _evaluate(self, checkpoint: int, data_iter,
         for batch in data_iter:
             batch = batch.load(device=self.device)
             with torch.inference_mode():
-                # Forward: use sockeye_model because (traced) training_model
```
fhieber, line 404: that comment might still be valuable to have. A user might wonder why not use `model_object.model` here instead.

More generally: would it make sense to wrap all of the different model instances and losses into one object, i.e. 
```
ModelWithLoss, containing
- model (traced)
- model (untraced)
- losses
```

sockeye/training.py
```diff
@@ -557,7 +569,8 @@ def _adjust_learning_rate(self, has_improved: bool):
                     self.sockeye_model.load_parameters(filename=self.best_params_fname, device=self.device)
                 if os.path.exists(self.best_optimizer_state_fname):
                     self._load_optimizer_state(self.best_optimizer_state_fname)
-            lr = scheduler.lr
+            # Assume same learning rate for all param groups
```
fhieber, line 572: Do you document somewhere (maybe in lr_scheduler.py) what parameter groups are?

sockeye/lr_scheduler.py
mjdenkowski: Fixed

sockeye/training.py
mjdenkowski: Added back a version of the comment

sockeye/training.py
mjdenkowski: Added a note to the docstring for `LearningRateScheduler` on where to find information about PyTorch's optimizer/scheduler APIs.

sockeye/training.py
mjdenkowski: This gets more complicated when we add DeepSpeed support, which automatically handles saving/loading the model object's state. If we have a regular model and a traced model, it tries to load parameters for both. The traced parameters don't exist, so we need additional logic to filter them out. With a single model instance, the `ModelWithLoss` object state can saved/loaded as a standard Pytorch module.

This also helps us minimize complexity. We have a `ModelWithLoss` object that turns input batches into loss outputs using the fully traced/wrapped model. We also have the "vanilla" `SockeyeModel` for saving parameters, running the checkpoint decoder, etc. This way the object with added complexity (`ModelWithLoss`) only needs to support a single use case.

I've added back a version of the comment about why we use the `SockeyeModel` here.

mjdenkowski:
Thanks Tobi and Felix! I've made some improvements based on your feedback.

sockeye/utils.py
```diff
@@ -599,6 +599,37 @@ def create_pool(max_processes):
...
+    for key in source:
+        if isinstance(source[key], dict):
+            if key not in dest:
+                dest[key] = {}
+            if not isinstance(dest[key], dict):
+                raise ValueError(f'Type mismatch for key {key}: {type(source[key])} vs {type(dest[key])}')
+            update_dict(dest[key], source[key])
+        else:
+            dest[key] = source[key]
+
```
fhieber, line 612: Small nitpick: pep8 expects 2 newlines here

fhieber (approved):
LGTM

sockeye/utils.py
mjdenkowski: My mistake. Fixed.

PR #1059, opened 2022-08-23
Add Support for DeepSpeed ZeRO Stage 1

mjdenkowski:
This PR adds training support for [DeepSpeed](https://github.com/microsoft/DeepSpeed) with [ZeRO stage 1](https://arxiv.org/abs/1910.02054v3):

- Install DeepSpeed: `pip install deepspeed`
- Launch training with DeepSpeed: `deepspeed --no_python ... sockeye-train ...`
- Optionally specify FP16 or BF16 mode with `--deepspeed-fp16` or `--deepspeed-bf16`
- DeepSpeed automatically passes the `--local_rank` option to Sockeye, which activates DeepSpeed mode.

Sockeye's DeepSpeed mode does the following:
- Initialize DeepSpeed with ZeRO stage 1 (optimizer state sharding).
- Skip model tracing because DeepSpeed makes runtime modifications to the model that may not be JIT-compatible.
- Guarantee model outputs are in FP32 before computing losses.
- Have DeepSpeed automatically handle several aspects of training: moving the model to the correct device, accumulating gradients across multiple updates, synchronizing states for model updates, and stepping the learning rate scheduler and optimizer.
- Use DeepSpeed to save/load model parameters, optimizer states, and learning rate scheduler states.

DeepSpeed saves checkpoints using its own format. The new `convert_deepspeed` module recovers standard Sockeye parameter files from these checkpoints. At the end of training, checkpoints are automatically converted. Checkpoints can also be converted manually using the CLI.

DeepSpeed is an optional dependency (like Apex) and these changes are backward compatible. Training results with DeepSpeed ZeRO stage 1 are identical for this PR and the `deepspeed` branch. Training results for standard FP32 mode, PyTorch AMP, and Apex AMP are identical for this PR and the `main` branch.

until you can check this box.
- [x] Updated CHANGELOG.md


sockeye/arguments.py
```diff
@@ -1043,6 +1043,24 @@ def add_training_args(params):
...
                               help="Manually specify names of parameters to fix during training. Default: %(default)s.")
 
+    # DeepSpeed arguments
+    train_params.add_argument('--local_rank',
+                               type=int_greater_or_equal(0),
+                               default=None,
+                               help='The DeepSpeed launcher (`deepspeed`) automatically adds this argument. When it is '
+                                    'present, training runs in DeepSpeed mode. This argument does not need to be '
+                                    'specified manually.')
+    train_params.add_argument('--deepspeed-fp16',
```
fhieber, line 1053: are `--deepspeed-fp16` and `--deepspeed-bf16` mutually exclusive arguments? Would it make sense instead something like `--deepspeed-dtype {fp16,bf16}` to limit the number of arguments added to CLIs ?

sockeye/utils.py
```diff
@@ -667,6 +675,31 @@ def all_gather_object(obj: T) -> List[T]:
...
+    """
+    Make sure all of the DeepSpeed modules we use can be imported, initialize
+    DeepSpeed, and set the global variable that tracks initialization.
+
+    """
+    global _using_deepspeed
+    try:
+        import deepspeed  # pylint: disable=E0401
+        import deepspeed.utils.zero_to_fp32  # pylint: disable=E0401
+        deepspeed.init_distributed()
```
fhieber, line 691: Would it make sense to add a logging statement here?

sockeye/convert_deepspeed.py
```diff
@@ -0,0 +1,88 @@
...
+def convert_model_checkpoints(model_dirname: str, keep_deepspeed: bool = False):
+    model_config_fname = os.path.join(model_dirname, C.CONFIG_NAME)
+    # Find and convert params.00000, etc.
+    for fname in os.listdir(model_dirname):
+        if fname.startswith(C.PARAMS_PREFIX) and fname[len(C.PARAMS_PREFIX):].isdigit():
+            params_fname = os.path.join(model_dirname, fname)
+            if os.path.isdir(params_fname):
+                # Move directory checkpoint to e.g., params.00000.ds
+                checkpoint_dirname = params_fname + '.ds'
+                shutil.move(params_fname, checkpoint_dirname)
```
fhieber, line 61: it could be useful to announce actions taken with a logging statement. Otherwise `convert_model_checkpoints` may silently return without action.

fhieber (commented):
Great addition, thanks!

I am wondering if there is a good way to test some of the added functionality in Github actions. I guess its difficult to install deepspeed in the workflows. Would it be possible to test the checkpoint conversion module somehow?

sockeye/arguments.py
mjdenkowski: I agree that a single option would be easier to work with. Right now we mirror the option names in the [DeepSpeed config](https://www.deepspeed.ai/docs/config-json/#fp16-training-options) which uses separate sections for `fp16` and `bf16`. Enabling both is a valid DeepSpeed config, although DeepSpeed may raise its own error about it.

sockeye/utils.py
mjdenkowski: We get free log messages here from [`init_distributed()`](https://github.com/microsoft/DeepSpeed/blob/b288cf1b9b90030cc2025f2681e77f3bc81dd409/deepspeed/comm/comm.py#L577).

sockeye/convert_deepspeed.py
mjdenkowski: Added logging in a new commit.

mjdenkowski:
> Great addition, thanks!
> 
> I am wondering if there is a good way to test some of the added functionality in Github actions. I guess its difficult to install deepspeed in the workflows. Would it be possible to test the checkpoint conversion module somehow?

That's a good idea. It's now possible to `pip install deepspeed` on CPU-only machines. Training won't run (GPUs required), but some utility functions will.

I've added a test that runs checkpoint conversion with existing model files if the `deepspeed` package is installed. I've also updated our workflows to install `deepspeed`, which should be compatible with all of our environments. We can see if anything breaks.

mjdenkowski:
Our automatic tests now cover installing `deepspeed` and converting a DeepSpeed checkpoint.

fhieber (approved):
Looks good, thanks for adding the test to the workflows!

PR #1062, opened 2022-08-27
Add kNN-MT

shuoyangd:
This is the refactored implementation from the MTMA-2022 [kNN-based Retrieval Module for Sockeye](https://docs.google.com/document/d/1_Lea0E4g-VyqiRTqVbMXfA9iLTcJyQNVQDdkiKosoJ4/edit#heading=h.656mkzwobicc) project. It re-implements the models as in [Khandelwal et al. 2021](https://arxiv.org/abs/2010.00710).

until you can check this box.
- [x] Updated CHANGELOG.md


mjdenkowski (commented):
Thanks @shuoyangd, @zhaogao-git, @jgwinnup, and @tdomhan for implementing this!

A few high level comments:

#### Installation

Can we make Faiss an optional dependency like Apex and DeepSpeed? Including `faiss-gpu` in requirements.txt means that any time a user installs Sockeye, Pip will first attempt to install the GPU version of Faiss from PyPI. If the package/version doesn't exist for the user's platform, Sockeye cannot be installed.

To make Faiss optional, we can wrap `faiss` imports with try/except blocks and check that the modules we need are available before running any Faiss code (example of [importing](https://github.com/awslabs/sockeye/blob/969049898d6f80b08a5c9eaba2cae036a3da30cc/sockeye/utils.py#L36-L42) and [testing](https://github.com/awslabs/sockeye/blob/969049898d6f80b08a5c9eaba2cae036a3da30cc/sockeye/utils.py#L681-L695) for `deepspeed`). This enables users to run the non-kNN-MT parts of Sockeye without Faiss. When users want to run kNN-MT, they can follow the instructions to install the right version of Faiss.


#### Naming

Can we find another verb to replace "dump" when describing how we handle decoder states? We could consider "generate", "output", "record", or something else.

#### Testing

Can we add unit tests for creating/using Faiss indexes and running individual components of kNN-MT? We can configure any tests that depend on `faiss` to only run when the module is available. We can also configure our GitHub workflows to install `faiss-cpu` so those tests run automatically for PRs. (Example of [installing](https://github.com/awslabs/sockeye/blob/969049898d6f80b08a5c9eaba2cae036a3da30cc/.github/workflows/push_pr.yml#L33-L34) `deepspeed` in a workflow and [testing](https://github.com/awslabs/sockeye/blob/969049898d6f80b08a5c9eaba2cae036a3da30cc/test/unit/test_deepspeed.py) DeepSpeed code)

sockeye/beam_search.py
```diff
@@ -83,10 +85,18 @@ def decode_step(self,
...
                     factor_vocab_size: Optional[int] = None):
-        logits, states, target_factor_outputs = self._model.decode_step(step_input, states, vocab_slice_ids)
+        logits, knn_probs, states, target_factor_outputs = self._model.decode_step(step_input, states, vocab_slice_ids)
         if not self._skip_softmax:
-            logits = pt.log_softmax(logits, dim=-1)
-        scores = -logits  # shape: (batch*beam, output_vocab_size/len(vocab_slice_ids))
+            if knn_probs is None:  # no knn used
+                probs = pt.log_softmax(logits, dim=-1)
+            else:
+                lmbda = self.knn_lambda
```
fhieber, line 93: this assignment is not necessary

sockeye/beam_search.py
```diff
@@ -163,7 +173,7 @@ def decode_step(self,
         for model, model_state_structure in zip(self._models, self.state_structure()):
             model_states = states[state_index:state_index+len(model_state_structure)]
             state_index += len(model_state_structure)
-            logits, model_states, target_factor_outputs = model.decode_step(step_input, model_states, vocab_slice_ids)
+            logits, knn_outputs, model_states, target_factor_outputs = model.decode_step(step_input, model_states, vocab_slice_ids)
```
fhieber, line 176: if ensembling does not yet support KNN, it should throw/warn.

sockeye/dump_decoder_states.py
```diff
@@ -0,0 +1,240 @@
+# Copyright 2018--2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fhieber, line 1: File was added in 2022, so no range needed that starts in 2018

sockeye/layers.py
```diff
@@ -117,6 +119,49 @@ def forward(self, data: pt.Tensor, vocab_slice_ids: Optional[pt.Tensor] = None)
         return F.linear(data, weight, bias)
 
 
+class KNN(pt.nn.Module):
+    def __init__(self, keys_index, vals, vocab_size, k=64, temperature=10, state_dump=None) -> None:
```
fhieber, line 123: can we add type annotations for the first 3 positional arguments?

sockeye/model.py
```diff
@@ -298,13 +302,21 @@ def decode_step(self,
...
         # the traced module returns a flat list of tensors
         decode_step_outputs = self.traced_decode_step(*decode_step_inputs)
-        step_output, *target_factor_outputs = decode_step_outputs[:self.num_target_factors]
-        new_states = decode_step_outputs[self.num_target_factors:]
-        return step_output, new_states, target_factor_outputs
+        # +1 for the decoder output, which will be used to generate kNN output
+        step_output, decoder_out, *target_factor_outputs = decode_step_outputs[:self.num_target_factors+1]
+
+        # do the query here because it cannot be traced (jit.ignore does not play well with tracing)
+        knn_output = None
```
fhieber, line 314: these 3 lines can be turned into one:
`knn_output = self.knn(decoder_out) if self.knn is not None else None`

sockeye/model.py
```diff
@@ -278,7 +282,7 @@ def _embed_and_encode(self,
     def decode_step(self,
                     step_input: pt.Tensor,
                     states: List[pt.Tensor],
-                    vocab_slice_ids: Optional[pt.Tensor] = None) -> Tuple[pt.Tensor, List[pt.Tensor], List[pt.Tensor]]:
+                    vocab_slice_ids: Optional[pt.Tensor] = None) -> Tuple[pt.Tensor, pt.Tensor, List[pt.Tensor], List[pt.Tensor]]:
```
fhieber, line 285: docstring needs update

sockeye/translate.py
```diff
@@ -72,12 +72,20 @@ def run_translate(args: argparse.Namespace):
         use_cpu = True
     device = pt.device('cpu') if use_cpu else pt.device('cuda', args.device_id)
     logger.info(f"Translate Device: {device}")
+    knn_index = args.knn_index
```
fhieber, line 75: why not pass the args members directly?

fhieber (commented):
Thanks for adding! A did a first pass but will need more time to look at this in depth.

shuoyangd:
Thanks @mjdenkowski and @fhieber for the initial comments!

re: Installation -- Great idea! I'll update the code to reflect that.

re: Naming -- Maybe let's do "store"? This establishes some connection with the original paper's notion of "create a datastore" (but IMO "datastore" is too vague of a concept for people who are not familiar with that paper).

re: Testing -- I'll look into this and add some tests to the PR.

mjdenkowski:
> re: Naming -- Maybe let's do "store"? This establishes some connection with the original paper's notion of "create a datastore" (but IMO "datastore" is too vague of a concept for people who are not familiar with that paper).

Following the original paper for clarity is a good idea. The paper refers to "a high-dimensional representation of the entire translation context computed by the MT decoder" and states that "the representations are generated by a single forward pass over each example".

How about using `python3 -m sockeye.generate_decoder_states --output-prefix [output prefix]` and `class DecoderStateGenerator`?

sockeye/layers.py
shuoyangd: Looking closer, adding type annotation for `keys_index` will be tricky because it is supposed to be a `faiss.Index` type, but we want to do optional import for faiss so I don't think there is an easy way to specify it.

(I could do a comment though to circumvent the import.)

sockeye/translate.py
shuoyangd: I'm not sure if I should change the existing signature of a high-level function such as `run_translate`, but if you think this is acceptable then of course we could pass `knn_index` directly.

sockeye/layers.py
fhieber: You could use a [forward reference](https://peps.python.org/pep-0484/#forward-references) as type hint, i.e. `keys_index: 'faiss.Index'`

sockeye/dump_decoder_states.py
```diff
@@ -0,0 +1,239 @@
...
+            sources=sources,
+            targets=targets,
+            source_vocabs=self.source_vocabs,
+            target_vocabs=self.target_vocabs,
+            batch_size=batch_size,
+            max_seq_len_source=self.max_seq_len_source,
+            max_seq_len_target=self.max_seq_len_target
+        )
+
+        for batch_no, batch in enumerate(data_iter, 1):
```
fhieber, line 159: You could consider wrapping this in a `with pt.inference_mode()` (also see [here](https://github.com/awslabs/sockeye/blob/fbbf360303b0db1e9160c9eb62ce24102ca8e772/sockeye/inference.py#L941))

sockeye/layers.py
shuoyangd: Ah TIL :)

sockeye/model.py
shuoyangd: @tdomhan seems like you changed this line -- any idea what docstring update needs to be done?

shuoyangd:
Hi All,

I have added some unit tests and addressed comments from the previous round. This should be ready for another round of reviews.

Both unit tests and style check should pass out-of-the-box, but without faiss, all the faiss-related tests will be skipped, so @mjdenkowski or @fhieber you may want to update the Github workflow to run those tests.

Shuoyang

fhieber:
Thanks @shuoyangd, I'll take a look in the next days.
I think you can actually change the workflow yourself by modifying the yml file in your branch:
You could add the following to https://github.com/awslabs/sockeye/blob/main/.github/workflows/push_pr.yml#L35:
```
- name : Install FAISS
  run: pip install  faiss
```
(or even create a faiss requirements file in your branch).

shuoyangd:
I updated push_pr.yaml following the deepspeed example, together with a few more type annotations to pass mypy tests for python 3.7.

mjdenkowski (commented):
Thanks for continuing to work on this!

The overall design looks good and it's great to see documentation and tests. We just need to work out some of the details so the code is ready to merge.

sockeye/arguments.py
```diff
@@ -1174,6 +1176,42 @@ def add_score_cli_args(params):
...
+    add_device_args(params)
+    add_batch_args(params, default_batch_size=56, default_batch_type=C.BATCH_TYPE_SENTENCE)
+
+    decode_params = params.add_argument_group("Decoder state generation parameters")
+
+    params.add_argument('--state-dtype', default=None, choices=[None, C.DTYPE_FP32, C.DTYPE_FP16],
+                        help="Data type of the decoder state store. Default: %(default)s infers from saved model.")
+
+    params.add_argument('--word-dtype', default=C.DTYPE_INT32, choices=[None, C.DTYPE_INT16, C.DTYPE_INT32],
+                        help="Data type of the word index store. Default: int32.")
```
fhieber, line 1191: You could also use `%(default)s` here as in line 1188.

sockeye/beam_search.py
```diff
@@ -83,10 +85,17 @@ def decode_step(self,
...
                     factor_vocab_size: Optional[int] = None):
-        logits, states, target_factor_outputs = self._model.decode_step(step_input, states, vocab_slice_ids)
+        logits, knn_probs, states, target_factor_outputs = self._model.decode_step(step_input, states, vocab_slice_ids)
         if not self._skip_softmax:
-            logits = pt.log_softmax(logits, dim=-1)
-        scores = -logits  # shape: (batch*beam, output_vocab_size/len(vocab_slice_ids))
+            if knn_probs is None:  # no knn used
+                probs = pt.log_softmax(logits, dim=-1)
+            else:
+                probs = pt.log(self.knn_lambda * pt.softmax(logits, dim=-1) + (1-self.knn_lambda) * knn_probs)
```
fhieber, line 93: have you considered moving this to a traced function? Might be worth benchmarking at some point.

fhieber: You could then also share the interpolation code between `SingleModelInference` and `EnsembleModelInference`

sockeye/beam_search.py
```diff
@@ -65,10 +65,12 @@ class _SingleModelInference(_Inference):
     def __init__(self,
                  model: SockeyeModel,
                  skip_softmax: bool = False,
-                 constant_length_ratio: float = 0.0) -> None:
+                 constant_length_ratio: float = 0.0,
+                 knn_lambda: float = 0.8) -> None:
```
fhieber, line 69: Good candidate for a constant.

sockeye/generate_decoder_states.py
```diff
@@ -0,0 +1,239 @@
...
+from .log import setup_main_logger
+from .model import SockeyeModel, load_model
+from .vocab import Vocab
+from .utils import check_condition
+from .knn import KNNConfig, get_state_store_filename, get_word_store_filename
+
+# Temporary logger, the real one (logging to a file probably, will be created in the main function)
+logger = logging.getLogger(__name__)
+
+# a general blocked mmap stroage interface that could be implemented with numpy memmap, h5py, zarr, etc.
```
fhieber, line 37: typo: storage

sockeye/generate_decoder_states.py
```diff
@@ -0,0 +1,239 @@
...
+        # info for KNNConfig
+        self.num_states = 0
+        self.dimension = None
+        self.state_data_type = utils.get_numpy_dtype(state_data_type)
+        self.word_data_type = utils.get_numpy_dtype(word_data_type)
+
+    @staticmethod
+    def probe_token_count(target: str, max_seq_len: int) -> int:
+        token_count = 0
+        with codecs.open(target, 'r', 'utf-8') as f:
```
fhieber, line 125: Why do we need `codecs.open` here? The default encoding for standard `open` should be utf-8  already in recent versions of Python.

sockeye/generate_decoder_states.py
```diff
@@ -0,0 +1,239 @@
...
+
+        self.state_store_file = NumpyMemmapStorage(get_state_store_filename(self.output_prefix),
+                                                  self.dimension, self.state_data_type)
+        self.words_store_file = NumpyMemmapStorage(get_word_store_filename(self.output_prefix),
+                                                  1, self.word_data_type)
+        self.state_store_file.open(initial_size, 1)
+        self.words_store_file.open(initial_size, 1)
+
+    def generate_states_and_store(self,
+                              sources: List[str],
```
fhieber, line 141: indentation is a bit off

sockeye/generate_decoder_states.py
```diff
@@ -0,0 +1,239 @@
...
+                # get decoder states
+                batch = batch.load(self.device)
+                model_inputs = (batch.source, batch.source_length, batch.target, batch.target_length)
+                if self.traced_model is None:
+                    trace_inputs = {'get_decoder_states': model_inputs}
+                    self.traced_model = pt.jit.trace_module(self.model, trace_inputs, strict=False)
+                decoder_states = self.traced_model.get_decoder_states(*model_inputs)  # shape: (batch, sent_len, hidden_dim)
+    
+                # flatten batch and sent_len dimensions, remove pads on the target
+                pad_mask = (batch.target != C.PAD_ID).squeeze(2)  # shape: (batch, seq_length)
```
fhieber, line 172: Will this work with factored models? `batch.target` should be of shape `(batch, target_length, num_target_factors)` and `num_target_factors` might be >1. Probably better to take the 0th slice of the 3rd dim here.

sockeye/knn.py
```diff
@@ -0,0 +1,175 @@
...
+
+        index.add(keys.astype(np.float32))  # unfortunately, faiss index only supports float32
+
+    def block_add_items(self, index, keys: np.array, block_size: int = 1024*1024):
+        item_count, key_dim = keys.shape
+        assert key_dim == self.config.dimension
+
+        n_blocks = item_count // block_size
+        for i in range(n_blocks):
+            logger.info("adding block no.{0}".format(i))
```
fhieber, line 92: maybe we can use `logger.info("adding block no. {}", i)` or `logger.info(f'adding block no.{i}')`

sockeye/knn.py
```diff
@@ -0,0 +1,175 @@
...
+            train_sample = FaissIndexBuilder.build_train_sample(keys, self.config.train_data_size)
+        
+        index = self.init_faiss_index(train_sample)
+        self.block_add_items(index, keys)
+
+        return index
+
+
+def get_state_store_filename(prefix: str) -> str:
+    return prefix + ".states.npy"
```
fhieber, line 125: +1 to creating constants for these

fhieber (commented):
Agreed that overall design looks good, just a few comments.

sockeye/model.py
tdomhan:
```
:return: logits, list of new model states, other target factor logits.
```

```
:return: logits, KNN output if present otherwise None, list of new model states, other target factor logits.
```

requirements/requirements.faiss.txt
```diff
@@ -0,0 +1 @@
+faiss-cpu >= 1.7.2
```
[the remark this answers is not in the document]
shuoyangd: This requirement file is created for unit test (which is only done on CPU). Do we want to add another requirement file for GPU or even add GPU tests?

sockeye/layers.py
```diff
@@ -117,6 +118,49 @@ def forward(self, data: pt.Tensor, vocab_slice_ids: Optional[pt.Tensor] = None)
...
+        self.state_store = state_store
+
+    def forward(self, data: pt.Tensor):
+        # faiss only supports float32
+        distances, indices = self.keys_index.search(data.cpu().numpy().astype(np.float32), self.k)
+        # Map indices to tokens
+        y = self.vals[(indices+1) % len(self.vals)]
+        y[y == C.BOS_ID] = C.EOS_ID  # no EOS is inserted in generated data store, so we need to use the BOS of the next sentence as EOS
+        # guard vocab index overflow caused by int16
+        # in general, it is recommended to use int32
```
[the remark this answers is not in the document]
shuoyangd: Thought about this a bit more and I think we should have only supported `int32`. `int16` brings down the disk footprint by half, but this is only relevant to the "value" data store and it's rather rare for the "value" data store i.e. vocab items (as opposed to the "keys", i.e. decoder states) to be larger than a few GBs. FYI, key stores are ~1000x the size of value stores.

test/unit/test_knn.py
```diff
@@ -0,0 +1,141 @@
...
+    import faiss  # pylint: disable=E0401
+    faiss_installed = True
+except:
+    faiss_installed = False
+
+
+def test_numpy_memmap_storage():
+
+    # test open
+    with patch('numpy.memmap') as mock_np_memmap:
```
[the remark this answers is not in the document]
shuoyangd: I could be wrong, but I thought one should not read/write disk files for unit tests?

If we can then of course we can use the real one.

sockeye/beam_search.py
shuoyangd: re: tracing -- Yes. Initially we put them in a traced function and added `jit.ignore` to the function that did kNN query (because faiss operations should not be traced). Turns out in that case kNN query is only done once during the initial computational graph compilation and then the same `knn_probs` is reused for all subsequent segments, which is obviously wrong. So the kNN query and the interpolation were later moved out of the traced function (see [this commit](https://github.com/awslabs/sockeye/pull/1062/commits/b9592d44cf638925052adf20a9b79f4e3472f997)).

I cannot find concrete quote from the document or any Googled result but my guess is that `jit.ignore` only works with TorchScript but not tracing.

sockeye/generate_decoder_states.py
shuoyangd: I've had encoding issues with `open` in Python 3.6 but probably with 3.7 and 3.8 it's fixed? And if that's the case, can we assume people will only run Sockeye in Python 3.7+?

sockeye/generate_decoder_states.py
shuoyangd: This also makes me realize that we've never test this with a factored model. Should probably have done it to ensure there is no other dimension bug like this.

requirements/requirements.faiss.txt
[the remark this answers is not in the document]
mjdenkowski: If we support creating/using GPU indexes, we can split the requirements file into `requirements.faiss-cpu.txt` and `requirements.faiss-gpu.txt` and recommend the GPU version for users.

test/unit/test_knn.py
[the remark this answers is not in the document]
mjdenkowski: Different projects may have different preferences about this. For Sockeye, we frequently use temporary files to test code that saves or loads data. For example, we create a temporary directory to test [model quantization](https://github.com/awslabs/sockeye/blob/a3088b4bb48dd1a962f8e88978bcc6b42c52e3ff/test/unit/test_quantize.py#L42).

sockeye/beam_search.py
mjdenkowski: Have you tried scripting the layer instead of tracing it? If you run `jit.script(layer)` when a layer is created, it should remain scripted (instead of traced) when its parent module is traced.

shuoyangd: I didn't -- will try!

sockeye/generate_decoder_states.py
fhieber: Yes, we require Python3.7+ (https://awslabs.github.io/sockeye/setup.html)

sockeye/beam_search.py
shuoyangd: Hi Felix and Michael -- it's been a while since I last worked on this! I looked into the possibility of scripting this module -- unfortunately even this is not possible because `faiss.Index` is not a supported Torchscript type so it cannot be passed as an argument into the constructor. Here is the part of the error message that explains this:

```
Module 'KNN' has no attribute 'keys_index' (This attribute exists on the Python module, but we failed to convert Python type: 'faiss.swigfaiss.IndexFlat' to a TorchScript type. Only tensors and (possibly nested) tuples of tensors, lists, or dictsare supported as inputs or outputs of traced functions, but instead got value of type IndexFlat..
```

Hence, there is no way we can pass a faiss index into a scripted module, as all attributes of a scripted module have to be set in the constructor (otherwise they won't be scripted). So it seems like we'll have to keep this part of code as-is.

I've already addressed the other comments and will update the PR soon.

shuoyangd:
Hmmm, it seems like faiss only works on macOS with `conda install -c pytorch faiss-cpu` but not with `pip`. The official document also recommends installing with `conda` instead of `pip`.

Is this something we can do for the macos test?

shuoyangd:
Disabling faiss-related tests on OS X is easier, but after a little digging I also found [this action that should set up miniconda](https://github.com/marketplace/actions/setup-miniconda).

test/unit/test_knn.py
```diff
@@ -0,0 +1,159 @@
...
+        assert (store.mmap[32:, :] == zeros_block_32).all()
+        # add with size overflow -- this should trigger a warning without doing anything
+        store.add(np.ones((64, 64), dtype=np.float16))
+        assert (store.mmap[:32, :] == ones_block_32).all()
+        assert (store.mmap[32:, :] == zeros_block_32).all()
+
+
+def test_decoder_state_generator():
+    # utils.is_gzip_file expects byte-like object, hence supplying a byte string
+    data = 'One Ring to rule them all, One Ring to find them'
```
mjdenkowski, line 63: Right now `data` is a `str`. Is it intended to be `bytes`?

mjdenkowski (commented):
Thanks for iterating on this!

Most of the remaining points are just around formatting and documentation. Two general notes:

1. It would be great to add docstrings to the various functions and methods in this PR. For simple functions, this can be a single sentence. For more complex functions, we add lines for params and returns.

2. There are several long lines that we would typically break into segments that fit within a 120-character margin.

test/unit/test_knn.py
shuoyangd: I forgot what caused the change exactly but seems like we can use a regular string for this test now. I should update that comment though.

shuoyangd:
Hi All,

I updated the Github workflow and addressed Michael's latest comments in the latest two commits. Looks like workflow checks are all passing now! This is probably a good time to take another look and let me know if there are more comments on the code.

Shuoyang

sockeye/arguments.py
```diff
@@ -1423,3 +1458,44 @@ def add_build_vocab_args(params):
...
+                             'https://github.com/facebookresearch/faiss/wiki/The-index-factory')
+    params.add_argument('--train-data-input-file',
+                        default=None,
+                        type=str,
+                        help='An optional field to reuse an already-built training data sample for the index. '
+                             'Otherwise, a (slow) sampling step might need to be run.')
+    params.add_argument('--train-data-size',
+                        default=None,
+                        type=int,
+                        help='An optional field to specify the size of the training sample. Will override settings in the config.')
```
mjdenkowski, line 1501: Formatting: Break line to fit 120-character margin.

sockeye/arguments.py
```diff
@@ -1423,3 +1458,44 @@ def add_build_vocab_args(params):
...
+                        help='The directory that contains the stored decoder states and values '
+                             f'({C.KNN_STATE_DATA_STORE_NAME} and {C.KNN_WORD_DATA_STORE_NAME}).')
+    params.add_argument('-o', '--output-dir',
+                        default=None,
+                        type=str,
+                        help='The path to the output directory. Will reuse input directory if not specified.')
+    params.add_argument('-t', '--index-type',
+                        default=None,
+                        type=str,
+                        help='An optional field to specify the type of the index. Will override settings in the config. '
```
mjdenkowski, line 1490: Formatting: Break line to fit 120-character margin.

sockeye/generate_decoder_states.py
```diff
@@ -0,0 +1,279 @@
...
+                if (batch_no + 1) % 1000 == 0:
+                    logger.debug("At batch number {0}".format(batch_no + 1))
+
+                # get decoder states
+                batch = batch.load(self.device)
+                model_inputs = (batch.source, batch.source_length, batch.target, batch.target_length)
+                if self.traced_model is None:
+                    trace_inputs = {'get_decoder_states': model_inputs}
+                    self.traced_model = pt.jit.trace_module(self.model, trace_inputs, strict=False)
+                decoder_states = self.traced_model.get_decoder_states(*model_inputs)  # shape: (batch, seq_len, hidden_dim)
```
mjdenkowski, line 188: Formatting: Break line to fit 120-character margin.

sockeye/knn.py
```diff
@@ -0,0 +1,209 @@
...
+    return os.path.join(dir, C.KNN_WORD_DATA_STORE_NAME)
+
+
+def get_config_path(dir):
+    """Get the path to the kNN config file given a kNN export directory."""
+    return os.path.join(dir, C.KNN_CONFIG_NAME)
+
+
+def build_knn_index_package(args):
+    """Top-level function that builds a kNN index package (kNN index and config file) from an existing state and word store."""
```
mjdenkowski, line 152: Formatting: Break line to fit 120-character margin.

sockeye/model.py
mjdenkowski: Formatting: Break line to fit 120-character margin.

mjdenkowski (commented):
Thanks Shuoyang!

I resolved most of the earlier comments and left a few notes where it looks like lines still need to be updated.

shuoyangd:
Sorry, thought the 120-line margin is a soft one (quite a few existing lines go beyond that). Just updated again.

PR #1066, opened 2022-09-12
Add --tf32 device flag for transparent float32 acceleration

graehl:
- --tf32 0|1 bool device (torch.backends.cuda.matmul.allow_tf32)
 enabling 10-bit precision (19 bit total) transparent float32
 acceleration. default true for backward compat with torch < 1.12.
 allow different --tf32 training continuation

- device.init_device called by train, translate, and score

- allow torch 1.12 in requirements.txt

- require pytest 2 (3 fails)

until you can check this box.
- [x] Updated CHANGELOG.md


mjdenkowski (commented):
Thanks for bringing up this point! The PyTorch documentation for [CUDA semantics](https://pytorch.org/docs/stable/notes/cuda.html#tensorfloat-32-tf32-on-ampere-devices) mentions that `allow_tf32` defaulted to `True` until version 1.12. Instead of having a flag for this, we could preserve the default behavior by always setting `torch.backends.cuda.matmul.allow_tf32 = True`. Users would get the performance improvements without needing to update their Sockeye configurations when updating PyTorch.

fhieber:
Thanks Johnathan! 
I agree that defaulting to `allow_tf32` to keep behavior consistent to previous versions would be preferred. 
While you are at it, you could also update [requirements.txt](https://github.com/awslabs/sockeye/blob/main/requirements/requirements.txt) to allow PyTorch 1.12.x (`<1.13.0`)

graehl:
I was unaware that it defaulted true previously. Agree.

graehl:
I rebased into a single commit for all of the above.

graehl:
had to revert pytest<3 requirement (but in fact tests fail w/ pytest 3) due to automated test failure above (tests work locally for me)

fhieber (approved):

fhieber:
Thanks for the changes, I realized I never submitted by pending review from over a month ago, apologies for the delay. I'll merge this now.

PR #1079, opened 2023-01-03
BFloat16 Integration/System Test Update

mjdenkowski:
#### Changes

This PR brings bfloat16 integration and system testing to all platforms with the following changes:
- The `check_train_translate` test function now correctly passes `compare_output` to `test_translate_equivalence` instead of always passing `True`.
- Integration tests do not compare scoring and translation outputs (original intention).
- System tests optionally compare outputs. All existing system tests compare outputs (default behavior unchanged).
- A new bfloat16 system test does not compare scoring and translation outputs. BLEU score is still checked.

#### Investigation

Integration tests that use `--dtype bfloat16` were failing the score/translate output comparison. This appears to be caused by bfloat16 rounding: different orders of calculations plus reduced precision. For a small digits sorting model:
- Float32 inference with `--batch-size 8` and `--batch-size 1` produce the same scores and translations.
- Bfloat16 inference with `--batch-size 8` and `--batch-size 1` produce different scores (average change of 1.5% with outliers up to 16%) but the same translations.
- This appears to be working as intended: bfloat16 is faster with lower precision that should not significantly change the final outputs. PyTorch has [inherent nondeterminism](https://pytorch.org/docs/stable/notes/randomness.html) that generally does not appear to impact float32 calculations, but does appear to impact some bfloat16 calculations.

This issue was identified when tests using nightly builds of PyTorch 2.0 [failed](https://github.com/awslabs/sockeye/actions/runs/3825897790/jobs/6509235124#step:6:567). It appears that a change in either bfloat16 handling or general algorithms (order of calculations) causes bfloat16 scores to differ between scoring and translation on "Linux" platforms where they were previously identical.

To resolve this, we turn off the scoring/translation comparison for bfloat16 cases. This enables running bfloat16 integration and system tests on both "Linux" and "Darwin".

Thanks @fhieber for reporting this issue!

until you can check this box.
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Each version section may have subsections for: _Added_, _Changed_, _Removed_, _D
 ### Fixed
 
 - Fixed sequence copying integration tests to correctly specify that scoring/translation outputs should not be checked.
-- The above fix enables `bfloat32` testing on all platforms.
+- Enabled `bfloat16` integration and system testing on all platforms.
 
 ## [3.1.30]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Note that Sockeye has checks in place to not translate with an old model that wa
 
 Each version section may have subsections for: _Added_, _Changed_, _Removed_, _Deprecated_, and _Fixed_.
 
+## [3.1.31]
+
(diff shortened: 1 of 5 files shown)


fhieber (approved):
Thanks!

PR #1085, opened 2023-02-14
Add blocking cross-attention between decoder and encoded prepended tokens

xingniu:
Add blocking cross-attention between decoder and encoded prepended tokens. Prepended tokens are source tokens before a specified tag (inclusive).

Add a new dictionary-based prepared data format that supports storing length of prepended source tokens (version 7). The previous format (version 6) is still supported.

until you can check this box.
- [x] Updated CHANGELOG.md


fhieber (commented):
Partial review by now, haven't made it into the tests yet

sockeye/inference.py
```diff
@@ -1019,9 +1024,14 @@ def _get_inference_input(self,
...
         for j, trans_input in enumerate(trans_inputs):
             num_tokens = len(trans_input)  # includes eos
-            max_output_lengths.append(self._get_max_output_length(num_tokens))
-            source_np[j, :num_tokens, 0] = tokens2ids(itertools.chain(trans_input.get_source_prefix_tokens(),
-                                                                      trans_input.tokens), self.source_vocabs[0])
+            primary_source_ids = tokens2ids(itertools.chain(trans_input.get_source_prefix_tokens(),
+                                                            trans_input.tokens), self.source_vocabs[0])
+            source_np[j, :num_tokens, 0] = primary_source_ids
+            length_np[j, 0] = num_tokens
+            if self.eop_id != C.INVALID_ID:
```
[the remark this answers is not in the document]
xingniu: I was trying to avoid the look-up operation in favor of latency.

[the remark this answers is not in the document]
fhieber: That's a good point. Would it make sense to make this check part of the get_prepended_token_length function. Its a function called lots of times (also in data preprocessing) and maybe its worth doing the check every single time to avoid the look up in the list.

sockeye/data_io.py
```diff
@@ -398,6 +398,22 @@ def create_shards(source_fnames: List[str],
...
+    """
+    Masks prepended tokens before the end-of-prepending tag (inclusive).
+
+    :param ids: List of token ids.
+    :param max_length: Max length of the sequence.
+    :param eop_id: End-of-prepending tag id.
+    :return: Prepended token mask in ndarray (1 = masked).
+    """
+    prepended_token_mask = np.full(max_length, 0)
+    if eop_id in ids:
```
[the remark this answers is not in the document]
xingniu: Avoided it by using try/except

sockeye/inference.py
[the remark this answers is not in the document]
xingniu: Sure, added.

sockeye/data_io.py
```diff
@@ -451,6 +472,8 @@ def load(self,
                        for (source_len, _), num_samples in zip(self.buckets, num_samples_per_bucket)]
         data_target = [np.full((num_samples, target_len + 1, num_target_factors), self.pad_id, dtype=self.dtype)
                        for (_, target_len), num_samples in zip(self.buckets, num_samples_per_bucket)]
+        # metadata contains the length of prepended tokens
+        data_meta = [np.full((num_samples, 1), 0, dtype=self.dtype) for num_samples in num_samples_per_bucket]
```
mjdenkowski, line 476: If "metadata" always refers to length information, can we rename this to something like `data_prepended_length`? Can we update all instances of "meta" or "metadata" to use the same convention?

test/unit/test_data_io.py
```diff
@@ -468,18 +499,19 @@ def test_get_training_data_iters():
     train_max_length = 30
     dev_line_count = 20
     dev_max_length = 30
-    expected_mean = 1.0
-    expected_std = 0.0
+    expected_mean = 0.9208152692746332
```
fhieber, line 502: Why do these values change? Does this test now run with prepended text always?

fhieber (approved):
looks good to me, thanks for adding this!

test/unit/test_data_io.py
xingniu: That's right. Do you think it's better to parametrize this test with and without using prepended text?

fhieber: Given that without prepended text is probably the more prominent case, I would be in favor of parametrizing the test. Thanks!

fhieber (approved):

test/unit/test_data_io.py
xingniu: Done and made some minor changes (two commits). Could you approve again?

fhieber (approved):

PR #1094, opened 2023-08-15
Refactor inference optimizations

xingniu:
This PR refactors inference optimizations:
- Make the `inference_only` mode switchable.
- Simplify inference optimizations by
  (1) using `eval()` to disable dropout instead of explicitly setting dropout modules to None;
  (2) always using default value `inplace=False` for activation modules.

until you can check this box.
- [ ] ~~Were system tests modified? If so did you run these at least 5 times to account for the variation across runs?~~
- [ ] ~~You have considered writing a test~~
- [x] Updated CHANGELOG.md

### DIFF
--- a/sockeye/layers.py
+++ b/sockeye/layers.py
@@ -1,4 +1,4 @@
-# Copyright 2017--2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2017--2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not
 # use this file except in compliance with the License. A copy of the License
@@ -26,12 +26,12 @@ from . import config
 logger = logging.getLogger(__name__)
 
 
-def get_activation(act_type: str, inplace: bool = False) -> pt.nn.Module:
+def get_activation(act_type: str) -> pt.nn.Module:
     if act_type == C.SWISH1:
-        return pt.nn.SiLU(inplace=inplace)
+        return pt.nn.SiLU()
     if act_type == C.GELU:
         return pt.nn.GELU()
-    return pt.nn.ReLU(inplace=inplace)
+    return pt.nn.ReLU()
 
 
 class LHUC(pt.nn.Module):
@@ -287,7 +287,7 @@ class DotAttentionCell(pt.nn.Module):
 
     def __init__(self, dropout: float = 0.0, heads: int = 1) -> None:
         super().__init__()
-        self.dropout = pt.nn.Dropout(p=dropout) if dropout > 0.0 else None
+        self.dropout = pt.nn.Dropout(p=dropout)
         self.heads = heads
 
     def forward(self,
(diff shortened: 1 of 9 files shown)


sockeye/model.py
```diff
@@ -189,6 +191,19 @@ def __init__(self,
 
         self.knn : Optional[layers.KNN] = None
 
+    def set_inference_only(self, inference_only: bool):
+        """
+        Turn inference_only/eval mode on or off.
+        """
+        if self.inference_only != inference_only:
```
mjdenkowski, line 198: Is this check needed? Does setting `inference_only` to the same value multiple times have any negative effects?

sockeye/model.py
```diff
@@ -144,10 +144,12 @@ def __init__(self,
                                                vocab_size=self.config.vocab_target_size,
                                                weight=output_weight,
                                                dtype=self.dtype)
+        self.output_layer_module_cached = self.output_layer
+        self.output_layer_script_cached = pt.jit.script(self.output_layer_module_cached)
         if self.inference_only:
             # Running this layer scripted with a newly initialized model can
             # cause an overflow error.
-            self.output_layer = pt.jit.script(self.output_layer)
+            self.output_layer = self.output_layer_script_cached
```
mjdenkowski, line 152: Could this be replaced with a call to `self.set_inference_only`?

sockeye/model.py
```diff
@@ -189,6 +191,19 @@ def __init__(self,
...
+        Turn inference_only/eval mode on or off.
+        """
+        if self.inference_only != inference_only:
+            self.inference_only = inference_only
+            if self.inference_only:
+                self.output_layer = self.output_layer_script_cached
+            else:
+                self.output_layer = self.output_layer_module_cached
+            self.decoder.set_inference_only(self.inference_only)
+            self.train(not inference_only)
```
mjdenkowski, line 205: Do we need to set `self.train` here? Could we say that `self.set_inference_only` controls Sockeye-level optimizations (cell state transform and output layer) while `self.train` controls PyTorch-level optimizations (dropout)? This would decouple the two settings to avoid potential unintended side effects when users change one or the other.

mjdenkowski (commented):
Thanks for working on this! I've added a few comments.

sockeye/layers.py
```diff
@@ -812,6 +823,13 @@ def __init__(self,
 
         self.relu = pt.nn.ReLU(inplace=False)  # inplace=False because we need to non-activated data as well
 
+    def set_inference_only(self, inference_only: bool):
+        """
+        Set inference_only.
+        """
```
mjdenkowski, line 829: For consistency, we could add:
```
self.inference_only = inference_only
```

sockeye/transformer.py
```diff
@@ -203,6 +203,13 @@ def __init__(self,
         if config.use_lhuc:
             self.lhuc = sockeye.layers.LHUC(config.model_size, dtype=dtype)
 
+    def set_inference_only(self, inference_only: bool):
+        """
+        Set inference_only.
+        """
```
mjdenkowski, line 209: For consistency, we could add:
```
self.inference_only = inference_only
```

mjdenkowski (commented):
Thanks Xing! Can you make another pass to double check that any class that has a `set_inference_only` method also sets `self.inference_only` in both `set_inference_only` and `__init__`? The idea is that any class that supports switching in/out of inference mode can always be queried for its current state.

mjdenkowski (approved):
Looks good!

PR #1104, opened 2024-02-10
feat(evaluate): add signwriting evaluation metrics

### LINKED ISSUE #1103: Adding custom metrics
There are four metrics available here https://github.com/sign-language-processing/signwriting-evaluation that I would like to report.
How can I add custom metrics? (I would like to optimize for one of the metrics)

AmitMY:
This pull request fixes #1103 

Changes here are complete (this works), but we might want to figure out a different solution, in case this is too much hardcoding for you all (perhaps something like an external evaluation script, and a single custom metric that can call that script)

Note! that unlike BLEU or other metrics, this evaluation takes all factors into account, as they matter for SignWriting. Even if using a non-factored setup though, this evaluation works.

What do you think? how can we integrate this into sockeye?
This is very important to me, to push forward the adoption of SignWriting as a written form of signed languages, allowing both transcription and translation models to be trained with sockeye.

Tests fail with the following error, despite an Nvidia Tesla T4 GPU available:
> ValueError: Expected a cuda device, but got: CPU

until you can check this box.
- [x] Updated CHANGELOG.md


mjdenkowski:
Hi Amit,

It's great to see that you have an implementation working! We can think about two questions for adding new features to Sockeye:
1. How can we make the feature general enough to be broadly useful for the community?
2. How can we make the feature available in a way where it can be easily understood and safely accessed?

For now, it looks like you have an implementation that works for your use case and that you could share with other interested users via a forked repository. In that case, you would control everything and wouldn't need to add extra layers of engineering.

Running your metric with the main version of Sockeye would be more challenging. Adding support for custom metric scripts is a great idea from a science perspective. Since Sockeye is also built to run in production environments, we need to avoid security vulnerabilities like arbitrary code execution. Designing, implementing, and documenting a secure way to run user-defined metrics would significantly increase the scope of this project.

I recommend using your fork of Sockeye for now and revisiting this if using the fork starts to block your work or the work of others working on SignWriting.

Best,
Michael

AmitMY:
Thanks.
I think that since my implementation is generic for SignWriting outputs, it might be valid to introduce inside of sockeye. 
It does not change the behavior for spoken language users, but all of a sudden allows for more experimentation with signed languages.

In any case, for the meanwhile I am successfully using my fork.

